PR #59, opened 2015-10-27
Merge remote-tracking branch 'refs/remotes/kbenoit/master'

pjsio:
reading csv files into textfiles.R with a missing 'textField' produces the following error: 
#### Error in get_doc: 'unrecognized filetype:csv'.
Rather than changing the `setMethod` for `get_data` (textField=index), I simply insert an error message in `get_doc`.


kbenoit:
Hi JK - Thanks for this. The reason it fails is that if the argument is a .csv, then it must also have a textField argument. But a better error message would be welcome. When I checked your "Files changed"  however none were indicated. Plus the branches have moved on since then (but not textfile.R). Can you try this again?

PR #92, opened 2016-01-23
Update README.Rmd

trinker:


kbenoit:
Thanks for catching that!

PR #152, opened 2016-05-13
Fix warning message for skipgrams with n=1

adamobeng:


kbenoit:
For the ngrams.R: how about:
```
    if (identical(as.integer(n), 1L)) {
        if (!identical(as.integer(skip), 0L))
            warning("skip argument ignored for n = 1")
```

For the code checks, please see earlier issue about skip grams, and make sure it passes the CHECK --as-cran.

adamobeng:
`R CMD check --as-cran` produces `Status: 5 WARNINGs, 6 NOTEs`, but no ERRORs...

[The codecov link](https://codecov.io/gh/kbenoit/quanteda/compare/5ebc984e473a271e7b0a629bac0d56d9a77944ce...b1456a98899734068d975cf353c29e9d2d4cdb51) also doesn't seem to have any content.

PR #158, opened 2016-05-14
Fix #157: dfm with non-unique features and empty docs

### LINKED ISSUE #157: dfm fails when there are non-unique features and empty documents
```
> dfm(corpus(c('What does the fox say?', 'What does the fox say?', '')))
Error in validObject(r) :
  invalid class “dgTMatrix” object: length(Dimnames[2]) differs from Dim[2] which is 11
```

The problem is on line 269 of dfm-main.R:
```
featureIndex <- c(featureIndex, rep(length(allFeatures)+1, length(emptyDocs)))
```
should be
```
featureIndex <- c(featureIndex, rep(length(uniqueFeatures)+1, length(emptyDocs)))
```

I'll submit a patch once I've written the tests.

adamobeng:


kbenoit:
Yes this was a bug that I introduced when making the treatment of zero-feature documents more efficient! Well-spotted.

PR #172, opened 2016-05-23
think we need this change to deal with local (not downloaded) zip files

pnulty:
Note this branch goes not include an actual GUI, it's just a small change I needed to help with prototyping a shiny app. First, flocal variable was not set if the file was originally local rather than a http download. I'v fixed this part. Then, it should be noted that this expects for the txt files to be in the top level of the zip. If there is an additional folder when unzipped it wouldn't work for me. I


adamobeng:
Hi @pnulty, I noticed this too. I think we're going to have to make some API decisions to cover the subdirectories case, as well as non-txt files inside zip archives.

This patch works for the current use-case, but it doesn't handle nested subdirectories. Also, I think it depends on a bug in glob2rx which allows wildcards to match directory separators (this will be going away soon).

Could you add some tests which describe the behaviour you expect, so that I can test my new implementation against them?

pnulty:
Thanks Adam, I added a simple testthat for the offline case. This should work for offline import when it's just a simple zip with txt files in the top level.

adamobeng:
lgtm!

kbenoit:
Thanks Paul, Adam

PR #263, opened 2016-10-30
Update plot.kwic for ggplot 2.2.0

adamobeng:
Signed-off-by: Adam Obeng github@binaryeagle.com


kbenoit:
Not passing check yet because ggplot2 is not yet CRANned so I will wait until approx Nov 4 when Hadley refreshes the CRAN version with the changes.

PR #266, opened 2016-11-01
add links in src/Makevars & src/Makevars.win

HaiyanLW:
Some configurations seems missing for applying RcppParallel.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -55,7 +55,8 @@ Suggests:
     testthat,
     RColorBrewer,
     ggplot2 (>= 2.1.0),
-    grid
+    grid,
+    text2vec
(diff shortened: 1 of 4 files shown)


kbenoit:
Same error. So: it's not giving this error when you run the check, with `--as-cran` in the build options, on your Linux machine?

kbenoit:
It gave this output on my CHECK:

```
* checking for GNU extensions in Makefiles ... WARNING
Found the following file(s) containing GNU extensions:
  src/Makevars
Portable Makefiles do not use GNU extensions such as +=, :=, $(shell),
$(wildcard), ifeq ... endif. See section ‘Writing portable packages’ in
the ‘Writing R Extensions’ manual.
```

So I think it's just the bits inside the `Makevars`* files.

PR #280, opened 2016-11-07
Dev join tokens

koheiw:
There should be almost not conflict at this stage.


koheiw:
Updated for those three points, but I cannot check it, because my RStudio is completely broken. Please take over or leave it until I will fix my RStudio. I am off.

Issue #252, opened 2016-10-21
methods for (hashed) tokens class objects


kbenoit:
Add methods for tokenizedTextsHashed class for:
- collocations  
- dfm
- kwic
- ntoken
- ntype  
- syllables  
- toLower  
- toUpper  
- wordstem  

Update 
- print

kbenoit:
Following https://github.com/kbenoit/quanteda/commit/2ed062502d737c805e215adc754ec8dffaea5933, only the following remain to be written for this class:

*  `collocations.tokenizedTextsHashed`  
*  `kwic.tokenizedTextsHashed`  
*  `selectFeatures.tokenizedTextsHashed`

Issue #188, opened 2016-05-30
Multi-word entries in dictionary appear to be ignored


pablobarbera:
The following code is a simplified version of the example in quanteda::dictionary

``` r
mycorpus <- subset(inaugCorpus, Year>1900)
mydict <- dictionary(list(country = "united states"))
sum(dfm(mycorpus, dictionary = mydict)[,"country"])
```

Another example:

``` r
mycorpus <- corpus("this should work")
mydict <- dictionary(list(example = "should work"))
sum(dfm(mycorpus, dictionary = mydict)[,"example"])
```

Trying with "_" as concatenator:

``` r
mycorpus <- corpus("this should work")
mydict <- dictionary(list(example = "should_work"), concatenator="_")
sum(dfm(mycorpus, dictionary = mydict)[,"example"])
```

Am I missing something? This came up as I was trying to use quanteda with Lexicoder, which has multi-word entries in the dictionary. I'm running quanteda 0.9.6-9

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/9231340245136/156067376936362)

conjugateprior:
If you've got Java 8 installed, which I guess you might if you're using lexicoder, then you could use [jca](https://github.com/conjugateprior/jca) while this gets fixed.

kbenoit:
@pablobarbera This feature is not live yet, but there are workarounds using "phrasetotoken". But lots of demand for this exists so we need to add it asap.

@koheiw let's talk about how the newest selectFeatures mods could be used for this purpose, prior to dfm construction.

koheiw:
Joining tokens before applying dfm is the way. It is not automated, but I am sure that you can extract multi-part words easily from the dictionary.
```
seqs <- list(c('should', 'work'), c('united', 'states'))
toks2 <- joinTokens(toks, seqs, "_")
```

kbenoit:
@koheiw I was working today on a foundation to get this issue resolved, in the **dev_multiWordDictionaries** branch. Please see the changes I made to `joinTokens()`. 

1.  I will eventually fold this function into `phrasetotoken` with a signature for tokenizedTexts, vector/list of phrases. The function I added for this (bottom of `phrases.R`) currently calls `joinTokens()`. 

2.  The examples in the roxygen2 header show some cases where it's not working. There is a bizarre outcome based on which match is first, probably because of the loop's sequential processing of the token sequences. Can you investigate this?  (See the @examples)

[...]

PR #288, opened 2016-11-08
Dev multi key dict

koheiw:
Join tokens is used for multi-words entries in applyDictionary.


koheiw:
Underlying function of joinTokens is qatd_cpp_replace_hash_list, which just replace sequence of IDs with a new ID, so it is not joining anything ether. This is faster than any R implementations. The complexity here is support of regex and glob. Much of the R code in joinTokens, including grid_sequences, is for this actually. These are why I think we are better off just wrapping joinTokens. We have a lot of other things to do...

kbenoit:
@koheiw two things that need to be addressed with the behaviour, which will solve the broken tests:

1.  `applyDictionary.tokens()` should return the dictionary keys only, to match the behaviour of `applyDictionary.tokenizedTexts()`.  I added this to `tests/testthat/test_dictionary_mutiword.R`. You can see the problem here:

```r
    txt <- c(d1 = "The United States is bordered by the Atlantic Ocean and the Pacific Ocean.",
             d2 = "The Supreme Court of the United States is seldom in a united state.",
             d3 = "It's Arsenal versus Manchester United, states the announcer.",
             d4 = "We need Manchester Unity in the Federal Republic of Germany today.",
             d5 = "United statehood is a good state.",
             d6 = "luv the united states XXOO!")
    toks <- tokenize(txt, removePunct = TRUE)
    toks_hashed <- tokens(txt, removePunct = TRUE)
    dict_mw_fixed <- dictionary(list(Countries = c("United States", "Federal Republic of Germany"),
                                     oceans = c("Atlantic Ocean", "Pacific Ocean"),
                                     Institutions = c("federal government", "Supreme Court"),
                                     team = c("Manchester United", "Arsenal")))

# not desired output - looks more like joinTokens
applyDictionary(tokens("The United States is big."), 
                dictionary = dictionary(list(COUNTRY = "United States")), 
                valuetype = "fixed", concatenator = " ")
## tokenizedTexts from 1 document.
## Component 1 :
## [1] "The"           "United States" "is"            "big"           "."            

# desired output
applyDictionary(tokens("The United States is big.", hash = FALSE), 
                dictionary = dictionary(list(COUNTRY = "United States")), 
                valuetype = "fixed", concatenator = " ")
## tokenizedTexts from 2 documents.
## Component 1 :
## [1] "COUNTRY"
##
## Component 2 :
## character(0)
```

2.  Partly because your method redefines the tokens, whereas the method for `applyDictionary.tokenizedTexts` simply counts the sequence matches, it cannot count overlapping sequences matches. Therefore the behaviour is different depending on the order of the dictionary keys. I'd like the behaviour to be invariant by counting overlapping matches. "Manchester United States" will count 1 for team = "manchester united" **and** 1 for country = "United States".  See line 110 of `tests/testthat/test_dictionary_mutiword.R`.

koheiw:
I did not issue a PR because of the Manchester United problem. In order to detect overlapping keys, we have do duplicate the tokens object. If all other words are dropped, the copy is small, but joinTokens should not be like this. Maybe we cannot to applyDictionary efficiently using that function.

koheiw:
I added a C++ function different from one for joinTokens, but I ended up making a function that directly create dfm with multi-words keys. I think this is better in many cases, but I am not sure what to do with this next. It is inappropriately in the same name as before, applyDicitonary.tokens().

kbenoit:
Actually we don't have to duplicate anything to solve the Manchester United problem, if we just count the matches to sequences. This is why I am using the `matchSequence()` in `applyDictionary.tokens()`, because it simply counts the frequencies of sequence matches and uses those to replace the matched tokens with the dictionary keys. That's the behaviour I want for `applyDictionary()`. Once that is working quickly - and I think it is! - then constructing the dfm is simply a matter of calling the `dfm.tokens()` on a hashed and greatly reduced set of tokens. I don't think we need a separate method to do that. In any case, it should not be named `applyDictionary.tokens()`. Since you are using a function of that name to produce a dfm, I think that is why the tests (and hence the build check) are failing.

kbenoit:
I branched from your branch today around 11am NY time, and started work to finish #252. This included writing (in R) a fast method for `selectFeatures.tokens()`. It uses a regex to match the vocabulary, but a fastmatch of integers to match the hash key for the vocabulary match to the hashed tokens. The regex part is small but it makes implementing the `case_insensitive` and the `valuetype` very easy. And no C++ was required. But happy to compare performance if you want.

If we do want to compare performance, I suggest writing a standalone method with a different name, that we can use to compare, without rewriting existing methods.

On the PR #288 in general, I suggest you merge the newest master into it, see what you want to keep, and make sure it passes check. This will be a lot easier for you if first you:

*  rename the function you called `applyDictionary.tokens()`, move it into a separate .R file  
*  change the names of your `test_that()` blocks in `tests/testthat/test_dictionary_multiword.R`, or even move them to a new test file, so that you don't get merge conflicts, since I kept or updated or extended (or all of these) that test file today as I ironed out issues with the existing methods.

kbenoit:
See my latest comment on Issue #188. 

Once you have changed the function name to something that can co-exist with the existing `applyDictionary.tokens()`, you should see the merge conflict warnings disappear. Also, you should then be able to merge `master` into this branch so that the rest of your code us up to date.

kbenoit:
@koheiw I fixed a few things and then merged this into master as `dictionaries2.R` with the new `applyDictionary2()`. It also contains your updates to `joinTokens.R` and the new `.cpp` code.

I'm deleting the branch now, any new changes should be made in new branches and new PRs.

PR #297, opened 2016-11-12
Dev multi key dict2

koheiw:
I am working in a separate repository, while I am waiting. But I need this to be merged to master, to compare my old code with my new code. Why don't we use dev branch for development of this kind? You would reviews the code, but that code will be overwritten by newer code immediately after the marge.


kbenoit:
I'll leave this branch active, since it contains the changes, but I suggest that you choose exactly what you want to test, and get a *new version of* `master`, and manually add the new files (e.g. `applyDictionary2()` and the associated `.cpp` files), not by `git merge`. 

Please explain any changes and limit the diff to those files actually changed.

And make sure the build passes before issuing a PR. Thanks!

PR #301, opened 2016-11-14
Dev mutli key dict3

koheiw:


kbenoit:
@koheiw please see my inline comments. Great that we are now passing both travis-ci checks, but I'd like to see tests added for the new `regex2fixed()` function, as well as for `applyDictionary2()`. 

You can create a new test file and adapt the tests from `test_dictionary_multiword.R` if you wish. We should test all conditions, including all valuetypes, and single and multi word.

Thanks!

PR #308, opened 2016-11-16
Dev join tokens

koheiw:
Faster and more stable version of joinToken for tokens objects.


kbenoit:
Also we should start adding tests for the C++ functions.

koheiw:
grid_sequence is killed along with old joinTokens. Its successor is regex2fix in a separate file. I can add test for C++ functions, but their bugs should be captured in existing tests, like test_joinTokens.

kbenoit:
RIP `grid_sequence()`. 

Adding tests and/or usage examples for the C++ functions would also serve as a type of documentation. But OK for now.

PR #326, opened 2016-11-18
Dev regex2fixed3

koheiw:
regex2fixed3() is about 5x faster than regex2fixed2(). This is my last push in this week.


kbenoit:
@koheiw Am I right in thinking that all of the functions that need a `regex2fixed()` function now switched over to `regex2fixed3()`, with the older ones retained just for benchmarking and comparison?

koheiw:
You are right. JoinTokens and applyDictionary2 are the only functions that were dependent on regex2fixed, but they are on regex2fixed3 now. You can delete old ones.

PR #331, opened 2016-11-22
fix #327

### LINKED ISSUE #327: fcm() different behaviour depending on tokens() v. tokenize()
I think this is due to the internal behaviour of the C++ functions. I prefer NOT to sort it, since in the branch `dev_API-issue317` I have implemented `fcm_sort()`, so we could easily and quickly do this in R rather than in C++.

```r
> txt <- "A D A C E A D F E B A C E D"
> fcm(tokens(txt), context = "window", window = 3)
Feature co-occurrence matrix of: 6 by 6 features.
6 x 6 sparse Matrix of class "fcm"
        features
features A D C E F B
       A 2 4 4 5 2 1
       D . . 3 4 1 1
       C . . . 3 . 1
       E . . . . 2 2
       F . . . . . 1
       B . . . . . .
> fcm(tokenize(txt), context = "window", window = 3)
Feature co-occurrence matrix of: 6 by 6 features.
6 x 6 sparse Matrix of class "fcm"
        features
features A B C D E F
       A 2 1 4 4 5 2
       B . . 1 1 2 1
       C . . . 3 3 .
       D . . . . 4 1
       E . . . . . 2
       F . . . . . .
```

HaiyanLW:
rename test file to test-fcm.R
use fcm-sort in test file


kbenoit:
@HaiyanLW looks good, but what about this, where the features are sorted in one case but not in the other?

(Also: We probably need a vignette to explain what co-occurrence means here, since at first it's a bit counterintuitive as to why the A, D cell should be 12. But it's because there are 6 unordered A, D pairs that make 12 ordered pairs? As I mentioned before we should write a whole paper on the definition and implementation of co-occurrence!)

```r
> txt <- "A D A C E A D F E B A C E D"
> fcm(tokens(txt), context = "document")
Feature co-occurrence matrix of: 6 by 6 features.
6 x 6 sparse Matrix of class "fcm"
        features
features A B C  D  E F
       A 6 4 8 12 12 4
       B . 0 2  3  3 1
       C . . 1  6  6 2
       D . . .  3  9 3
       E . . .  .  3 3
       F . . .  .  . 0
> fcm(tokens(txt), context = "window", window = 4)
Feature co-occurrence matrix of: 6 by 6 features.
6 x 6 sparse Matrix of class "fcm"
        features
features A D C E F B
       A 2 7 4 6 2 2
       D . . 3 4 1 2
       C . . . 3 2 1
       E . . . 2 2 2
       F . . . . . 1
       B . . . . . .
```

R/fcm.R
```diff
@@ -184,9 +184,7 @@ fcm.tokenizedTexts <- function(x, context = c("document", "window"),
                                           dims = c(lengthToken , lengthToken))
         diag(result) <- 0
         result <- result + diagCount
-        
-        # order the features alphabetically
-        result <- result[order(rownames(result)), order(colnames(result))]
+        result <- result[rownames(result), colnames(result)]
```
kbenoit: This simply selects all rows and columns, and is therefore redundant. Better to comment out original line 189 to show that it was considered but disabled.

PR #364, opened 2016-12-06
fix #363

### LINKED ISSUE #363: textstat_dist(x,y,margin="features") is not working when y in not NULL for some methods
for methods: euclidean, jaccard, eJaccard

HaiyanLW:


R/textstat_simil.R
```diff
@@ -124,15 +124,15 @@ textstat_simil <- function(x, selection = character(0), n = NULL,
     
     # truncate to n if n is not NULL
     if (!is.null(n))
-        result <- head(result, n)
+        result <- head(as.matrix(result), n)  # result returned by some method is a sparse matrix
```
kbenoit: Doesn't the `as.matrix()` coerce it to a dense matrix?? I thought we were trying to avoid that.

PR #369, opened 2016-12-08
Hotfix 363

HaiyanLW:
Call as.dist() to replace in-code generation of dist object
Rename "tri" to "upper" to be consistent with proxy::dist()


R/textstat_dist.R
```diff
@@ -134,39 +97,37 @@ textstat_dist <- function(x, selection = character(0), n = NULL,
...
+    # # only retain lower triangular elements for the dist object
+    # distM <- result[row(result) > col(result)]
+    # 
+    # # set the attributes of the dist object
+    # attributes(distM) <- NULL
+    # attr(distM, "Size") <- nrow(result)
+    # if (!is.null(rownames(result))) attr(distM, "Labels") <- rownames(result)
+    # attr(distM, "Diag") <- diag
+    # attr(distM, "Upper") <- upper
+    # attr(distM, "method") <- method
```
kbenoit: Note: You can still add the additional attributes (such as "method") after the as.dist() call

R/textstat_simil.R
```diff
@@ -113,41 +113,48 @@ textstat_simil <- function(x, selection = character(0), n = NULL,
...
+    #     attr(ans,"Labels") <- colnames(result)
+    # attr(distM, "Diag") <- diag
+    # attr(distM, "Upper") <- upper
+    # attr(distM, "method") <- method
+    # attr(distM, "call") <- match.call()
+    # attr(distM, "dimnames") <- NULL
+    # class(distM) <- "dist"
     
+    # create a new dist object
+    distM <- stats::as.dist(result, diag = diag, upper = upper)
```
kbenoit: Ditto on the comments on the dist function.

kbenoit:
See my comments

PR #371, opened 2016-12-09
Dev corpuszip

koheiw:


kbenoit:
This is a very significant reimplementation of the `corpus` class, and we would need to implement and extensively test all existing methods to make sure that they work as expected (meaning: same as for uncompressed corpus methods).

A list of those to implement:

  - [ ] `texts`, `docvars`
  - [ ] `texts<-`, which currently does not work with indexed assignment
  - [ ] `docvars`, `docnames`
  - [ ] `ndoc`
  - [ ] `as.character`
  - [ ] indexing methods, `[`, `[[`, `[[<-`
  - [ ] concatenation methods, `c`, `+`
  - [ ] `print` method (to indicated that the corpus is zipped). Possibly we could record the compression % by measuring this before constructing and compressing, and getting the object.size after, and then using this stored value to print the compression advantage.
  - [ ] `summary`

koheiw:
I added print and summary functions, but I these are using existing functions for corpus objects. I think corpuszip is an immutable object, so if users want to change its content, corpuszip should converted to corpus by as.corpus.

koheiw:
I was shocked when I found a lot of `\n` in `data_corpus_*`, because memCompress uses `\n` as a string delimiter. I hope your `quanteda_document_delimiter` will save this branch.

kbenoit:
Yes I think I fixed the \n issue. See tests-corpuszip.R. Also because the docvars are still in $documents now, most of the corpus methods work for corpuszip objects. But needs full testing.

PR #372, opened 2016-12-10
Dev tokens

koheiw:
This is the full implementation of a function formerly called `tokenize_hashed`. With this blocked tokenization, we will have no problem with large corpus (Twitter or Gruardian).


kbenoit:
I like it. This does two things, from my reading of the code. First, it batch hashes the tokens, so that only 10,000 documents at a time are affected. I'd like to be able to manipulate this setting, so I might implement that as one of the options - probably set using `options(quanteda_tokens_batch_size = 10000)` - to test performance on things that have lots of short documents, such as Twitter data.

Second, it separates the code in `tokens()` depending on the `what` argument. I like this too and was meaning to do the same soon, especially as part of the plan to make tokenization modular. It's this part I want to review a bit more and possibly modify before accepting the PR.

koheiw:
The first thing is to limit peak memory usage. Parameterization of block sizes is a good idea - either via an argument or package settings.  The second thing was just to make maintenance easier. I did not change tokenization per se at all. Aki needed this to tokenize the twitter post actually, and I also need this to tokenize the Guardian corpus.

Issue #378, opened 2016-12-20
dfm() works only up to 40.000 documents, does not work with 40.001 documents


bachl:
There seems to be a problem with converting a corpus with more than 40.000 documents to a document-feature-matrix. Here is a reproducible example:

``` r
random_texts = replicate(4e4+1, paste0(replicate(sample(15:25, 1), paste0(sample(letters[1:6], 5, T), collapse = "")), collapse = " "))
random_corpus = corpus(random_texts)

# does not work; message: "Error in validObject(r): invalid class “dgTMatrix” object: length(Dimnames[1]) differs from Dim[1] which is 7776"
random_dfm = dfm(random_corpus, tolower = F, stem = F, remove = NULL, removeNumbers = F, removePunct = F, removeSymbols = F, what = "fastestword", verbose = T)

#works
random_dfm = dfm(random_corpus[1:4e4], tolower = F, stem = F, remove = NULL, removeNumbers = F, removePunct = F, removeSymbols = F, what = "fastestword", verbose = T) 
```

``` r
> sessionInfo()
R version 3.3.1 (2016-06-21)
Platform: x86_64-apple-darwin13.4.0 (64-bit)
Running under: OS X 10.11.4 (El Capitan)

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] parallel  stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] quanteda_0.9.8.9024 dplyr_0.5.0         purrr_0.2.2         readr_1.0.0         tidyr_0.6.0        
 [6] tibble_1.2          ggplot2_2.2.0       tidyverse_1.0.0     forcats_0.1.1       ggstance_0.3       
[11] stringi_1.1.2       urltools_1.6.0      tuber_0.7.0         Rfacebook_0.6.10    httpuv_1.3.3       
[16] rjson_0.2.15        httr_1.2.1         

loaded via a namespace (and not attached):
 [1] Rcpp_0.12.8         ca_0.70             magrittr_1.5        munsell_0.4.3       lattice_0.20-34    
 [6] colorspace_1.3-2    R6_2.2.0            fastmatch_1.0-4     plyr_1.8.4          tools_3.3.1        
[11] grid_3.3.1          data.table_1.10.0   gtable_0.2.0        DBI_0.5-1           RcppParallel_4.3.20
[...]

PR #379, opened 2016-12-20
Fix incremental tokenization bug

koheiw:
This is a hotfix for issue #378


koheiw:
@kbenoit, this is a simple but urgent fix, so I hope you can accept this remotely.

Issue #391, opened 2016-12-30
tokens_ngrams(x, n = ...) fails when ntokens(x) < n


kbenoit:
```r
packageVersion("quanteda")
## [1] ‘0.9.8.9029’

tokens_ngrams(tokens("a"), n = 2)
## Warning: stack imbalance in '.Call', 38 then 39
## Warning: stack imbalance in '{', 35 then 36
## Warning: stack imbalance in '<-', 28 then 29
## tokens from 1 document.
## Component 1 :
## character(0)
```

Note this is not the case for `char_ngrams()`:
```r
char_ngrams("a", n = 2)
## NULL
```

koheiw:
I found this is coming from `stringi::stri_encode` when it tries to set encoding of empty type vectors.

kbenoit:
Still not solved; I think it's part of the C++ code in `qatd_cpp_tokens_ngrams` rather than coming from `stringi::stri_encode()`. I did use the `stringi::stri_encode()` in a modified `types()` in the latest commit (87251aa) however. Please use the `types()` functions instead of directly addressing the `attr(x, "types")` since this encapsulates and centralises this functionality, meaning if we change the object type for some reason, we can change the behaviour in the single function.

koheiw:
That is strange. I do not see warnings in the branch.
```
> tokens_ngrams(tokens("a"), n = 2)
tokens from 1 document.
Component 1 :
character(0)
```
I am not using `type()` here because it does not work with the list object returned from the C++ function. If I add class attributes in the C++ code, it will work, but not sure I should do this. 
```
 x <- qatd_cpp_tokens_ngrams(x, types(x), concatenator, n, skip + 1)
types_ngm <- attr(x, 'types')
```

kbenoit:
That branch is deleted. While you might have a local copy, it's gone from the repo. You should create a new branch to solve this.

I modified `types()` to make it work, and as per comments above prefer you do as well to fix this. And make the test pass when it's commented out.

koheiw:
Do you mean that you deleted bugfix_391-392 without merging my fix to the master, and I have to push another branch with the same fix?

kbenoit:
No those fixes are all in the master now, except the last fix part of #391. make a new issue-391 and all your previous work will be in there. (Previous work subject to a few modifications on my end.)

koheiw:
I can see the warnings in the master. It looks like another trick by tokens methods.

```
[...]

PR #395, opened 2016-12-30
Bugfix 391 393

koheiw:
Set of small fixes to reach v0.99.


R/tokens_ngrams.R
```diff
@@ -81,17 +81,7 @@ tokens_ngrams <- function(x, n = 2L, skip = 0L, concatenator = "_") {
 #' @noRd
 #' @export
 tokens_ngrams.character <- function(x, n = 2L, skip = 0L, concatenator = "_") {
-    # trap condition where a "text" is a single NA
```
kbenoit: I'm replacing the original behaviour since this is not intended to run tokens on a character - it only works on vectors of characters (tokens).

kbenoit:
@koheiw A few things to note:

1. This does not fix #391:

```r
tokens_ngrams(tokens("a", n = 2))
# Warning: stack imbalance in '.Call', 75 then 76
# Warning: stack imbalance in '{', 72 then 73
# Warning: stack imbalance in '{', 64 then 65
# tokens from 1 document.
# Warning: stack imbalance in '.Call', 82 then 83
# Warning: stack imbalance in '{', 79 then 80
# Warning: stack imbalance in '{', 71 then 72
# Component 1 :
# character(0)
```

See the tests in `test-tokens_ngrams.R`, lines 89-93.  We should be able to pass that test when these are uncommented, to solve #391.

2. I encapsulated the attribute preservation into a function in `utils.R`, that takes an object or its attributes and can re-assign these to another object, with a character vector of exceptions (e.g. `"types"`). I used these in `tokens_ngrams()` rather than the ad hoc methods used in the PR. An even better way would be to do this in the `qatd_cpp_tokens_ngrams()` function (C++) through the sort of setting in the last line, so that attributes sent are returned, except the ones that need to be updated (e.g. `n`, `skip`, `concatenator`, in addition to the already-set `types`).

Issue #388, opened 2016-12-30
tokens_lookup() does not match different keys to overlapping values


kbenoit:
Compare the old behaviour (my preference) to the new:

```r
txt <- c(d1 = "The United States is bordered by the Atlantic Ocean and the Pacific Ocean.",
         d2 = "The Supreme Court of the United States is seldom in a united state.")
toks <- tokens(txt)

dict_fixed <- dictionary(list(Countries = c("States"),
                        oceans = c("Atlantic", "Pacific"),
                        gameconsoles = c("Xbox", "Nintendo"),
                        swords = c("States")))
tokens_lookup(toks, dict_fixed, valuetype = "fixed")
# tokens from 2 documents.
# d1 :
# [1] "swords" "oceans" "oceans"
# 
# d2 :
# [1] "swords"

applyDictionary(tokens(txt), dict_fixed, valuetype = "fixed")
# tokens from 2 documents.
# d1 :
# [1] "Countries" "oceans"    "oceans"    "swords"   
# 
# d2 :
# [1] "Countries" "swords"  
```

Same behaviour for other `valuetype` values:

```r
dict_glob <- dictionary(list(Countries = c("Stat*"),
                             oceans = c("*ic"),
                             gameconsoles = c("?box", "Nintendo*"),
                             swords = "*s"))
tokens_lookup(toks, dict_glob, valuetype = "glob")
# tokens from 2 documents.
# d1 :
# [1] "swords" "swords" "oceans" "oceans"
# 
# d2 :
# [1] "swords"    "swords"    "Countries"

dict_regex <- dictionary(list(Countries = c("Stat.*$"),
                              oceans = c("[A-Z][a-z]+ic"),
                              gameconsoles = c("Xbox"),
                              swords = "s$"))
tokens_lookup(toks, dict_regex, valuetype = "regex")
# tokens from 2 documents.
# d1 :
# [1] "swords" "swords" "oceans" "oceans"
# 
# d2 :
# [1] "swords"    "swords"    "Countries"
```

Note that the existence of the `valuetype = "regex"` and `valuetype = "glob"` means that we can close #292.

koheiw:
I was not thinking about entries in multiple categories. I will fix this using `unordered_multimap`.

koheiw:
bugfix_388 outputs like this:

```
> txt <- c(d1 = "The United States is bordered by the Atlantic Ocean and the Pacific Ocean.",
[...]

PR #396, opened 2016-12-31
Bugfix 388

koheiw:


kbenoit:
@koheiw I've pulled this into `master` now, but it does not close #388 yet (see notes there). I'll leave the branch in place so you can complete #388 and then issue a new PR.

PR #402, opened 2017-01-02
Fix bug in corpus text_field

christophergandrud:
The corpus function mentions that an integer can be supplied to the `text_field` argument in `corpus`. However, this returns an error where `text_fieldi` is not found. 

The cause of this error is that `text_fieldi` is not created when the `text_field` value is not a character. 

This pull request fixes this bug, checks that if  `text_field` is present that it is length = 1.


kbenoit:
Thanks @christophergandrud! This is some of the oldest parts of original code, so I'm not surprised it lacks robustness. I'll review and accept the PR soon. Will probably add some tests to increase coverage and catch more, similar errors.

christophergandrud:
Great!

PR #403, opened 2017-01-02
Dev kwic

koheiw:
Please give kwic2 the final touch.


kbenoit:
See my latest commits to this branch. It's great to fix the internals, but not to redefine the object class, since this breaks the other methods that rely on it. Can you fix the structure so that the new and old versions are the same output object? Then the (old) print method will work, and the `textplot_xray` and other methods.

koheiw:
Do you want me to add `docname` and `position` to the data frame?

kbenoit:
yes please, so that the structure of `kwic(x, pattern new = FALSE)` is the same as the structure of `kwic(x, pattern,  new = TRUE)`

koheiw:
OK I will, but don't you think `positionLabel` and `contextPre` are really long variable names?

kbenoit:
let’s use the existing ones, and if we want to change them to shorter names we make that adjustment separately.

“First, do no harm”

conjugateprior:
How about LHS and RHS? (Or lowercased versions thereof)

kbenoit:
Currently:

```r
> names((kwic(data_char_inaugural, "nuclear")))
[1] "docname"     "position"    "contextPre"  "keyword"     "contextPost"
```

koheiw:
Changes are made.
```
> str(kwic(toks, list("is", "a", c("is", "it")), valuetype = "fixed", new = TRUE))
Classes ‘kwic2’, ‘kwic’ and 'data.frame':	9 obs. of  5 variables:
 $ position   : chr  "2:3" "2:3" "2:2" "4:4" ...
 $ contextPre : chr  " This" " This" " What" " What is in" ...
 $ keyword    : chr  "is a" "is it" "is" "a" ...
 $ contextPost: chr  "test " ". " "in a train ? " "train ? " ...
 $ docname    : chr  "text1" "text2" "text3" "text3" ...
```
KWIC would be better presented in HTML just like profvis. This is for the future improvement.

kbenoit:
Sounds like a good issue - maybe a type of view method that calls up a browser?

koheiw:
Yes, we can steal some code from profvis.

PR #406, opened 2017-01-04
Dev kwic2

koheiw:
Just to check with Appveyor.


kbenoit:
@koheiw did you branch this from master, or from `dev_kwic`? If from `master`, then it's not a solution to the broken `dev_kwic`...

koheiw:
Is the new kwic in the master?

kbenoit:
No - we are trying to fix `dev_kwic` first so that we can push it to `master`. 

How about you add to `dev_kwic` whatever fixes are now in `dev_kwic2`. Of course make sure you fetch the latest changes to `dev_kwic2` first.

koheiw:
I am only messing with dev_kwic2 to identify the cause now. Fix will be added to dev_kwic.

Issue #370, opened 2016-12-09
tokens_hashed_recompile should compress types not found too


kbenoit:
Right now, slicing off values can still result in types not found in remaining documents. We could add an option to preserve this behaviour, but turn it off by default to make the objects as compact as possible.

Currently:
```r
toks <- tokens(c(d1 = "one two three", d2 = "four five six", d3 = "seven eight"))
str(toks)
# List of 3
# $ d1: chr [1:3] "one" "two" "three"
# $ d2: chr [1:3] "four" "five" "six"
#  $ d3: chr [1:2] "seven" "eight"
#  - attr(*, "class")= chr [1:3] "tokens" "tokenizedTexts" "list"
#  - attr(*, "types")= chr [1:8] "one" "two" "three" "four" ...
#  - attr(*, "what")= chr "word"
#  - attr(*, "ngrams")= int 1
#  - attr(*, "concatenator")= chr ""
str(quanteda:::tokens_hashed_recompile(toks[1]))
# List of 1
#  $ d1: chr [1:3] "one" "two" "three"
#  - attr(*, "class")= chr [1:3] "tokens" "tokenizedTexts" "list"
#  - attr(*, "types")= chr [1:8] "one" "two" "three" "four" ...
#  - attr(*, "what")= chr "word"
#  - attr(*, "ngrams")= int 1
#  - attr(*, "concatenator")= chr ""
```
If this were compact, then the length of `attr(*, "types")` in the last output would only have three elements, not the original eight.

koheiw:
The easiest way is to reconstruct character tokens and serialize again. The cost is very small when tokens objects are small.

PR #416, opened 2017-01-05
Issue 413

koheiw:
This branch is to modernize `findSequence()` with tokens objects and parallel computation.


kbenoit:
In the current code, the types from the original are retained even though the selected object changes them:

```r
> tokens_select(toksh, feats, selection = "keep")
tokens from 2 documents.
doc1 :
[1] "This"   "is"     "SAMPLE"

doc2 :
[1] "this"   "sample" "is"    

> attributes(tokens_select(toksh, feats, selection = "keep"))$types
[1] "This"   "is"     "a"      "SAMPLE" "text"   "this"   "sample" "better"
```

We have two options to resolve this:

1. return the types as a list as per qatd_cpp_tokens_ngrams, and copy all attributes *except* types as per https://github.com/kbenoit/quanteda/blob/master/R/tokens_ngrams.R#L129-L134 

2. Compress the types in a modified `tokens_hashed_recompile(x)`.

kbenoit:
Actually: If you change the last lines of `tokens_select.tokens()` to this:

```r
    # # the fix for issue #394
    # if (padding) x <- tokens_hashed_recompile(x)
    
    tokens_hashed_recompile(x)
```
then it solves the issue, since I added types reindexing to remove those not found in #370.

I did not make this change since I see you committing since my comment.

PR #418, opened 2017-01-06
Issue 355

koheiw:
For adding `exclusive` argument of tokens_lookup.


kbenoit:
Good - please add a few tests too.

PR #448, opened 2017-01-12
Issue 447

koheiw:


kbenoit:
Not sure entirely what your latest commit did, but I noticed that the performance of `sequences()` suffered when building from this branch.

Don't be shy on using the second and subsequent lines of a commit message for describing in more detail what the commit is designed to accomplish. e.g. http://chris.beams.io/posts/git-commit/

koheiw:
As you correctly said, I changed `sequences()` to the wrong direction. It is fixed in the last push.

PR #451, opened 2017-01-13
Issue 449

koheiw:


koheiw:
IDs of ngrams can be different in parallel computation, because they are serial numbers (correspondingly the order of types), but the order of the characters tokens should be the same.

PR #462, opened 2017-01-14
Issue 459

koheiw:
This branch adds functions specifying levels in a hierarchical dictionary to *_lookup, and an interface for  the YAML format. Interfaces for other dictionary formats are unchanged.


kbenoit:
Looks very good. We'd need to add the levels functionality to `dfm_lookup` too. 

I'll experiment with this a bit and add a few tests for more hierarchy in a YAML dictionary.

koheiw:
By the way, `dfm_lookup` has the levels functionality already.

koheiw:
The AppVeyor error appeared a real bug in parallel ngram generation.

```r
txt <- c(one = toLower("Insurgents killed in ongoing fighting."),
         two = "A B C D E")
toks <- tokenize(txt, removePunct = TRUE)
toksh <- tokens(txt, removePunct = TRUE)

for(i in 1:1000) {
    cat(i, "\n")
    out1 <- tokens_ngrams(toks, n = 2:3)
    out2 <- tokens_ngrams(toksh, n = 2:3)
    testthat::expect_equivalent(as.list(out1),
                                as.list(out2))
}
attr(out1, 'types')
attr(out2, 'types')

**Comment by kbenoit** [comment]
I'm just catching up on this thread - looks like some tough issues related to the ngrams generator. I suspect this is what was causing our intermittent, unfathomable earlier failures on the test-tokens_ngrams.R line 40 too. 

Please do add the 1000 iteration loop test above to `test-tokens_ngrams.R`.

**Comment by koheiw** [comment]
Exactly. Good to find the real cause, but this is a tough one. This will take sometime to solve.

**Comment by kbenoit** [comment]
OK - once we have this ironed out we should resubmit to CRAN, since the 10,000 document limit in the v0.9.9-3 release is already causing some users problems.

**Comment by koheiw** [comment]
Mutex prevented duplication of ngram ID, and the error is not found in 10000 loops. tbb::spin_lock does not crash the program either (tthread::mutex recommended by the RcppParallel package does) .

**Comment by koheiw** [comment]
Windows tests does not proceed for some reason. Is the repeated test too much?

**Comment by kbenoit** [comment]
@koheiw I just added some tests and fleshed out the documentation, and pushed to the branch, but I got an error when I tried to run Check on my system.

```r
> dfm(data_corpus_inaugural)
Document-feature matrix of: 57 documents, 9,174 features (91.6% sparse).
> dfm(data_corpus_inaugural, ngrams = 2)
 Show Traceback
 
 Rerun with Debug
 Error in validObject(r) : 
  invalid class “dgTMatrix” object: length(Dimnames[1]) differs from Dim[1] which is 62961 
```

❓

kbenoit:
Here are some edge cases - we might consider whether the behaviour is what we expect.

Setup:
```r
txt <- c(d1 = "The United States has the Atlantic Ocean and the Pacific Ocean.",
         d2 = "Britain and Ireland have the Irish Sea and the English Channel.")
toks <- tokens(txt)
dict <- dictionary(list('US'=list(
    Countries = c("States"),
    oceans = c("Atlantic", "Pacific")),
    'Europe'=list(
        Countries = c("Britain", "Ireland"),
        oceans = list(west = "Irish Sea", east = "English Channel"))))
dict
# Dictionary object with 7 key entries.
# - US:
#   - Countries: States
#   - oceans: Atlantic, Pacific
# - Europe:
#   - Countries: Britain, Ireland
#   - oceans:
#     - west: Irish Sea
#     - east: English Channel
```

**Equivalent sub-levels:**

In this case, because level 2 has the same names, we are considering level 2 equivalent even though they are in different level 1 categories.
```r
tokens_lookup(toks, dict, levels = 2)
# tokens from 2 documents.
# d1 :
# [1] "Countries" "oceans"    "oceans"   
# 
# d2 :
# [1] "Countries" "Countries" "oceans"    "oceans"  
```
If we do not agree with this, then we should only make lower levels possible if they also include higher levels. Which means instead of having a vector, we would just give a "maximum level" integer through the `levels` argument, which would produce for `tokens = 2 ` the equivalent of:

```r
tokens_lookup(toks, dict, levels = 1:2)
# tokens from 2 documents.
# d1 :
# [1] "US.Countries" "US.oceans"    "US.oceans"   
# 
# d2 :
# [1] "Europe.Countries" "Europe.Countries" "Europe.oceans"    "Europe.oceans"  
```

**padding added for lower levels, sometimes:**

```r
tokens_lookup(toks, dict, levels = 3)
# tokens from 2 documents.
# d1 :
# [1] "" "" ""
# 
# d2 :
# [1] ""     ""     "west" "east"
```
To be consistent with `levels = 2` above, there should be no pads.

**no error when `levels` exceeds the lowest level of a dictionary:**

```r
tokens_lookup(toks, dict, levels = c(1,4))
# tokens from 2 documents.
# d1 :
# [1] "US" "US" "US"
# 
# d2 :
# [1] "Europe" "Europe" "Europe" "Europe"
```
is okay but 
```r
tokens_lookup(toks, dict, levels = 4)
# tokens from 2 documents.
# d1 :
# [1] "" "" ""
# 
# d2 :
# [1] "" "" "" ""
```
should produce a warning (at least) and should not produce pads, it should return `character(0)` in each list.

kbenoit:
Padding issue also happens in `dfm_lookup`:

```r
txt <- c(d1 = "The United States has the Atlantic Ocean and the Pacific Ocean.",
         d2 = "Britain and Ireland have the Irish Sea and the English Channel.")
dict <- dictionary(list('US'=list(
    Countries = c("States"),
    oceans = c("Atlantic", "Pacific")),
    'Europe'=list(
        Countries = c("Britain", "Ireland"),
        oceans = list(west = "Sea", east = "Channel"))))
testdfm <- dfm(txt)
dfm_lookup(testdfm, dict, levels = 4)
# Document-feature matrix of: 2 documents, 5 features (0% sparse).
# 2 x 5 sparse Matrix of class "dfmSparse"
#     features
# docs          
#   d1 1 2 0 0 0
#   d2 0 0 2 1 1
```

koheiw:
This is ready to be merged to the master. the fix for ngram related issue #469 is in the `issue-469` branch.

koheiw:
No it does not.

PR #475, opened 2017-01-18
textstat_keyness is added 

koheiw:
Please modify the dfm to accept external vector for the groups argument. Then, the example should work.


koheiw:
Hand this branch over to @kbenoit. Please work on dfm a bit.

kbenoit:
See the new RFM.ddm

koheiw:
What is RFM.ddm?

kbenoit:
That's what happens when you type on an iPhone with cold fingers while at a parent-teacher meeting!

it's `dfm.dfm()`.  See `?dfm`, last examples.

koheiw:
Ready to be merged.

koheiw:
Why does codecov stop in this branch?

kbenoit:
Interestingly, `stats::chisq.test` does not seem to compete the scores correctly for small tables - but this warrants investigation.

```r
    ## stats::chisq.test is wrong for small tables
    mat <- matrix(c(3, 2, 14, 10), ncol = 2)
    chi <- stats::chisq.test(mat)
    ## Warning message:
    ## In stats::chisq.test(mat) : Chi-squared approximation may be incorrect
    
    # from the function
    chi$statistic
    ##    X-squared 
    ## 1.626059e-31 
    
    # as it should be (with Yates correction)
    sum((abs(chi$observed - chi$expected) - 0.5)^2 / chi$expected)
    ## [1] 0.1851001
```

kbenoit:
@koheiw see commit message for e00d92e and please review.

koheiw:
`keyness_chi2_dt` looks great. It must be much much faster than `keyness_chi2_stat`.

kbenoit:
Much faster but also I think more correct, inasmuch as there is any accurate computation of chi2 for tables with counts < 5.

I'm implementing a Fisher exact test now using an approach like yours but it's slow. I could compute it directly but the factorials involved would blow it up. I'll put it on the to-do list to examine how R computes `stats::fisher.test()`. From the R code it looks like there are some underlying C calls.

Issue #480, opened 2017-01-19
tokens_lookup leaves attr(*, 'padding')=TRUE even when exclusive=TRUE


koheiw:
This a bug in tokens_lookup, but I also wonder why there is no `featnames()` for tokens.
```r
toks <- tokens(c('a b c d', 'A B C D'))
featnames(toks) # does not exist
toks <- tokens_remove(toks, 'd', padding = TRUE) # toks@padding becomes TRUE
keys <- tokens_lookup(toks, dictionary(list(alpha=c('A', 'B'))), exclusive = TRUE)
attr(keys, "padding") # TRUE
featnames(dfm(keys)) # has empty token
```

kbenoit:
No `featnames.tokens()` because tokens objects don't have "features" - that is an element of a dfm only. tokens objects have just tokens, types, and documents (docnames).

PR #483, opened 2017-01-19
Issue 480

koheiw:
Just a small fix for #480 with tests.


koheiw:
OK. `featnames.tokens()` is removed. Exported `types()` is good, but I am not sure if it should return null tokens, when `padding=TRUE`, because `types()` is used internally as well. Let's leave it and fix the bug first.

Issue #485, opened 2017-01-20
regex2fixed can't handle characters that need to be escaped


kbenoit:
I am trying to use the LIWC as a dictionary, which includes values that are emoticons containing characters like `(` that need to be escaped in regular expressions, e.g.
```
- netspeak: ;), :(, :), :/, (;, (:, ):, /:, 4ev*, abt, af, afk, ahh*, aight,
```

This causes a problem when I try to lookup using a dictionary containing such values:

```r
tokens_lookup(tokens("Some generic text"), dictionary = list(happy = "word"))
# tokens from 1 document.
# Component 1 :
# character(0)

tokens_lookup(tokens("Some generic text"), dictionary = list(happy = ":)"))
#  Error in stringi::stri_detect_regex(types_search, regex) : 
#   Incorrectly nested parentheses in regexp pattern. (U_REGEX_MISMATCHED_PAREN) 
```

And note that this does not fail when `valuetype = "fixed"`:
```r
tokens_lookup(tokens("Some generic text"), dictionary = list(happy = ":)"), valuetype = "fixed")
# tokens from 1 document.
# Component 1 :
# character(0)
```

but does when `valuetype = "glob"`. Earlier code changed valuetype to fixed when a "glob" had no `*` or `?` characters, but the new function does not do this.

kbenoit:
We don't want the user to have to do this, since we want to preserve the ability to import directly glob formatted dictionaries such as the LIWC and Wordstat.

So perhaps replacing the patterns with `utils::glob2rx(quanteda:::escape_regex(features))`is enough?

Note: We should make it clear then that we only support the glob subset `*` and `?`, not `!`, `[`, or `(`.

koheiw:
It there are not such chars like `[` and `(` it should be fine. Let's see.

kbenoit:
Can you code the fix and add a couple of tests?

koheiw:
Working on.

PR #489, opened 2017-01-20
Issue 485

koheiw:
Fix for #485.


koheiw:
Done with emoji, hopefully.

Issue #497, opened 2017-01-22
Add exclusive but ordered mode to tokens_lookup


koheiw:
Extensively using `tokens_lookup` with multi-word dictionaries, I noticed that we need a mode that keeps the order but only returns the keys. Currently when `exclusive=TRUE`, it sorts the keys by ID; when `exclusive=FALSE`, it returns other tokens as well.
```r

txt <- c(d1 = "The United States is bordered by the Atlantic Ocean and the Pacific Ocean.",
         d2 = "The Supreme Court of the United States is seldom in a united state.")
toks <- tokens(txt)

dict <- dictionary(list(Countries = c("States"),
                        oceans = c("Atlantic", "Pacific"),
                        court = c("Supreme Court")))

tokens_lookup(toks, dict, exclusive = TRUE, capkeys = TRUE) # COURT is the last in d2
tokens_lookup(toks, dict, exclusive = FALSE, capkeys = TRUE) # COURT is the first in d2
```
It is also a bit dangerous to not keeping the order of tokens by default (I walked into this trap my self). It would be better to add `overlap` argument for detecting overlapping at the expense of the original orders of tokens.

koheiw:
Now `tokens_lookup` retains the order of tokens in the keys. When multiple entries match tokens in the same (starting) position, keys of those entries are ordered by the key IDs.

PR #499, opened 2017-01-22
Issue 497

koheiw:
This is an attempt to solve #497 by keeping the order of tokens as much as possible when `exclusive=TRUE`. The impact of the speed would be small, but need to be checked with large datasets.

### DIFF
--- a/src/quanteda.h
+++ b/src/quanteda.h
@@ -94,15 +94,14 @@ namespace ngrams {
 
 #if RCPP_PARALLEL_USE_TBB
     // TBB header is loaded automatically by the macro
-    typedef tbb::atomic<int> IdNgram; // tbb::atomic does not suppport unsigined int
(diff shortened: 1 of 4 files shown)


koheiw:
There is no impact on the speed. Much of the time is for `regex2id` in both cases...
```r
load("/home/kohei/Documents/Brexit/Analysis/data_corpus_guardian.RData")
toks <- tokens(data_corpus_guardian)

dict <- dictionary(list(country = "united states", 
                        law=c('law*', 'constitution'), 
                        freedom=c('free*', 'libert*')))
system.time(
    tokens_lookup(toks, dict, valuetype='glob', verbose=FALSE),
)

# New
   user  system elapsed 
 64.704   0.304  24.466 

# Old
   user  system elapsed 
 76.368   1.008  24.774
```

PR #503, opened 2017-01-23
Issue 502

koheiw:
With this fix, `tokens_lookup` do not double count the countries with these entries:

```
- CZ: Czech Republic, Czech*
- US: United States, United States of America
```


kbenoit:
From the LIWC 2015 Operator Manual:

> Be careful in the use of asterisks to avoid misclassifying words. For example, the following case is problematic:
> NOT OK:  `Thank* 27 Thanksgiving 94`
> In this case, the word “Thanksgiving” would only be counted in category 27 since LIWC2015 does not double­-count words. The asterisk at the end of `Thank*` will capture all words that begin with `Thank`.

I actually think this is wrong behaviour, and that we should double-count words. So the correct behaviour ought to be what it currently is (from this branch):

```r
dict <- dictionary(list("27" = "Thank*", "94" = "Thanksgiving"))
tokens_lookup(tokens("Thank goodness for Thanksgiving!"), dictionary = dict)
## tokens from 1 document.
## Component 1 :
## [1] "27" "27" "94" 
```

I'd rather make the user be responsible for not overlapping dictionaries. To help with this, we could run a check at the time of dictionary declaration that finds patterns that match other patterns and report on this, and also add the double-counting as an option. But if we disable double-counting, then we need clear rules for which category gets counted and which does not.

kbenoit:
Bug found: order seems to matter with multi-word values.

```r
tokens_lookup(tokens("The Czech Republic played the United States of America in football."),
              dictionary(list(CZ = c("Czech Republic", "Czech*"))))
## tokens from 1 document.
## Component 1 :
## [1] "CZ"

tokens_lookup(tokens("The Czech Republic played the United States of America in football."),
              dictionary = list(CZ = c("Czech*", "Czech Republic")))
## Error in if (concatenator != concatenator_dict) entries <- stringi::stri_replace_all_fixed(entries,  : 
##   argument is of length zero
```

Works ok for single word values:
```r
tokens_lookup(tokens("The Czech Republic played the United States of America in football."),
              dictionary(list(Rep = c("Rep*", "Republic"))))
## tokens from 1 document.
## Component 1 :
## [1] "Rep"
tokens_lookup(tokens("The Czech Republic played the United States of America in football."),
              dictionary(list(Rep = c("Republic", "Rep*"))))
## tokens from 1 document.
## Component 1 :
## [1] "Rep"
```

And that small behaviour applied to `dfm_lookup()`:
```r
dfm_lookup(dfm("The Czech Republic played the United States of America in football."),
              dictionary(list(Rep = c("Republic", "Rep*"))))
## applying a dictionary consisting of 1 key
## Document-feature matrix of: 1 document, 1 feature (0% sparse).
## 1 x 1 sparse Matrix of class "dfmSparse"
##        features
## docs    Rep
##   text1   1

dfm_lookup(dfm("The Czech Republic played the United States of America in football."),
           dictionary(list(Rep = c("Rep*", "Republic"))))
## applying a dictionary consisting of 1 key
## Document-feature matrix of: 1 document, 1 feature (0% sparse).
## 1 x 1 sparse Matrix of class "dfmSparse"
##        features
## docs    Rep
##   text1   1
```
(and notice the absence of double-counting already in `dfm_lookup()`.

koheiw:
If you use `dictionary`, it woks. Should the function accept a list? The error is due to the lack of `concentrator` attribute in the list.
```r
tokens_lookup(tokens("The Czech Republic played the United States of America in football."),
              dictionary = dictionary(list(CZ = c("Czech*", "Czech Republic"))))
```

kbenoit:
Totally right, my syntax on the call above was wrong. It has to be a dictionary object not a regular list.

koheiw:
What is this warning?
```
kpathsea: Running mktexfmt pdflatex.fmt
mktexfmt [WARNING]: hidden sys mode found, switching to sys mode.
mktexfmt: mktexfmt is using the following fmtutil.cnf files (in precedence order):
mktexfmt:   /home/travis/texlive/texmf-dist/web2c/fmtutil.cnf
```

koheiw:
The name of the new argument was changed  `overlap` to `distinctive`. `overlap` shall be used in the same way as in the new `tokens_compound` in other PR.

koheiw:
The original intention for the `distinct` was avoid counting the **same features** for the **same key**, for example:
```
tokens_lookup(tokens("A B C"), dictionary(list(key = c("A B", "a b"))), distinctive = TRUE)
## tokens from 1 document.
## Component 1 :
## [1] "key"
```
so this is not a bug because "B C" and "A B C" match **different features**.
```
tokens_lookup(tokens("A B C"), dictionary(list(key = c("B C", "A B C"))), distinctive = TRUE)
## tokens from 1 document.
## Component 1 :
## [1] "key" "key"
```
However, I can change to one key for any features, returning one "key" in the second case. I think this is a matter of the level of distinctiveness or uniqueness, but I do not think it is so common that the people want binary results from `tokens_lookup`.

This does not return two "key" simply because `regex2id` removes duplicated sequences with `unique()`. This is easy to solve but cannot image a situation when people need to add duplicated entities to the same key.
```
tokens_lookup(tokens("A B C"), dictionary(list(key = c("A B", "A B"))), distinctive = FALSE)
## tokens from 1 document.
## Component 1 :
## [1] "key"
```

kbenoit:
But is this not inconsistent:

```r
tokens_lookup(tokens("A B C"), dictionary(list(key = c("A B", "A B C"))), distinctive = TRUE)
## tokens from 1 document.
## Component 1 :
## [1] "key"

tokens_lookup(tokens("A B C"), dictionary(list(key = c("B C", "A B C"))), distinctive = TRUE)
## tokens from 1 document.
## Component 1 :
## [1] "key" "key"
```

koheiw:
Agree. It should return one `key`, but a bit difficult to implement. I will try.

koheiw:
How about this?
```r
# Detect all the overlapped entries
tokens_lookup(tokens("A B C"), dictionary(list(key = c("B C", "A B C"))), overlap = TRUE)
# tokens from 1 document.
# Component 1 :
# [1] "key" "key"

# Nested entries are overlap
tokens_lookup(tokens("A B C"), dictionary(list(key = c("B C", "A B C"))), overlap = FALSE)
# tokens from 1 document.
# Component 1 :
# [1] "key"

# Non-nested entries are also overlap
tokens_lookup(tokens("A B C D"), dictionary(list(key = c("C D", "A B C"))), overlap = FALSE)
# tokens from 1 document.
# Component 1 :
# [1] "key"

# Entries for different keys are not treated as overlap, preserving the order of tokens 
tokens_lookup(tokens("A B C"), dictionary(list(key1 = "B C", key2 = "A B C")), overlap = FALSE)
# tokens from 1 document.
# Component 1 :
# [1] "key2" "key1"

```

kbenoit:
I like that new behaviour. So

```r
tokens_lookup(tokens("A B C D"), dictionary(list(key = c("C D", "A B C"))), overlap = TRUE)
```

Will generate two matches of key?

koheiw:
True, like this
```r
tokens_lookup(tokens("A B C D"), dictionary(list(key = c("C D", "A B C"))), overlap = TRUE)
# tokens from 1 document.
# Component 1 :
# [1] "key" "key"

PR #504, opened 2017-01-23
Logo

koheiw:
I tried to use SVG, because it is better on high resolution screen, but no luck...


koheiw:
I finally managed to use SVG in README.md on the GitHub page. There are three things to note: 
- GitHub strips all the CSS directives
- SVG should not be responsive (becomes too big)
- SVG should be linked through RawGit (otherwise GitHub treats SVG as plain text file).

PR #522, opened 2017-01-26
Issue 517

koheiw:
Address #517


kbenoit:
### Thoughts

This is a tricky issue that overlaps with PR #503, since it also involves the mutual exclusivity of sequences that match and/or overlap one another.

I thought the behaviour desired was to combine lower-length sequences into longer-length sequences, when found adjacent, e.g.
```r
tokens_compound(tokens('A B C D'), seqs <- c('A B', 'B C'))
# tokens from 1 document.
# Component 1 :
# [1] "A_B_C"                 "D"
```

Instead the function as it is now gives:
```r
tokens_compound(tokens("A B C D"), c("A B", "B C"), overlap = TRUE)
# tokens from 1 document.
# Component 1 :
# [1] "A_B" "B_C" "D"  

tokens_compound(tokens("A B C D"), c("A B", "B C"), overlap = FALSE)
# tokens from 1 document.
# Component 1 :
# [1] "A_B" "C"   "D"  
```

The way I see it, this is not really "overlap" as much as it is similar to what you programmed for `tokens_lookup` and the `distinctive = FALSE` option (see #503). It counts value matches in the same "key" (here, always a single key/sequence) once for any sequence match, regardless of how much duplication this causes.

### Recommendations

1. The current behaviour of `overlap = FALSE` is too arbitrary.

The compounding is arbitrary:
```r
pairs <- c("a b", "b c", "c d", "d e")
# tokens_compound(tokens("a b c d e"), pairs, overlap = FALSE)
# tokens from 1 document.
# Component 1 :
# [1] "a_b" "c_d" "e"  
```

The compounding seems to form the first match from all sequences regardless of order, so in this expression:
```r
tokens_compound(tokens("Z B C D"), c("B C", "Z B"), overlap = FALSE)
# tokens from 1 document.
# Component 1 :
# [1] "Z_B" "C"   "D" 
```
the `"Z B"` is still concatenated first because it occurs first in the tokens (and not because of the alphabetical ordering of the sequences clearly).

**Alternative:** 

The default behaviour should become:
```r
tokens_compound(tokens("a b c d e"), c("a b", "b c", "c d", "d e"), overlap = FALSE)
# tokens from 1 document.
# Component 1 :
# [1] "a_b_c_d_e"  
```

2. The `overlap = TRUE` is ok.

With `overlap = TRUE`: (and this is the current branch behaviour)
```r
tokens_compound(tokens("a b c d e"), c("a b", "b c", "c d", "d e"), overlap = FALSE)
# tokens from 1 document.
# Component 1 :
# [1] "a_b" "b_c" "c_d" "d_e"
```

3. We should try to come up with a name that is more descriptive than `overlap`.

    If we use my own suggested default behaviour, it could be `join = TRUE` for instance.

4. We should consider the same nesting issue as per my comments on PR #503, related to the "strictness" of nesting. So for instance the first below "double-counts" because the first sequence is nested inside the second, but we do not consider in the second instance that they are nested.
```r
tokens_compound(tokens("a b c d e"), c("a b", "a b c"), overlap = TRUE)
# tokens from 1 document.
# Component 1 :
# [1] "a_b_c" "a_b"   "d"     "e"    

tokens_compound(tokens("a b c d e"), c("a b", "a b"), overlap = TRUE)
# tokens from 1 document.
# Component 1 :
# [1] "a_b" "c"   "d"   "e" 
```

koheiw:
It is easy to implement the first alternative behavior. The question is how to generate a pattern `c("a b c d e", "a b c d", "a b c", "a b", "b c d e", "b c d", "b c", "c d e", "c d", "d e")` from `c("a b", "b c", "c d", "d e")` in the R side? Is this a question for the SO community?

kbenoit:
Solution probably involves figuring out the longest linked sequence, then using some recursion to generate subsequences.

But there might be a simpler solution, like forming new patterns of adjacent pairs with wildcards, e.g.

`c"A B", "B C")`

generates new pattern (here, with regex matches):

`c("A[<concatenator>]B C", `"A B[<concatenator>]C")`

and then calling `tokens_compound` a second time to join the patterns already joined with the new compound patterns.

koheiw:
Here is `token_compound()` with `join` functionality:
```r
tokens_compound(tokens("a b c d e x y z"), c("a b", "b c", "c d", "d e"), join = TRUE)
## tokens from 1 document.
## Component 1 :
## [1] "a_b_c_d_e" "x"         "y"         "z"        

tokens_compound(tokens("a b c d e x y z"), c("a b", "b c", "c d", "d e"), join = FALSE)
## tokens from 1 document.
## Component 1 :
## [1] "a_b" "b_c" "c_d" "d_e" "x"   "y"   "z" 
```
There is an option to set `overlap = FALSE` when `join = FALSE`, but hidden from the users at the moment. Join functionality is implemented on the C++ side in a similar way as `tokens_ngrams`.

kbenoit:
@koheiw please resolve the conflict and I will accept and merge.

Issue #524, opened 2017-01-28
Question: Using quanteda for co-occurence tests


methodds:
Hi,
this is not a bug report but rather a question on how you would use quanteda to solve a simple problem:

Let's say that I have a corpus and use a dictionary to divide it into two parts, e.g. "migration-related" and "not migration-related".

I then find the 10 most frequent  terms (other than those from the dictionary) in the "migration-related" part and would like to know if these terms are significantly (based on a proper test) more often contained within "migration-related" documents than in "not migration-related" documents.

What would be the best way to do this? Do you have a suggestion and maybe even a small example?

Thanks and Cheers
Carsten

kbenoit:
In the newest version (now on CRAN) we added `textstat_keyness()`. You would need to partition the corpus first using `dfm(yourcorpus, groups = yourmigrationgroupingvariable)` first. 

We're very interested to hear how this works for you.

methodds:
I'll check it out and report back to you, thanks =)

methodds:
Okay I just tried the examples in the manual and it worked very well. I will test it on another corpus soon.

Two thoughts: 
- The ``exact`` test is rather slow. is it computational intensive? Also in this context, I assume the smoothing from the example is necessary so that there are no 0's for the fraction denominators ?
- Would you consider adding a parameter that would also return p-values, e.g. for the chi² tests? Personally I don't like them, but I guess they are still widely used in our discipline.

kbenoit:
Yes the exact test is slow because literally it has to compute the exact combinatorics of the test once for each feature.

Easy to recover p-values for the Chi-squared test or CIs for the lr test. Will add these soon.

kbenoit:
I've moved this to #495 and created a branch `issue-495` with these changes. 

You can try installing it with:

```r
devtools::install_github("kbenoit/quanteda", ref = "issue-495")
```

and let us know what you think.

Issue #465, opened 2017-01-16
Review sequences algorithm


koheiw:
The formula presented in Blaheta, D., & Johnson, M. (2001) "Unsupervised learning of multi-word verbs" seems to have atheoretical and unnecessary terms. Check empirically how these terms are affecting.

koheiw:
Removal of mysterious `(n - 1)` weight does not affect the result almost at all, so it should be removed.

```r
seqs_both <- merge(seqs, seqs_old, by='sequence')
cor(seqs_both$lambda.x, seqs_both$lambda.y)
[1] 1
cor(seqs_both$sigma.x, seqs_both$sigma.y)
[1] 0.9466666
cor(seqs_both$z.x, seqs_both$z.y)
[1] 0.9998167
```

PR #525, opened 2017-01-28
Issue 495

kbenoit:
This implements improvements to `textstat_keyness`. Addresses issues #465 and a request in issue #524.  The return is now a data.frame with rownames matching features, and an added `p` column for the p-values associated with each test.  This makes `textstat_keyness` similar in output to the other `textstat_*` functions.


koheiw:
Addition of p-value looks good. I was thinking of adding `reference` that allows entire dfm to be the reference, but it cannot be done with dfm's grouping function, so I will postpone (I do not think I need the function now).

PR #543, opened 2017-02-05
Issue 510

koheiw:
`qutd_cpp_tokens_recompile()` is isolated be merged onto master.


koheiw:
I see `1. Failure: as.corpus.corpuszip works (@test-corpus-compress.R#13) ` and package installation errors in the log, but C++ recompiler should be fine.

koheiw:
Worth restarting Travis.

kbenoit:
@koheiw this looks fine now, so merge it into master when you are ready. Might be worth adding tests for the exceptions you addressed in 93c314d and a0a5eeb

koheiw:
I added some tests of robustness. If this commit passes the test, all done.

PR #548, opened 2017-02-08
Issue bitshift

koheiw:
Please test this.


kbenoit:
test how?

koheiw:
May be using your rhub key. If you think its OK, please try CRAN again.

koheiw:
What the most strict test we can access?

Issue #508, opened 2017-01-24
Carry docvars inside tokens and dfm objects


kbenoit:
This would allow selected docvars to be included in tokens and dfm objects, and appropriately selected following any indexing or subsetting etc functions that trim documents from the tokens and dfm objects.

Purpose: to make dfm-based operations that need docvars easier to access.

kbenoit:
Done for tokens in https://github.com/kbenoit/quanteda/commit/8ba266b4426afab4e1174439f275b0d9f4059ea7. 

Need to add the same for `dfm()`.

kbenoit:
Done now for dfm too.

PR #566, opened 2017-02-24
Issue 508

koheiw:
Upgrade for #508. Document variables are passed to `tokens`, but not yet to `dfm`. I am happy without it, but please add this part if you want.


kbenoit:
Looks good in general. I will disable the docvars<- method for tokens however, since this should be avoided.

koheiw:
We might need `docvars<-` sometime, but you can disable it if you think so.

kbenoit:
Take a look if you want - it's pretty consistent now and while the behaviour has only changed in that I left the `docvars<-.tokens` internal only, the code is much cleaner and more consistent. Will be easy to extend this now.

koheiw:
It is great to implement document-level meta data consistently, but saving all the variable in the same data frame making the code unnecessarily complex. How about separating them into `docvars` and `metavars`, and create a generalized accessor?

kbenoit:
well, it's no different from the way it was before, where a "metadoc" variable was just a docvar with an underscore. In practice I think almost no one uses this, although some functions such as `corpus_reshape(x, to = "sentences")` record the document and sentence number as metadoc variables. We don't want these exposed as standard document variables because they are not really attributes of the documents or document sources, but rather meta-data about the documents.

On your question:  `get_docvars()` is the generalized assessor in the new `docvars.R`. That was the biggest structural change in the code.

koheiw:
I understand that there are user docvars and internal docvars, and that internal docvars have underscored names. If they are in the data frame, we always need to check if fields are underscored, but we do not need to do anything if internal docvars are in a separate data frame. This is what I meant. We can do this by only adding `internal` argument to `docvars()` based on which it switches to `attr(x, 'metavars')`.

PR #568, opened 2017-02-26
Dev recompile

koheiw:
This branch integrates tokens recompiling function into `tokens_*` functions in the C++ to cut the conversion overhead between C++ and R. This reduces the run time by around 20%.


kbenoit:
Does this make any existing functions unnecessary? E.g. `reassign_attributes()`?

kbenoit:
Also we need to do heavy duty cross-platform testing before accepting the C++ changes into master.

koheiw:
`quanted::attributes()` does the same thing as `reassign_attributes()`, but the C++ recompiler still needs reassignmening attributes. I kept `reassign_attributes()` because I am not sure how it is used for objects other than `tokens`.

koheiw:
There is no change in platform sensitive part like TBB or math, but I can do extra tests. What should be the cross-platform test?

kbenoit:
Cross-platform tests would be me running this on:
* OS X El Capitan
* macOS
* Windows 10

and you could try it on:
* Solaris x86

Can we do UBSan tests using rHub?

koheiw:
It builds successfully on my Windows 10 and rHub:
https://builder.r-hub.io/status/quanteda_0.9.9-28.tar.gz-5239ce22959c4b43afee739cb72070f8

Testing with a Solaris does not make sense, because the set up of CRAN's Solaris is not specified (my Solaris is not yet working).

PR #575, opened 2017-03-03
Dev wordfish sparse

HaiyanLW:
1) Rename the existing file `wordfish.cpp` into a function called `wordfish_dense.cpp`
2) Create a sparse version `wordfish.cpp` using Armadillo sparse matrix to handle `dfm` 
3) Add an option to `textmodel_wordfish()` called `sparse = TRUE` where if `FALSE`, it calls  `wordfishcpp_dense()`. When `sparse = TRUE`, `textmodel_wordfish()` will call the new `wordfish.cpp` 
4) Add an option to `textmodel_wordfish()` called `svd_sparse = FALSE` , where if `TRUE`  `svds()`, which only calculates the top singular vectors, is applied to initialize the starting values of theta. When `svd_sparse = FALSE`, the starting values of theta is set without applying `svds()`. This option only applies to the new `wordfish.cpp` 
5) Accordingly, add `residual_floor = 0.0` to define the minimium threshold for the residual matrix which is used in `svds()`. Larger values leads to more sparse residual matrix and hence less memory space and less running time. This option only applies to the new `wordfish.cpp` 
6) Add an option called `abs_err = FALSE`,   where if `TRUE` absolute value of the convergence error is used. This option is set to allow the inspection of the converging.
7) Add some in process printing to allow the inpection of the converging.


kbenoit:
Some good news: it's significantly faster.

```r
    sparse = textmodel_wordfish(ie2010dfm, dir = c(6,5), sparse = TRUE),
    old = textmodel_wordfish(ie2010dfm, dir = c(6,5), sparse = FALSE), 
    unit = "relative", times = 30
)
##   expr    min       lq    mean   median       uq      max neval cld
## sparse 1.0000 1.000000 1.00000 1.000000 1.000000 1.000000    30  a  
##    old 2.6088 2.733463 2.71565 2.750114 2.723566 2.679322    30   b

**Comment by HaiyanLW** [comment]
Wait to see the parallel version :)

**Comment by kbenoit** [comment]
@conjugateprior Would be great if you look at this too.

File: src/wordfish.cpp [hunk 1]
```diff
@@ -171,19 +206,20 @@ Rcpp::List wordfishcpp(SEXP wfm, SEXP dir, SEXP priors, SEXP tol, SEXP disp, SEX
...
+        for (std::size_t i=0; i < N; i++){
+            for (std::size_t k=0; k < K; k++){
                 loglambdaik = alpha(i) + psi(k) + beta(k)*theta(i);
-                lp = lp + loglambdaik*Y(i,k)-exp(loglambdaik);
+                lp = lp + loglambdaik*wfm(i,k)-exp(loglambdaik);
             }
         }
-        // Rprintf("%d: %f2\\n",outeriter,lp);
-        
+        Rcout<<"outeriter="<<outeriter<<"  lp - lastlp= "<<lp - lastlp<<std::endl;
```
HaiyanLW: Yes, of course.

HaiyanLW: I just remembered I kept that line for you and @conjugateprior to review the convergence.

HaiyanLW:
Tested the sparse code on another dataset, 
```usdfm <- dfm(data_corpus_inaugural, verbose = FALSE)
    wfm_d <- textmodel_wordfish(usdfm, dir = c(6,5), sparse = TRUE, svd_sparse = TRUE, residual_floor = 0.5)
    wfm_s <- textmodel_wordfish(usdfm, dir = c(6,5), sparse = FALSE)
    cc<-cor(wfm_d@theta, wfm_s@theta)
    expect_gt(cc, 0.99)```
it works (against `dense` or `austin` package) with starting thetas constructed by `svds()` (on sparsified `residual matrix`). It also works if I made the `residual matrix` more sparse (i.e. `residual_floor =1 or 2`)

However, ` textmodel_wordfish(usdfm, dir = c(6,5), sparse = TRUE)` didn't work without applying `svds()`--seems the thetas are still need a reasonable starting value. 

The sparse code works on `data_corpus_irishbudget2010` with or without applying `svds()`  -- Probably' dir = c(6,5)' also matters

**Comment by kbenoit** [comment]
@conjugateprior @lauderdale any thoughts on this branch and the approach to starting values??

**Comment by conjugateprior** [comment]
The Goodman 1979 / Becker algorithm that I think Ben's code implements (I haven't looked too carefully so I might be wrong) is particularly sensitive to bad start values in my experience. Not that that helps much, I'm afraid. 

iirc the Proksch and Slapin code fits a sequence of linear model on the log counts, which might be better. Probably slower, but could also be parallelized.

**Comment by kbenoit** [comment]
@conjugateprior Any chance you could try some tests with the sparse v. dense versions and see if this makes any difference in an application? We noticed very minor differences but still r >= .99.

**Comment by kbenoit** [comment]
@HaiyanLW I'm going to accept this but first can you fix one small "bug": When there are unsmoothed features, the default sparse approach produces NAs. The solution is probably to use some form of smoothing, to adjust the  approximations based on the marginals you are using to get starting values?

```r
> textmodel_wordfish(data_dfm_LBGexample, dir = c(1,5))
Fitted wordfish model:
Call:
	textmodel_wordfish(data = data_dfm_LBGexample, dir = c(1, 5))

Estimated document positions:

  Documents theta  SE lower upper
1        R1   NaN NaN   NaN   NaN
2        R2   NaN NaN   NaN   NaN
3        R3   NaN NaN   NaN   NaN
4        R4   NaN NaN   NaN   NaN
5        R5   NaN NaN   NaN   NaN
6        V1   NaN NaN   NaN   NaN

Estimated feature scores: showing first 30 beta-hats for features

 A  B  C  D  E  F  G  H  I  J  K  L  M  N  O  P  Q  R  S  T  U  V  W  X  Y  Z ZA ZB ZC ZD 
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0 
> textmodel_wordfish(data_dfm_LBGexample, dir = c(1,5), sparse = F)
Fitted wordfish model:
Call:
	textmodel_wordfish(data = data_dfm_LBGexample, dir = c(1, 5), 
    sparse = F)

Estimated document positions:

  Documents       theta         SE       lower       upper
1        R1 -1.29056981 0.01448422 -1.31895889 -1.26218074
2        R2 -0.61643907 0.01059613 -0.63720748 -0.59567065
3        R3  0.03350982 0.01085375  0.01223647  0.05478318
4        R4  0.70961244 0.01214106  0.68581597  0.73340892
5        R5  1.52333687 0.01744178  1.48915098  1.55752276
6        V1 -0.35945025 0.01070073 -0.38042368 -0.33847683

Estimated feature scores: showing first 30 beta-hats for features

          A           B           C           D           E           F           G           H           I           J           K           L 
 -8.4452617  -9.3002556 -11.2580501 -12.3643578 -13.3323903 -13.2373464 -13.4867108 -12.3588189 -11.5848271 -10.4239011  -9.2561345  -8.1485428 
          M           N           O           P           Q           R           S           T           U           V           W           X 
 -6.9820863  -5.8782223  -4.7955288  -3.5739210  -2.5522985  -1.3458753  -0.2889734   0.7696623   1.9353486   2.9295159   4.1021898   5.0019413 
          Y           Z          ZA          ZB          ZC          ZD 
  5.9373363   6.8722030   7.7519960   8.7447725   9.6708363  10.4464336 
> textmodel_wordfish(data_dfm_LBGexample, dir = c(1,5), svd_sparse = T)
Fitted wordfish model:
Call:
	textmodel_wordfish(data = data_dfm_LBGexample, dir = c(1, 5), 
    svd_sparse = T)

Estimated document positions:

  Documents       theta         SE       lower       upper
1        R1 -1.29056981 0.01448422 -1.31895889 -1.26218074
2        R2 -0.61643907 0.01059613 -0.63720748 -0.59567065
3        R3  0.03350982 0.01085375  0.01223647  0.05478318
4        R4  0.70961244 0.01214106  0.68581597  0.73340892
5        R5  1.52333687 0.01744178  1.48915098  1.55752276
6        V1 -0.35945025 0.01070073 -0.38042368 -0.33847683

Estimated feature scores: showing first 30 beta-hats for features

          A           B           C           D           E           F           G           H           I           J           K           L 
 -8.4452617  -9.3002556 -11.2580501 -12.3643578 -13.3323903 -13.2373464 -13.4867108 -12.3588189 -11.5848271 -10.4239011  -9.2561345  -8.1485428 
          M           N           O           P           Q           R           S           T           U           V           W           X 
 -6.9820863  -5.8782223  -4.7955288  -3.5739210  -2.5522985  -1.3458753  -0.2889734   0.7696623   1.9353486   2.9295159   4.1021898   5.0019413 
          Y           Z          ZA          ZB          ZC          ZD 
  5.9373363   6.8722030   7.7519960   8.7447725   9.6708363  10.4464336 
```

HaiyanLW:
Ok. Actually I'm thinking of setting `svd_sparse = TRUE` as default, as it produces more consistent results --not just this case, also on both `data_corpus_inaugural` and `ie2010dfm`.

kbenoit:
Oh but it is SO FAST with `svd_sparse = TRUE`, and in my tests including with data_corpus_inaugural and ie2010dfm the results with and without svd_sparse were identical. I'd leave it as is.

HaiyanLW:
En, interesting. In my machine, if I switch the `svd_sparse` off for `data_corpus_inaugural `, it returns
```r
> wfm_s <- textmodel_wordfish(usdfm, dir = c(6,5), sparse = FALSE)
> wfm_d <- textmodel_wordfish(usdfm, dir = c(6,5), sparse = FALSE)
> wfm_s <- textmodel_wordfish(usdfm, dir = c(6,5), sparse = T)
> wfm_s_svd <- textmodel_wordfish(usdfm, dir = c(6,5), sparse = T, svd_sparse = T)
svd done
> cc<-cor(wfm_d@theta, wfm_s@theta)
> expect_gt(cc, 0.99)
Error: `cc` is not strictly more than 0.99. Difference: -1.45
> cc<-cor(wfm_d@theta, wfm_s_svd@theta)
> expect_gt(cc, 0.99)
```

HaiyanLW:
I will leave it as it is now. Maybe we can change the it to `svd_sparse` after I parallalize generating `residual matrix`. By the way, I raised the `residual_floor = 0.5`, so at least it is faster than dense version. 
```r
+     sparse = textmodel_wordfish(ie2010dfm, dir = c(6,5), sparse = TRUE, svd_sparse = T),
+     old = textmodel_wordfish(ie2010dfm, dir = c(6,5), sparse = FALSE), 
+     unit = "relative", times = 3
+ )
svd done
svd done
svd done
Unit: relative
   expr     min       lq     mean   median       uq      max neval
 sparse 1.00000 1.000000 1.000000 1.000000 1.000000 1.000000     3
    old 1.16097 1.161139 1.163296 1.161307 1.164451 1.167565     3
```

kbenoit:
You're right, I am getting very different results from the dense v. sparse methods. (Note that there is no need for the `dir = c(6,5)` here, that is just for the Irish budget cases.)

There appears to be a huge performance hit from `sparse = TRUE, svd_sparse = TRUE` - presumably because the use of SVD for the starting values means we lose the advantage of sparsity anyway. So under what circumstances would we want  to use `sparse = TRUE, svd_sparse = TRUE`?

```r
require(quanteda)
usdfm <- dfm(data_corpus_inaugural, removePunct = TRUE, tolower = TRUE)
    wfm_d = (wfm_d <- textmodel_wordfish(usdfm, sparse = FALSE)),
    wfm_s = (wfm_s <- textmodel_wordfish(usdfm, sparse = TRUE)),
    wfm_s_svd = (wfm_s_svd <- textmodel_wordfish(usdfm, sparse = TRUE, svd_sparse = TRUE)),
    times = 1
)
# Unit: milliseconds
#      expr        min         lq       mean     median         uq        max neval
#     wfm_d  9920.4495  9920.4495  9920.4495  9920.4495  9920.4495  9920.4495     1
#     wfm_s   111.7884   111.7884   111.7884   111.7884   111.7884   111.7884     1
# wfm_s_svd 51704.9840 51704.9840 51704.9840 51704.9840 51704.9840 51704.9840     1

cor(cbind(data.frame(dense = wfm_d@theta, 
                     sparse = wfm_s@theta, 
                     sparse_svd = wfm_s_svd@theta)))
#                dense    sparse sparse_svd
# dense      1.0000000 0.4981304  1.0000000
# sparse     0.4981304 1.0000000  0.4981304
# sparse_svd 1.0000000 0.4981304  1.0000000
```

HaiyanLW:
The main advantage is that the `dfm` stay sparse. The current dataset is probably too small to show that.  
If the `residual_floor` is set to, say `0.5`( and can be even larger), It is slightly faster than the dense version even for these two datasets.  Since we only need the top singular value as the starting values, it entirely makes sense to set the `residual_floor` to only consider principle contributors.

kbenoit:
OK, got it. I just verified the correlation is still > 0.999 with `residual_floor = 0.5`. Let's set the default value of `residual_floor` to 0.5 in the function signature.

PR #591, opened 2017-03-10
Issue589

koheiw:


kbenoit:
Some exceptions thrown from the documentation, but these look easy to fix.

PR #599, opened 2017-03-14
Dev collocaton

koheiw:
This branch is far from complete, but I request a pull, because it is becoming too much for me to work solely. The goal of this branch is to integrate all the collocation related functions under `textstat_collocations()`:

* collocations
* sequences
* fcm
* kwic
* collocations_target

Integration of `collocations` and `sequences` are relatively easy, so I started working on the last three. I added `contexts` function, which is a special type of `tokens` whose documents are contexts. This can be used for target word-based collocation analysis. `kwic`'s shares its underlying C++ code with this.

The relationship between `contexts` and `tokens` is similar to `fcm` and `dfm`, so they can be integrated consistently with `textstat_collocations.fcm(x = fcm, y = dfm)` and  `textstat_collocations.contexts(x = contexts, y = tokens)`. in both functions, the second objects are used to obtain frequencies of non-collocations. I implemented those functions already, although they are more works to be done.

If  `textstat_collocations.fcm` is the function for all-all collocation, what `collocations` should be for? It looks like for fixed length sequences identification (as it was called `phrases`). If so, we can use `sequences` C++ underlying functions to count collocations. This makes the results more consistent.


kbenoit:
OK, following our discussion and having thought about it more, here's how I think we should proceed. My suggestion involves creating new branches from `master`, and drawing from the work in branch `dev-collocaton` in order to implement three related, but independent fixes. Doing this separately allows us to manage the error checking and testing better, as well as split up the work.

1.  A `textstat_collocations()` branch, which creates the new function, hides the existing functions `collocations()` and `sequences()` (using `@keywords internal`), and creates a single API to call all four existing scoring methods (the three currently in `collocations()` and one in `sequences()`).

    We can focus our attention on designing the interface the way we want it, so that we know what we want this function to input and output. 

    Input:
    * character, corpus, tokens objects for now
   
    Output:
    * a data.frame with the sequence of words in one column, the count, and scores in the others.

2.  A redesign of `kwic()` to take in tokens and return tokens, as context() does now. We can add an argument to replace the target word(s) with pads, and to overlap or not. Once again, branch from master, and borrow from `dev-collocaton` to create a new branch and PR. We can continue the design discussion there.

3.  Once 1 and 2 are done, we work on the implementation of how the collocation identification and scoring is done. We leave this for last.

koheiw:
There are documentation/examples related errors but `collocations` and `sequences` are integrated.

koheiw:
Old functions like `selectFeatures` were causing a lot of problems in this integration. I think it is time to remove those.

kbenoit:
Feel free to ignore selectFeatures() when you integrate, but it's probably too complicated to remove just that from this branch. I'd prefer a branch to remove the old ones, to be done separately once the current development branches have been wrapped up.

But soon!

koheiw:
I commented out some of the texts and examples of selectFeatures. What I haven't done is documentation, because details of `textstat_collocations` would change. This is the time for you to try it out!

PR #601, opened 2017-03-15
Fix unicode

koheiw:
This fixes error in `regex2id()` when `case_insensitive = FALSE` in Japanese


kbenoit:
Take a look at this issue https://github.com/kbenoit/readtext/issues/14 which came from https://github.com/kbenoit/quanteda/issues/189. This is worth implementing in a modular way, the question is how to make it modular enough to be useful generally but also finding the appropriate place in text workflow to put it.

For **readtext**, @adamobeng and I were thinking this would be done at the import stage. Default would be nfc but allow the user to set this option if they prefer.

koheiw:
Interestingly, applying `stri_trans_nfc()` before tokenization did not help, so I moved it to `regex2id`.

Issue #612, opened 2017-03-23
Question regarding the new wordfish implementation


methodds:
Hi,

I just tested the new wordfish implementation and it is muuuch faster than before which is totally awesome. Did you test the new version against the old one for corpora that you know and achieved similar results? 
At the moment I'm playing around with wordshoal at lot and as far as I understand, wordshoal directly uses wordfish in the first stage. After now rerunning some wordshoal models with the new wordfish implementation, I can definitely notice some differences. Is this to be expected?

kbenoit:
See #613. If you install that branch, you will be able to pass arguments to the wordfish function that give you a bit more control over the new behaviours.

But remember that `textmodel_wordshoal()` remains quite experimental, until @lauderdale works on the code and tests for this function a bit more!

methodds:
Thanks, I'll check out that branch. And yes, I'm aware of wordshoal being quite experimental for the moment, but I still like and am very much looking forward to when @lauderdale will find the time to further test and improve it :)

PR #613, opened 2017-03-23
Add ... argument to textmodel_wordshoal()

kbenoit:
Allows passing of arguments to textmodel_wordfish() nested in the first part of the procedure. This allows user control as per #612 so that the user can specify `sparse = FALSE` for instance, to get the "classic wordfish" with normal SVD starting values. (For the smaller matrixes in the example, at least, this also is unlikely to harm performance.)

Also, remove the dir argument, since this was unused in the function. @lauderdale I think you might have intended to set this for the groups? It cannot be passed to the textmodel_wordfish() call, since the members of each group vary. If it was intended for groups, you need to add code to enforce this orientation.


lauderdale:
Ok, on allowing argument passing.

The dir argument should be applied at the Wordshoal level, not within the wordfish calls, it is a constraint on the general scale not the individual debate/group scales.  If there is not code to do that now, it should be easy to add, just a sign flip on some of the parameters at the Wordshoal level in post-processing before returning the estimates.

BL

On 23 Mar 2017, at 23:50, Kenneth Benoit <notifications@github.com<mailto:notifications@github.com>> wrote:


Allows passing of arguments to textmodel_wordfish() nested in the first part of the procedure. This allows user control as per #612<https://github.com/kbenoit/quanteda/issues/612> so that the user can specify sparse = FALSE for instance, to get the "classic wordfish" with normal SVD starting values. (For the smaller matrixes in the example, at least, this also is unlikely to harm performance.)

Also, remove the dir argument, since this was unused in the function. @lauderdale<https://github.com/lauderdale> I think you might have intended to set this for the groups? It cannot be passed to the textmodel_wordfish() call, since the members of each group vary. If it was intended for groups, you need to add code to enforce this orientation.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/kbenoit/quanteda/pull/613

Commit Summary

  *   Add ... argument to textmodel_wordshoal()

File Changes

  *   M R/textmodel-wordshoal.R<https://github.com/kbenoit/quanteda/pull/613/files#diff-0> (26)
  *   M man/textmodel_wordshoal.Rd<https://github.com/kbenoit/quanteda/pull/613/files#diff-1> (24)

Patch Links:

  *   https://github.com/kbenoit/quanteda/pull/613.patch
  *   https://github.com/kbenoit/quanteda/pull/613.diff

kbenoit:
@lauderdale would be great if you could tidy up this branch so I can merge it into `master` and close the PR.

PR #622, opened 2017-03-26
Add textmodel_wordscores_predicted

stefan-mueller:
Add wordscores plots and examples.


R/textplot_scale1d.R
```diff
@@ -135,4 +155,108 @@ textplot_scale1d.textmodel_wordfish_fitted <-
...
+        p <- if (sort) {
+          ggplot(data = results, aes(x = reorder(doclabels, textscore), y = textscore))
+        } else {
+          ggplot(data = results, aes(x = doclabels, y = textscore))
+        }
+        p <- p + 
+          coord_flip() + 
+          { if (!is.null(groups))
+            facet_grid(as.factor(groups) ~ ., scales = "free_y", space = "free") } +  
+          { if (rescaling_plot != "mv")
```
stefan-mueller: Something seems wrong here. In my examples `predict(...rescaling = "mv")` did not produce confidence intervals, so I added this line to avoid an error. But I presume there are better ways around this.

PR #623, opened 2017-03-27
Resolved #562: parallelize wordfish

### LINKED ISSUE #562: Parallelize textmodel_wordfish

HaiyanLW:
Benchmark is also added to compare with non-mt version and dense version.
See if need to remove other versions.


kbenoit:
Looks great @HaiyanLW! I'm just thinking a bit about what to call the argument (perhaps `parallel = TRUE` instead of `mt`) and what to call the options. 

Perhaps `options(quanteda_parallel = TRUE)`? We could make all **quanteda** options have the same prefix. (I think `mt` is not descriptive enough.) @koheiw, thoughts?

kbenoit:
OK, @koheiw and @HaiyanLW take a look at the new function in this branch called `quanteda_options()`. It's made to behave just like `options()`, but it also sets `RcppParallel::setThreadOptions()` for the option `threads`. I called this `threads` instead of `cores` since cores are a physical machine attribute and threads and the number of separate processes that usually outnumber the cores, so threads is more accurate (and standard).

HaiyanLW:
`quanteda_options()` looks good - a systemic approach. We could add more options if needed in the future.

Issue #484, opened 2017-01-20
Why tokens() has camel-cased arguments like removePunct?


koheiw:
We have both camel and snake-cased arguments in the functions. Why not changing `removePunct` to `remove_punct`, for example?

kbenoit:
It's planned - but since this is one of the most commonly used functions, I wanted to figure out how to deprecate the older names first. See approach suggested in #478.

New names would be `remove_punct`, `remove_numbers`, etc.

koheiw:
How about this?

```r
catcher <- function(arg_new, argOld){
    if (missing(arg_new)) return(argOld)
    if (missing(argOld)) return(arg_new)
    return(arg_new)
}

test_func <- function(arg_new, ...){
    arg <- catcher(arg_new, ...)
    cat('arg:', arg, "\n")
}

test_func(arg_new = 'aaa')
test_func(argOld = 'aaa')
test_func(arg_new = 'ccc', argOld = 'ddd') # new one has the priority
```

PR #630, opened 2017-04-01
Fix for #484

koheiw:
This is a fix for #484 in a way that you suggested. It is working well, but I could not figure out how to suppress warnings like `Argument removeNumbers not used`. Do you know how?


kbenoit:
I've made more changes to make this pretty global, including all calls to the functions. Also fixed the warning you were getting. 

However we need to change the deprecation calls, since it works unless the value is something defined in the calling environment/parent frame. That's the call in textstat_readability() fails. We need to evaluate each supplied argument from the parent frame to get its value. Will fix soon or you can have a go.

kbenoit:
Done now, and for `collocations()` (soon to be radically changed anyway) and for `textmodel_wordfish()`, the only other two functions as far as I know that have camelcase arguments.

@koheiw See my mods to the function and to tokens.R if you are curious about the problem I encountered with enclosing frames.

Issue #564, opened 2017-02-24
Implement sparse textmodel_ca


kbenoit:
This would use `svds` to perform correspondence analysis, which is SVD on the input-transformed dfm, followed by rescaling. @kbenoit can provide references but the code is also in the `ca` package and is quite simple. Right now we have `textmodel_ca` that uses that package but it operates only after coercing the dfm to a (dense) `matrix` class.

PR #640, opened 2017-04-06
solve #564 implement sparse ca

HaiyanLW:


HaiyanLW:
Wait, don't pull yet. :)

kbenoit:
OK - I was going to say it all looks fantastic. Two questions:

1. Do we need RSpectra as an option too? would not rsvd be sufficient?
2. The option `mt` and `threads` are different from how we implement this in wordfish, where it's `sparse` and `threads`. In text model_ca, what happens when `mt = TRUE, threads = 2`?

HaiyanLW:
RSpectra is based on an established method "ARPACK", which is also used in `matlab` etc.

HaiyanLW:
Removed the `method` option.
Still kept the `sparse = FALSE` and `threads = 1` so that when `sparse = TRUE` and `threads = 1` to retain the sparsity and keep the memory cost low.

kbenoit:
Ok so it's like textmodel_worfish now.

PR #642, opened 2017-04-07
Issue 460

koheiw:


koheiw:
I was not sure who wants to concatenate tokens either, but just did it for consistency with corpus objects. Maybe, tokens created from different kwic will be joined together by someone.

Good idea to strip off docvars, but docnames can be the same in above cases, so concatenation should be allowed regardless of docnames.

koheiw:
I disallowed concatenating tokens with same document names until someone complains about this.

kbenoit:
Fine with me on disallowing concentrating tokens.

Issue #643, opened 2017-04-08
Show raw word frequency in targe and reference groups in textstat_keyness()


koheiw:
Outputs from textsat_keyness looks like this 
```r
               chi2 p
писпанен  -398.0878 0
казнин    -342.9210 0
news      -255.0564 0
cобчак    -227.8563 0
путина    -212.3355 0
власть    -209.6379 0
засурский -200.4484 0
артамонов -189.4854 0
власти    -175.1391 0
фишман    -163.7517 0
```
but should be 

```r
               chi2 p target reference
писпанен  -398.0878 0     15       327
казнин    -342.9210 0    108       455
news      -255.0564 0      6       202
cобчак    -227.8563 0      0       168
путина    -212.3355 0    558       880
власть    -209.6379 0    767      1094
засурский -200.4484 0      0       148
артамонов -189.4854 0      0       140
власти    -175.1391 0   1579      1811
фишман    -163.7517 0      2       126
```

PR #644, opened 2017-04-08
Issue 643

koheiw:
Changes for #643


koheiw:
This decreases codecove only by <.01%. Please merge.

PR #645, opened 2017-04-09
Fix kwic that blows up when padding is at the top of documents

koheiw:


koheiw:
My test had done somewhere..., but I recovered it.

Issue #657, opened 2017-04-15
char_segment raises worng warning


koheiw:
It is not segmenting into tokens, apparently.
```r
out <- char_segment(data_char_ukimmig2010, what = "sentences", delimiter = '. ')
Warning message:
In get_delimiter(what, delimiter, valuetype) :
  delimiter is not used for tokens
```

koheiw:
@kbenoit  `corpus_segment` has a option `what = tags`, which looks like a legacy from earlier versions. Do we still need this?

kbenoit:
The `what = "tags"` is just a convenience for users who have marked up their documents with `##TAG` type tags, and who don't know much about regular expressions. It's the same as "other" but with a specific delimiter in the regex. I know a lot of users who are find this useful, so prefer to keep the option.

Issue #619, opened 2017-03-25
corpus_segment needs an overhaul


kbenoit:
1. `corpus_segment()` and `corpus_reshape()` overlap in functionality, so we need to figure out the precise differences and separate out any redundant functionality. This might mean removing `corpus_segment()`. Note also that `char_segment()` substantially overlaps with `tokens()`, especially for `what = "sentence"`.

2. The man page for `corpus_segment()` is pretty horrible, especially for the delimiter argument.

kbenoit:
I'm starting to think that we should separate the functionality of `tokens()` and `char/corpus_segment()` as well. Right now our `'tokens()` object generates tokens, but offers options for a) the method and object of segmentation, and b) selection of tokens. The main reason for b) is that these options are built into **stringi** and therefore efficient to perform at the tokenisation level.

Ideas:

* It would be clearer to separate the functions so that `tokens()` only creates our tokens object, and `*_segment()` performs the segmentation. The segmentation could have different engines, such as:
    - fixed white space:  `stringi::stri_split_fixed(txt, " ")`
    - whitespace character class: `stringi::stri_split_charclass(txt, "\\p{WHITE_SPACE}")`
    -  `stringi::stri_split_boundaries(txt, type = "word")`
    - **spaCy**
    - `tokenizers::tokenize_word()`
    
* for non-words, `*_segment` could also have different targets:
    - sentences
    - paragraphs
    - tags
    - other

* because `char_segment()` would segment a character object, we could easily parallelise it using `parallel::mclapply()` within R.

koheiw:
Related to #648, `corpus_segment` has to keep original document IDs in `docvar` for downstream processes.

koheiw:
Initiated by#657, I am overhauling `corpus_segment`. I will keep segmentation by `what = tags`, but it seems that segmentation by `what = tokens` should be removed.

kbenoit:
Ok, how about describing the plan first? We can design it and then code it?

koheiw:
New version is fully compatible with old one, but I can solve several issues in the branch. I will send a PR soon.

koheiw:
corpus_reshape segments into only `c("sentences", "paragraphs", "documents")`, so it isa simple version of corpus_segment.
```r
mycorpus <- corpus(c(textone = "This is a sentence.  Another sentence.  Yet another.",
                     textwo = "Premiere phrase.  Deuxieme phrase."),
[...]

PR #659, opened 2017-04-16
Issue 657

koheiw:
For #619 #657. This should be much faster and consistent than old one as many operations are vectorized.


koheiw:
With 138K documents, 
```r
Old:
   user  system elapsed 
 83.104   0.000  83.124 

New:
   user  system elapsed 
 34.912   0.212  35.203
```

kbenoit:
Very nice. I like the streamlined code and the efficiency gain. Two comments:

1.  The addition of the `docid` docvar for `corpus_segment()` takes this one step closer to `corpus_reshape()`, with the main difference of `corpus_reshape` of course being that it can be returned to its original shape. This points to several options to differentiate or collapse them:
    a. We merge the functions and eliminate one of them.
    b. We remove `docid` from `corpus_segment()`, or at least make it `_docid` to hide it.
    c. We leave it as it is now.

2.  The elimination of the ability to set the `delimiter` when `what = "paragraphs"` limits the user's options while focusing on the definition of a paragraph. Obviously the `what = "other"` offers complete flexibility around this, but that may not be obvious to all users.

    Here's an idea: We add a new option `quanteda_paragraph_delimiter = "(\\r{0,1}\\n{1}){2}"` that sets the number of newlines expected to 2, but this can be redefined to 1 as an option. We could just make this an integer, and define the newlines internally. I like the idea of trying to capture the differences in antique Windows newlines, versus Unix-style. I also recall however that antique Mac was just "\r". What if we normalise this in the corpus upon corpus creation?

    ```r
    stringi::stri_split_regex("line1\r\n\r\nline2\n\nline3", "(\\r{0,1}\\n{1}){2}", 
                              omit_empty = TRUE)
    ## [[1]]
    ## [1] "line1" "line2" "line3"
    ```

    but that still does not get antique Mac formats.

koheiw:
@kbenoit , thank you for reviewing and commenting. I made several changes:

- EOL normalization is done i when `corpus` is created (\r\n and \r are all converted to \n)
- corpus_segment accepts `delimiter` when `what = paragraphs`
- corpus_segment records document ID and segment ID in internal docvars (_docid and _segid).

Rearding the `quanteda_paragraph_delimiter` option, it is always good to centralize project level settings options, but EOL markers differ document to document (e.g. Bush 2005 in `data_corpus_inaugural`). I also don't think users who do not understand regex can correctly set the number of \n for paragraph endings, in the first place.

Having added `segid`, `corpus_segment` is one more step closer to `corpus_reshape`. If we are to keep `corpus_reshape` I can easily rewrite the function using `corpus_segment` internally for sentences and paragraphs, but not sure what to do with tags. I can still reconstruct tagged texts nearly completely, it does not look like 'reshaping'. Also, I think users should not transform corpus back and forth as conversion potentially destroy the data.

koheiw:
Travis got stuck.

Issue #660, opened 2017-04-17
Error message: In is.na(x) : is.na() applied to non-(list or vector) of type 'NULL'


Ninoninoninonino:
I get the error message "In is.na(x) : is.na() applied to non-(list or vector) of type 'NULL'" after `quantdfm <- dfm_select(quantdfm, stopwords("english"), "remove")`. Before, when loading the object, it is `> class(quantdfm)
[1] "dfmSparse"
attr(,"package")
[1] "quanteda"`

The same error message can be found here: http://quanteda.io/reference/dfm.html
`> sessionInfo()
R version 3.3.2 (2016-10-31)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Red Hat Enterprise Linux Server 7.2 (Maipo)

locale:
 [1] LC_CTYPE=en_GB.UTF-8          LC_NUMERIC=C                 
 [3] LC_TIME=en_GB.UTF-8           LC_COLLATE=en_GB.UTF-8       
 [5] LC_MONETARY=en_GB.UTF-8       LC_MESSAGES=en_GB.UTF-8      
 [7] LC_PAPER=en_GB.UTF-8          LC_NAME=en_GB.UTF-8          
 [9] LC_ADDRESS=en_GB.UTF-8        LC_TELEPHONE=en_GB.UTF-8     
[11] LC_MEASUREMENT=en_GB.UTF-8    LC_IDENTIFICATION=en_GB.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] digest_0.6.12          tidytext_0.1.2         dplyr_0.5.0           
 [4] cldr_1.1.0             stm_1.2.1              data.table_1.10.4     
 [7] quanteda_0.9.9-47      tm_0.7-1               NLP_0.1-10            
[10] qdap_2.2.5             RColorBrewer_1.1-2     qdapTools_1.3.1       
[13] qdapRegex_0.6.0        qdapDictionaries_1.0.6 stringr_1.2.0         
[16] XML_3.98-1.6           RCurl_1.95-4.8         bitops_1.0-6          

loaded via a namespace (and not attached):
 [1] Rcpp_0.12.10        lattice_0.20-34     tidyr_0.6.1        
 [4] xlsxjars_0.6.1      gtools_3.5.0        assertthat_0.1     
 [7] psych_1.7.3.21      slam_0.1-40         R6_2.2.0           
[...]

PR #662, opened 2017-04-17
Issue 660

koheiw:


koheiw:
This is to make `dfm_select()` more robust for #660.

Issue #668, opened 2017-04-19
`tokens_compoud()` fails when dictionary concatenator is not a white space


koheiw:
`tokens_compoud()` fails when dictionary concatenator is not a white space:
```r
#FAIL:
toks <- as.tokens(list(c('a', 'b', 'c', 'd')))
dict <- dictionary(list(comp = c('a', 'b_c')), concatenator = '_')
tokens_compound(toks, dict, concatenator = '')

#OK:
toks <- as.tokens(list(c('a', 'b', 'c', 'd')))
dict <- dictionary(list(comp = c('a', 'b c')), concatenator = ' ')
tokens_compound(toks, dict, concatenator = '')
```

Issue #450, opened 2017-01-12
features should behave consistently


kbenoit:
Right now, functions that take a "features" argument have inconsistent handling of this argument type. From `tokens_select`:

> `features`  one of: a character vector of features to be selected, a dictionary class object whose values (not keys) will provide the features to be selected.

But in fact a list works too: 

```r
toks <- tokens(c(d1 = "a b c d e", d2 = "a d e f g"))
mydfm <- dfm(toks)
feats <- c(one = "a", two = "c", three = "f")

tokens_select(toks, feats)
# tokens from 2 documents.
# d1 :
# [1] "a" "c"
# 
# d2 :
# [1] "a" "f"
#
tokens_select(toks, as.list(feats))
# tokens from 2 documents.
# d1 :
# [1] "a" "c"
# 
# d2 :
# [1] "a" "f"
#
tokens_select(toks, dictionary(as.list(feats)))
# tokens from 2 documents.
# d1 :
# [1] "a" "c"
# 
# d2 :
# [1] "a" "f"
#
```

However this is not the case for `dfm_select`:
```r
dfm_select(mydfm, feats)
# kept 3 features, from 3 supplied (glob) feature types
# Document-feature matrix of: 2 documents, 3 features (33.3% sparse).
# 2 x 3 sparse Matrix of class "dfmSparse"
#     features
# docs a c f
#   d1 1 1 0
#   d2 1 0 1
dfm_select(mydfm, as.list(feats))
# Error in dfm_select.dfm(mydfm, as.list(feats)) : 
#  features must be of type character, dictionary, or dfm 

dfm_select(mydfm, dictionary(as.list(feats)))
# kept 3 features, from 3 supplied (glob) feature types
# Document-feature matrix of: 2 documents, 3 features (33.3% sparse).
# 2 x 3 sparse Matrix of class "dfmSparse"
#     features
# docs a c f
#   d1 1 1 0
#   d2 1 0 1
```

**Preferred approach:**

We consolidate the features/sequences function as in `tokens_compound`, which uses `@inheritParams sequence2list` which has its own argument list. This way, as with `valuetype`, we centralise and standardise the handling of the arguments for features and sequences.

These take:
*  character vector, whose elements will be split on whitespace;
*  list of characters, consisting of a list of token patterns, separated by white space;
*  tokens object;
*  dictionary object;
*  collocations object.

**Functions affected**:

* `tokens_compound`  
* `tokens_select`
* `sequences`  
* `dfm_select`
[...]

PR #669, opened 2017-04-19
Issue668

koheiw:
Solves #450 and #668, but there are some documentation warnings... Hope @kbenoit can solve this:
```
* checking Rd \usage sections ... WARNING
Undocumented arguments in documentation object 'tokens_compound'
  ‘sequences’

Functions with \usage entries need to have the appropriate \alias
entries, and all their arguments documented.
The \usage entries must correspond to syntactically valid R code.
```


kbenoit:
@koheiw The problem here is that in `tokens_compound()`, we need to document a `@param` called `sequences`. Formerly this was obtained by `@inheritParams sequence2list` because `sequence2list` had an argument called `sequences`. When you changed this to `features2list` which has the argument `features`, it means we are importing a parameter without the name we need in `tokens_compound()`.

Solutions include:
* change argument in `features2list()` to `sequences`
* change argument in `tokens_compound()` to `features` - which I do not prefer since this changes an exposed function API
* make a non-inherited entry in `tokens_compound()` for `@param sequences`. This is the simplest fix and the one I will use, since no where else do we currently inherit the parameters for the `sequence/feature2list` function.

kbenoit:
The fix makes it less modular but we can make it modular again the next time another function needs to document an identical argument.

PR #671, opened 2017-04-19
Fix NOTES on CRAN

kbenoit:
* Reduces the installed package size < 5MB by removing deprecated data objects, and all references to them.
* Wraps examples `textplot_scale1d` in `\donttest{}` so that these don't exceed the 5 second limit.


kbenoit:
@HaiyanLW If we could also remove the dependency on GNU Make then we would not even have a single CRAN NOTE. (But only if this is possible.)

PR #674, opened 2017-04-19
Issue670

### LINKED ISSUE #670: `corpus_reshape.corpus` fails on `corpus` objects which contain zero length texts
I'm using the latest github version 0.9.9-48.

sessionInfo()
> R version 3.3.3 (2017-03-06)
> Platform: x86_64-w64-mingw32/x64 (64-bit)
> Running under: Windows >= 8 x64 (build 9200)
> 
> locale:
> [1] LC_COLLATE=English_United Kingdom.1252  LC_CTYPE=English_United Kingdom.1252   
> [3] LC_MONETARY=English_United Kingdom.1252 LC_NUMERIC=C                           
> [5] LC_TIME=English_United Kingdom.1252    
> 
> attached base packages:
> [1] stats     graphics  grDevices utils     datasets  methods   base     
> 
> loaded via a namespace (and not attached):
>  [1] NLP_0.1-9               Rcpp_0.12.10            quanteda_0.9.9-48      
>  [4] git2r_0.18.0            plyr_1.8.4              base64enc_0.1-3        
>  [7] tools_3.3.3             digest_0.6.12           jsonlite_1.2           
> [10] memoise_1.0.0           evaluate_0.10           tibble_1.2             
> [13] gtable_0.2.0            lattice_0.20-34         Matrix_1.2-8           
> [16] fastmatch_1.1-0         curl_2.3                yaml_2.1.14            
> [19] parallel_3.3.3          withr_1.0.2             httr_1.2.1             
> [22] stringr_1.1.0           knitr_1.15.1            devtools_1.12.0        
> [25] rprojroot_1.2           grid_3.3.3              data.table_1.10.4      
> [28] R6_2.2.0                XML_3.98-1.6            rmarkdown_1.3          
> [31] tm.plugin.factiva_1.6.1 ggplot2_2.2.1           magrittr_1.5           
> [34] scales_0.4.1            backports_1.0.5         htmltools_0.3.5        
> [37] assertthat_0.1          colorspace_1.3-2        stringi_1.1.5          
> [40] RcppParallel_4.3.20     lazyeval_0.2.0          munsell_0.4.3          
> [43] slam_0.1-40             tm_0.6-2
> 

Calling `corpus_reshape.corpus` with the "paragraphs" option, but I expect you'd see the same problem with "sentences". Zero length texts should retained because they may have useful information in the docvars, and the `corpus.VCorpus` constructor function respects this, which is how I can get to a situation where I'm calling `corpus_reshape.corpus` on a corpus containing zero-length texts.

The failure occurs in line 44 in `corpus_reshape.corpus` contains a call to seq(n = 0), when there are no paragraphs that returns "1" "0" if lengthSegments equals zero, when really no sentenceid is meaningful since there is no sentence -- only a zero-length document. 
`sentenceid <- lapply(lengthSegments, function(n) seq(from = 1, to = n))`

Since sentenceid must be filled, to make this list the same length as other list with which it is to be combined in a dataframe, I suggest giving it the value NA, though zero could be an alternative (could leading to unexpected behaviour though).

Suggested fix is an if statement for the case where n == 0. i.e.
```
gen_seq <- function(n) if (n == 0) NA else seq(from = 1, to = n))
sentenceid <- lapply(lengthSegments, gen_seq) 
```

koheiw:
This is to fix #670 by making `corpus_reshape` based on `corpus_segment`. `corpus_trimsentences` is generalized as `corpus_trim`.


koheiw:
Also for #619

koheiw:
Why this PR is not merge to master yet?

kbenoit:
Not merged yet because it changes a behaviour and introduces a new command, and I want to thin this through first. Also I have two other packages (not public) that depend on the function that was renamed. Will get to a full review soon.

PR #682, opened 2017-04-23
Save memory by recycling input objects

koheiw:
Partially for #691. Recycle input objects in C++ to reduce memory usage. Yet, input R objects are not affected.


kbenoit:
@koheiw I think a recent PR created a merge conflict - can you fix the one in `src/tokens_kwic_mt.cpp` before I accept the PR?

Issue #650, opened 2017-04-10
Allow subsetting dfm with lgCMatrix


koheiw:
Relational operations with dfm return lgCMatrix, but it cannot be used with dfm.
```r
txt <- c('a b c', 'a b c', 'b c d e')
mx <- dfm(txt)
mx[,'d'] > 0 
mx[mx[,'d'] > 0,]  # error
```
We have to do like this at the moment:
```r
mx[which(mx[,'d'] > 0),] 
```

koheiw:
dfm's signature need to be defined unambiguously to do subsetting by a logical vector. This is related to #494.

koheiw:
This involves reimplementation of `dfm`. Should be solved along with #494.

kbenoit:
I think the will be easier to solve than that, just involves a new method for `[`. I'll give it a try.

kbenoit:
May not be our fault:

```r
as(mx, "dgCMatrix")[mx > .5, ]
# Error in as(x, "generalMatrix")[i = i, , drop = TRUE] : 
#   invalid or not-yet-implemented 'Matrix' subsetting
```
(appears to be an issue in **Matrix**)

koheiw:
I could write
```r
setMethod("[",
          signature = c("dfm", i = "lgCMatrix", j = "missing", drop = "missing"),
          function(x, i, j, ..., drop = FALSE) {
              return(x[as.vector(i),])
          })
```
to make it work, but there is not guarantee that the **lgCMatrix** has only one column.
The right way of doing this is 
```r
mx[rowSums(mx[,'d']) > 0,] 
```
then we do not need a new method for **lgCMatrix**.

Issue #494, opened 2017-01-21
make all dfm types a single object class


kbenoit:
Make all dfm objects `dgCMatrix` objects even if completely "dense". Right now we have a dense object type `dfmDense` for a matrix that has been weighted or smoothed so that it no longer contains zeroes. 

It would be simpler to make every dfm always be a `dgCMatrix` even without zeros. See the implementation in `dfm_compress.dfmDense()` - this coerces the `dfmDense` into a `dfmSparse` and then calls `dfm_compress.dfmSparse()`. We could take this approach generally and make a single `dfm` class object that is always based on `dgCMatrix`, and just call it `dfm`.

The only exception would be if the **Matrix** package implements the sparse integer matrix, in which case it would be worth using that to record count matrixes (basically, every dfm that has not been weighted).

PR #739, opened 2017-05-18
Remove dfmDense and name dfmSparse just dfm

koheiw:
This is a branch to solve issue #650, but I noticed that multiple inheritance is making creation of new **dfm** methods difficult. So I started working on #494 and removed the **dfmDense** class. The new **dfm** only directly inherits **dgCMatrix** and is working in the same way as the old one. However, I am not sure how to design **fcm** which inheris **dfm** and **dgCMatrix**.


kbenoit:
This is a non-trivial, basic infrastructural change which is why I didn't think it appropriately set for the milestone of the v0.9.9000 release.  Let's put it on the back burner until we sort out smaller issues, and then address it when we get rid of the ancient wiring, including `tokenizedTexts`, `dfmDense`, etc.

koheiw:
OK. I will leave this.

Issue #653, opened 2017-04-12
Fix registerRoutines NOTE


kbenoit:
I'm getting this after `devtools::build_win()`:

```
* checking compiled code ... NOTE
File 'quanteda/libs/i386/quanteda.dll':
  Found no calls to: 'R_registerRoutines', 'R_useDynamicSymbols'
File 'quanteda/libs/x64/quanteda.dll':
  Found no calls to: 'R_registerRoutines', 'R_useDynamicSymbols'

It is good practice to register native routines and to disable symbol
search.

See 'Writing portable packages' in the 'Writing R Extensions' manual.
```

We need to fix before resubmitting to CRAN (next week!).

PR #750, opened 2017-05-24
updated NAMESPACE

### LINKED ISSUE #749: Need to register native routines and disable symbol search
I'm seeing this in the CRAN check on Windows R 3.4:

```
* checking compiled code ... NOTE
File 'quanteda/libs/x64/quanteda.dll':
  Found no calls to: 'R_registerRoutines', 'R_useDynamicSymbols'

It is good practice to register native routines and to disable symbol
search.

See 'Writing portable packages' in the 'Writing R Extensions' manual.
```

Please fix ASAP since I need to resubmit to CRAN.

HaiyanLW:
fix #749 
Last time for #653, I also changed this in `NAMESPACE` and it worked. If this doesn't work, we may need to use `R_useDynamicSymbols(dll, TRUE);` instead of `R_useDynamicSymbols(dll, FALSE);` in `src/init.c`.


kbenoit:
Thanks Haiyan, I'll try this in Windows and see if the warning disappears.

kbenoit:
Strange... I tried it both ways, and even regenerated init.c on Windows R 3.4, and it still produces the exception.

HaiyanLW:
When you regenerated `init.c`, is it same as before? Did you update the `NAMESPACE` after generating a new `init.c`?

kbenoit:
Not sure - check the code I pushed to this branch, I did this from windows.

HaiyanLW:
The `init.c` is exactly same. Last line in NAMESPACE,  need to be updated to  `useDynLib(quanteda, .registration = TRUE)` (Don't let roxgen2 rewrite it).

Issue #763, opened 2017-06-02
consolidate initial, default, and user-set quanteda_options


kbenoit:
Two changes here:

- [x] consolidate options for startup and  `quanteda_options(reset = TRUE)` 

    Currently, we have options set in both `quanteda_options.R` and `zzz.R`. These have different starting defaults, but these should be consolidated into a single function. I prefer the idea of `.onAttach()` in `zzz.R` calling `quanteda_options(reset = TRUE)` to set these, and having the defaults set in `quanteda_options.R`.

- [x] check that user-set options are not overridden on startup. So if a user sets these in `~/.Rprofile` for instance, they should not be overridden by `.onAttach()` in `zzz.R`.

HaiyanLW:
Yeah, seems we double `reset` in `zzz.R`-- [`zzz.R`](https://github.com/kbenoit/quanteda/blob/master/R/zzz.R#L6) does call `quanteda_options(reset = TRUE)`. 
I think we should still allow users to call `quanteda_options(reset = TRUE)`.

kbenoit:
Yes of course we want to keep `quanteda_options(reset = TRUE)`. I am proposing that `zzz.R` call that function to set the options. However we don't want to overwrite values already present, which would be the case if a user had set them in `~/.Rprofile`, so we need to check if they exist or not before we reset those values.

PR #768, opened 2017-06-06
fix Issue #763

HaiyanLW:
Two methods implemented. Instead of reading from .Rprofile, the commented out part recovers the `options` after `reset`. @kbnoit which way do you prefer?


kbenoit:
@HaiyanLW I'm going to merge this into master, but thought you might want to take a look first at how I did it. The problem was if the package was not attached, and people called

```r
quanteda::dfm(mytexts)
```
it generated an error because `quanteda_options("verbose")` was unassigned. It now checks for that if if present, calls `quanteda_initialize()`. This preserves any settings already defined, so that if a user had set them in `.Rprofile` they would not get unset. (To unset all variables and set to the default values, the function is `quanteda_options(reset = TRUE)`, as before.)

HaiyanLW:
Oh, right. Lots of changes. :+1:

PR #777, opened 2017-06-07
Fix textmodel_NB with Bernoulli distribution

### LINKED ISSUE #776: different likelihoods in textmodel_NB give same result
Could it be the case that only one of these distributions is actually implemented? I get the same results with the following example from the manual (and with my own data).
```
(nb.p261 <- textmodel_NB(trainingset, trainingclass, distribution="Bernoulli"))
(nb.p261 <- textmodel_NB(trainingset, trainingclass, distribution="multinomial"))
```

kbenoit:
- textmodel_NB(x, y, distribution = "Bernoulli") previously had no effect - it was always multinomial.
- Fixes #776

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: quanteda
-Version: 0.9.9.67
+Version: 0.9.9.68
 Title: Quantitative Analysis of Textual Data
 Description: A fast, flexible framework for for the management, processing, and
     quantitative analysis of textual data in R.
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,7 @@
 ### Bug fixes and stability enhancements
 
 *  For `sequences()`: Corrected the word matching, and lambda and sigma calculation methods for `unigram subtuples` algorithm in `sequences_mt.cpp`, and consequently the p-values on the bigrams are correct. 
+*  `textmodel_NB(x, y, distribution = "Bernoulli")` now correctly sets the features to Bernoulli features prior to fitting the model (#776).
 
 
 ## Changes since v0.9.9-50
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -41,12 +41,12 @@ qatd_cpp_fcm <- function(texts_, n_types, count, window, weights, ordered, tri, 
     .Call(quanteda_qatd_cpp_fcm, texts_, n_types, count, window, weights, ordered, tri, nvec)
 }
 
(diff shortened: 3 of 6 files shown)


thomasd2:
Thanks. 
`(nb.p261 <- textmodel_NB(trainingset, trainingclass, distribution="Bernoulli", smooth=0))`
Could it be that the likelihoods need to be multiplied with 2? Beijing appears in one of three Yes class docs, and Chinese in all of them?

kbenoit:
You would never want to run NB without smoothing, since you will not be able to apply this as a predicted model in the case where if you fitted the model above, you had both Beijing and Japan in the test set.

```r
Fitted Naive Bayes model:
Call:
	textmodel_NB(x = trainingset, y = trainingclass, smooth = 0, 
    distribution = "Bernoulli")


Training classes and priors:
  Y   N 
0.5 0.5 

		  Likelihoods:		Class Posteriors:
6 x 4 Matrix of class "dgeMatrix"
                 Y         N   Y   N
Beijing  0.1666667 0.0000000 1.0 0.0
Chinese  0.5000000 0.3333333 0.6 0.4
Japan    0.0000000 0.3333333 0.0 1.0
Macao    0.1666667 0.0000000 1.0 0.0
Shanghai 0.1666667 0.0000000 1.0 0.0
Tokyo    0.0000000 0.3333333 0.0 1.0
```

If this is not what you meant, please clarify.

thomasd2:
My point is this: Beijing appears in one of the three training documents in the yes class, so I expect the [1,1] entry in the output to be Pr(Beijing  | Y) = 1/3, and [2,1] Pr(Chinese | Yes ) = 1. I set smooth to 0 to rule out that the deviation is due to smoothing. Am I completely off or is there a bug in the code?

thomasd2:
Actually, I suspect there is also an issue in the multinomial version.
```
trainingset
(nb.p261.nosmooth <- textmodel_NB(trainingset, trainingclass, smooth=0))
(nb.p261 <- textmodel_NB(trainingset, trainingclass))
```
In the unsmoothed version, the likelihoods look fine. There are in total 8 words in the three training docs where class == Y. 1/8 of the words are Beijing and 5/8 are Chinese.
For the yes class, smoothing adds 3*6 = 18 to the total count yields 26. But the likelihood for Beijing is not 4/26 here. In the no class, it seems fine (adding 6 to the total count gives nine), perhaps because there is only one row.

kbenoit:
Moved to https://github.com/kbenoit/quanteda/issues/780.

PR #790, opened 2017-06-13
Stop transposing sparseMatrix in compile_dfm.tokens

amatsuo:
This is a small revision to compile_dfm.tokens. It seems not necessary to transpose sparseMatrix if sparseMatrix is constructed in this way. Transposing can be costly if a `dfm` is huge.

### DIFF
--- a/R/dfm.R
+++ b/R/dfm.R
@@ -434,12 +434,12 @@ compile_dfm.tokens <- function(x, verbose = TRUE) {
         index <- index + 1
     }
     
-    temp <- t(sparseMatrix(i = index, 
-                           p = cumsum(c(1, lengths(x))) - 1, 
-                           x = 1L, 
-                           dims = c(length(types), length(names(x))),
-                           dimnames = list(features = as.character(types), 
-                                           docs = names(x))))
+    temp <- sparseMatrix(j = index, 
+                         p = cumsum(c(1, lengths(x))) - 1, 
+                         x = 1L, 
+                         dims = c(length(names(x)), length(types)),
+                         dimnames = list(docs = names(x),
+                                         features = as.character(types)))
     new("dfmSparse", temp)
 }
 


kbenoit:
@koheiw just glancing at the change it's not going to work because it changes the `p`. `p` is the index of the first non-zero element in the column, for the column sparse matrix format, and since this comes from the features, the features have to be rows (so that their lengths in terms of the input tokens object define the column sparse index).

I've run tests and the cost of transposing is actually small compared to the benefits of constructing the sparse column matrix directly, versus converting from the triplet (with i, j, x instead of i, p, x). (Although converting from the latter to sparse column format is pretty fast too!)

kbenoit:
Note that if the checks pass, then either a) I am wrong or b) we don't actually have tests for this, in which case we need to add tests!

amatsuo:
@kbenoit Please refer to the help for `sparseMatrix`. In matrix construction, `p` could be the index of the first non-zero element in the column (or row). Here is a quote from the help 

> p: numeric (integer valued) vector of pointers, one for each column (or row), to the initial (zero-based) index of elements in the column (or row). Exactly one of i, j or p must be missing.

In my understanding, when `j` is supplied, `p` works as the index of the first non-zero element in the row (not column).

kbenoit:
@amatsuo you're right! I'm very glad to see that it was a) above. Thanks for the contribution!

amatsuo:
Although the pull request is already merged, I post a test code to show that my revision produces the same output and some benchmark

```{r}
set.seed(123)
msize <- 1e5
temp <- sparseMatrix(i = sample(msize, msize * 10, TRUE),
                     j = sample(msize, msize * 10, TRUE),
                     x = rnorm(msize * 10))
temp_with_i_t <- t(sparseMatrix(i = temp@i + 1,
                                p = temp@p,
                                dims = rep(msize, 2),
                                x = temp@x))
temp_with_j <- sparseMatrix(j = temp@i + 1,
                            p = temp@p,
                            dims = rep(msize, 2),
                            x = temp@x)
all.equal(temp_with_i_t, temp_with_j)
## TRUE
```

**Benchmark:**
```{r}
    temp_with_i_t <- t(sparseMatrix(i = temp@i + 1,
                                    p = temp@p,
                                    dims = rep(msize, 2),
                                    x = temp@x)),
    temp_with_j <- sparseMatrix(j = temp@i + 1,
                                p = temp@p,
                                dims = rep(msize, 2),
                                x = temp@x)
)
## Unit: milliseconds
##                                                                                                 expr
##  temp_with_i_t <- t(sparseMatrix(i = temp@i + 1, p = temp@p, dims = rep(msize,      2), x = temp@x))
##       temp_with_j <- sparseMatrix(j = temp@i + 1, p = temp@p, dims = rep(msize,      2), x = temp@x)
##        min        lq     mean   median       uq      max neval
##  158.77940 174.34143 231.5304 189.2401 270.5018 770.8080   100
##   90.31064  99.05893 156.8212 106.3107 177.8883 829.0426   100
```

PR #792, opened 2017-06-13
Dev network

koheiw:
If am not sure about the default values of the arguments, but should be fine as the first version.


koheiw:
@stefan-muelle, please have a look at the initial version. Are there arguments that I have to add to the function for customization?

stefan-mueller:
Thanks, @koheiw. I installed this branch using `devtools::install_github("kbenoit/quanteda", ref = "dev-network")`. 

When I try to replicate the example from the manual, I get a warning. As a result the words are not printed in the network plot. 

```r
devtools::install_github("kbenoit/quanteda", ref = "dev-network"

require(quanteda)
require(igraph)

toks <- tokens(corpus_subset(data_corpus_inaugural, President=="Obama"), remove_punct = TRUE)
toks <- tokens_remove(toks, feature = stopwords("english"), padding = FALSE)
myfcm <- fcm(toks, context = 'window', tri = FALSE)
feats <- names(topfeatures(myfcm, 20))
textplot_network(myfcm[feats,feats], ignore = 0.5)

# There were 50 or more warnings (use warnings() to see the first 50)

warnings()
# Warning messages:
#   1: In text.default(x, y, labels = labels, col = label.color,  ... :
#                        no font could be found for family "sanserif"
#   2: In text.default(x, y, labels = labels, col = label.color,  ... :
```
I assume I haven't loaded these fonts (I tried loading the **extrafont** package too, but it didn't make a difference). Do you have an idea which package needs to be loaded?

koheiw:
@kbenoit what test can I do for a plot?

stefan-mueller:
Great, thanks for the changes, @koheiw – now I can reproduce the example. I think the default colour and transparency are fine. Options we could include:

- Adjust size of circles (seems to be a bit complicated in igraph(?
- Reduce white space around network.

I tried to use [**ggnet**](https://github.com/briatte/ggnet) to replicate your example, but the package has quite a lot of dependencies many users probably won't have installed, so the **igraph**-based solution might be preferable. Let me know what you think.

koheiw:
@stefan-mueller we cannot add arguments for **igraphs** numerous options, but it is good to add these two. Feel free do modify this branch.

stefan-mueller:
@koheiw, I made a PR with these options (and an option to change the shape). Have a look whether you can replicate the plots.

koheiw:
I saw this error in my other machine
```
Error in assign(".igraph.pars", current, envir = env) : 
  cannot change value of locked binding for '.igraph.pars'
```
It looks like a bug in **igraph**, and I yet do not know how to prevent this. Regarding test, I saw test-textplot.R, but `plot.igraph()` does not return objects but just draw charts.

Issue #795, opened 2017-06-13
Subset quanteda corpus based on token count


JumpyAF:
Hi,

I have a corpus object with 306,108 documents I would like to subset based on word count of each document. The corpus summary() command already lists the number of tokens in each document, but I am unable to access the token count using the corpus_subset() command.

I believe I can add a docvar() meta variable giving the number of tokens of each document using the ntoken() function and then use corpus_subset() to subset based on the meta variable, but this seems unnecessarily redundant. Is there any efficient way to compute this subset?

Thanks!

kbenoit:
Try:
```r
subcorpus <- corpus_subset(wholecorpus, ntoken(wholecorpus) > threshold)
```
to select only the documents whose token count exceeds `threshold`. Note that unless you supply options to `ntoken()`, the defaults for `tokens()` will be used, which include punctuation.

JumpyAF:
Hi Ken,

The command did work, but it ate up my entire 16gb of RAM and some swap space doing the computation on the corpus which was 2gb. Is there anything I can do to reduce my memory usage?

Thanks again.

koheiw:
@kbenoit the underlying function of `ntoken` is the classic function...
```r
ntoken.character <- function(x, ...) {
    ntoken(tokenize(x, ...))
}
```
Also we can use `stri::stri_count_boundaries()` for rough estimates for corpus objects.

kbenoit:
@koheiw that's a very good suggestion -  I will change the code for `ntoken.character()` now.

kbenoit:
Actually, we cannot just count boundaries in `ntoken()`, since it passes through options to `tokens(x, ...)` and this would fail to provide full control.

However @JumpyAF I suggest you try @koheiw's suggestion as:

```r
subcorpus <- corpus_subset(wholecorpus, stringi::stri_count_boundaries(texts(wholecorpus)) > threshold)
```

JumpyAF:
Hi all,

I tried Koheiw's command and while the computation took a few seconds, it did not use up any memory at all.  However the word counts it gives is different (strictly less than) from the ntoken command and the information given by summary(wholecorpus). I am wondering why there is not just a quick command to extract the token counts which seem to be already computed and displayed by the corpus.

Best.

JumpyAF:
Hi Ken,

I commented on the closed thread just now, and am just sending this email
as a heads-up.

I tried Koheiw's command and while the computation took a few seconds, it
[...]

PR #801, opened 2017-06-14
tokenize() to tokens() - Fix 795

kbenoit:
Does not completely solve the memory issue in #795 but helps. 

* `ntoken.character()` and `ntype.character()` methods now use `tokens(x, hash = FALSE)` instead of `tokenize()`, since this is faster and since `tokenize()` is deprecated. Turning off hashing avoids the unnecessary construction of the hash table, which is not needed for these functions. (My tests showed this to be faster.)
* Changed usage in other functions from `tokenize()` to `tokens()` - in the few hidden corners where this was still used.
* Fixed a bug in the corpuszip constructor that overwrote the `created` field in metacorpus data, even when this had been set previously by the user.


kbenoit:
But the intermediate object inside  `tokens()` is the unhashed, segmented words, before we hash it, so here we are not changing that, but simply avoiding the additional time required to hash the tokens.

koheiw:
In #795, 306,108 documents will be split into 31 blocks, and serialized one by one. So the intermediate object is much smaller than unserialized tokens.

kbenoit:
@koheiw you are right about the loop. I changed it back to the default, removing the `hash = FALSE` in the `tokens()` calls.

Once the corpus is an index of tokens this will become irrelevant and we will not need to compute this in the same way again, since we will be able to take this directly from querying subsets of the tokenised corpus.

koheiw:
This is the difference:

The data is only the half, but `tokens` uses 1.3GB when `hash = TRUE`, and 3.5GB while `hash = FALSE`.
```r
> load("/home/kohei/Documents/Brexit/Analysis/data_corpus_guardian.RData")
> ndoc(data_corpus_guardian)
[1] 177115

> gc(reset = TRUE) 
            used   (Mb) gc trigger   (Mb)  max used   (Mb)
Ncells   2762817  147.6    4703850  251.3   2762817  147.6
Vcells 173160717 1321.2  323266785 2466.4 173160717 1321.2
> out <- tokens(data_corpus_guardian, hash = FALSE)
 Show Traceback
 
 Rerun with Debug
 Error in UseMethod("docvars<-") : 
  no applicable method for 'docvars<-' applied to an object of class "c('tokenizedTexts', 'list')" > gc()
            used   (Mb) gc trigger   (Mb)  max used   (Mb)
Ncells   3008426  160.7    4703850  251.3   4647974  248.3
Vcells 178317679 1360.5  465680170 3552.9 465680096 3552.9
```
I can see `dovars` error, but let's ignore this as we defunction tokenizedTexts soon.

Issue #787, opened 2017-06-12
Create a page for features in documentation


koheiw:
There is some confusion on how `features` are specified, so we need to detail this in the documentation. It will look like a page for `valuetype`.

Issue #719, opened 2017-05-06
features handling is inconsistent between dfm and tokens


kbenoit:
This is an issue arising from #711. Currently for functions that require feature selection, we have alternative ways of selecting them depending on the function and how sequences are handled. I'd like to see these made consistent across the functions.

### Example of the inconsistency: `tokens_remove()` v. `dfm_remove()`:

```r
require(magrittr)
txt <- "one two three four"

dfm(txt, ngrams = 1:2, concatenator = " ") %>%
    dfm_remove(features = c("three", "four", "one two"))
## Document-feature matrix of: 1 document, 4 features (0% sparse).
## 1 x 4 sparse Matrix of class "dfmSparse"
##        features
## docs    one two two three three four
##   text1   1   1         1          1

tokens(txt, ngrams = 1:2, concatenator = " ") %>%
    tokens_remove(features = c("three", "four", "one two"))
## tokens from 1 document.
## Component 1 :
## [1] "one two"    "two three"  "three four"

```

In order to get `tokens_remove()` to behave as it did for `dfm_remove()`, we have to specify a different object type for `features`:
```r
tokens(txt, ngrams = 1:2, concatenator = " ") %>%
    tokens_remove(features = list(c("three", "four", "one two")))
## tokens from 1 document.
## Component 1 :
## [1] "one"        "two"        "two three"  "three four"
```
### What should be the behaviour?

We should have a discussion on this, but my preference is for the way that `dfm_select()` does it above.

### Other functions that could be affected

* `fcm_select()`
* `dfm_select()`
* `tokens_select()`
* `sequences()`
* `kwic()`
* `tokens_compound()`

koheiw:
The difference is coming from the fundamental difference between `dfm` and `tokens`. The latter supports multi-word features, so it need a list of character vectors. Space-separated features are only short-handed command (which is very handy).

PR #804, opened 2017-06-15
feature selection consistency

koheiw:
For #787 and #450


kbenoit:
A good start on this. I've changed the arguments in `dfm_select`, `dfm_remove`, `tokens_select`, `tokens_remove` to inherit the parameter from this function, similar to how we do with `valuetype`.

What we still need:
* `tokens_compound()` has a `sequences` argument - but this is similar to `features` in many ways
* `kwic()` has a `keywords` argument - similar to `features` in many ways too

The differences have to do with how multi-word features are handled. I like the convenience of splitting the patterns in `kwic()`, but I'd also like to make this all consistent.

I'd like to see a demonstration and test of all of these functions to make sure they are all consistent. We don't have to hold up this PR request for this, since it's an improvement to the current system already, but it should ultimately solve #719 too.

koheiw:
Please see the tests. Functions for `tokens` work consistently but differently from those for `dfm`. I am not sure in how we can fill the gap between `tokens` and `dfm`.

kbenoit:
OK, I have thought a lot about this now, and have devised a comprehensive proposal to deal with feature matches that covers the `tokens_select`, `dfm_select`, `tokens_compound`, and `kwic` functions. I also implement a lot of tests showing what I think is currently wanted and unwanted behaviour. 

See the file [features_comments.Rmd](https://github.com/kbenoit/quanteda/blob/issue-787/tests/misc/features_comments.Rmd) on branch `issue-787`.

This would solve the remaining issues pertaining to feature selection and make their use across different functions far more consistent.

koheiw:
It is not a bad idea to make a function to explicitly define multi-word features, but should it be `sequence()`? It looks a bit too long. How about `multi()` or even `s()` or `m()`? I am also thinking how we can sell quanteda's multi-word feature handling ability, which other packages are lacking (earlier quanteda also lost users because of this).

kbenoit:
Hmmm, I'm not a big fan of keystroke shortcuts at the expense of clarity. We could have an alias, but I prefer `sequence(c("capital", "gains", "tax"))`.

koheiw:
Long functions names are fine, if we have aliases. We have been calling "capital gains tax" a sequence, but how do other people call these?

koheiw:
NEAR operator
https://lucene.apache.org/core/2_9_4/queryparsersyntax.html
https://docs.oracle.com/cd/B19306_01/text.102/b14218/cqoper.htm#i997808

Phrase search
http://www.bl.uk/aboutus/searchtips.html
https://en.wikipedia.org/wiki/Phrase_search

Phrase match
https://www.elastic.co/guide/en/elasticsearch/guide/current/phrase-matching.html

Sequence of words
http://rdues.bcu.ac.uk/bncweb/manual/bncwebman-standard.htm
https://www.gnu.org/software/emacs/manual/html_node/emacs/Word-Search.html

Mutliword query (different one)
https://www.elastic.co/guide/en/elasticsearch/guide/current/match-multi-word.html

kbenoit:
We could call it `phrase()` instead of `sequence()` - saves two characters. 

I understand the other APIs use of phrase searching, but that confuses our ability to supply vectors of non-phrasal searches. I don't think we should parse character vectors unless they are wrapped in the `sequence/phrases` function.

koheiw:
Please have a look at https://github.com/kbenoit/quanteda/blob/issue-787/tests/misc/concatenators_comments.Rmd, there are issues in token concatenation that is strongly related to this. Addressing this issue will solve many of the inconsistency problems.

koheiw:
Working examples of a proposed design (with minor changes in `tokens_select()`):
```r
txt <- c(d1 = "a b c d e g h",  d2 = "a b e g h i j")
toks <- tokens(txt)
toks_bi <- tokens_ngrams(toks, n = 2, concatenator = ' ') # concatenator should be NULL actually

# features input is not split by whitespace with ngrams 
identical(dfm(tokens_select(toks_bi, c('a b', 'c d'))),
          dfm_select(dfm(toks_bi), c('a b', 'c d')))

toks2 <- quanteda:::tokens_concatenate(toks_bi, concatenator = '_')
toks3 <- quanteda:::tokens_concatenate(toks_bi, concatenator = '+')
toks4 <- quanteda:::tokens_concatenate(toks_bi, concatenator = ' ')

# features input is split by whitespace with unigrams
tokens_select(toks2, c('a_b', 'c_d'))
tokens_select(toks3, c('a+b', 'c+d'))
tokens_select(toks4, c('a b', 'c d')) # life is difficult indeed
tokens_select(toks4, list('a b', 'c d'))
```

kbenoit:
@kbenoit to:
- [ ] convert `features_comments.Rmd` expectations to unit tests (even if currently failing)
- [ ] give a hard think to issue of naming the "pattern" argument
- [ ] think through the dfm selection on documents (`dfm_select()`)

We all review the proposed behaviours.

Once that's done, @koheiw to modify code to pass the tests.

Issue #803, opened 2017-06-15
check calculations of textstat_collocations()


kbenoit:
Here's a stab at a log-linear model equivalency. The focus is on "capital gains tax".

```r
require(quanteda)

txt <- "A gains capital B C capital gains A B capital C capital gains tax gains tax gains B gains C capital gains tax"
toks <- as.character(tokens(txt, ngrams = 3, concatenator = " "))

require(data.table)
toks_df <- data.table(do.call(rbind, strsplit(toks, " ")), stringsAsFactors = FALSE)
names(toks_df) <- paste0("word", 1:3)

# replace non-target words with "other"
targets <- c("capital", "gains", "tax")
toks_df[word1 != targets[1], word1 := "other"]
toks_df[word2 != targets[2], word2 := "other"]
toks_df[word3 != targets[3], word3 := "other"]
toks_df[, n := 1]
toks_df_n <- toks_df[, list(n = sum(n)), by = c("word1", "word2", "word3")]
toks_df_n
     # word1 word2 word3  n
# 1:   other gains other  3
# 2:   other other other 12
# 3: capital other other  2
# 4: capital gains other  1
# 5: capital gains   tax  2
# 6:   other gains   tax  1


# MASS::loglm(n ~ word1:word2:word3, data = toks_df)

stats::loglin(table(toks_df[, 1:3]), margin = 1:3)
# 2 iterations: deviation 1.776357e-15 
# $lrt
# [1] 10.91587
# 
# $pearson
# [1] 16.93125
# 
# $df
# [1] 4
# 
# $margin
# [1] "word1" NA      NA  

textstat_collocations(tokens(txt), method = "lr", min_size = 3, max_size = 3)[1:3, ]
#         collocation length count       G2
# 1   C capital gains      3     3 20.17961
# 2   gains tax gains      3     2 11.18707
# 3 capital gains tax      3     2 10.91587

textstat_collocations(tokens(txt), method = "chi2", min_size = 3, max_size = 3)[1:3, ]
#         collocation length count       X2
# 1   C capital gains      3     3 43.95563
# 2   gains tax gains      3     2 23.64158
# 3 capital gains tax      3     2 16.93125
```

HaiyanLW:
[...]

PR #809, opened 2017-06-20
Solve issue #803

HaiyanLW:


kbenoit:
@haiyan as we discussed, please

- [ ] remove `nested` argument
- [ ] make `textstat_collocations()` call the `sequences()` only, not the older `collocations2()`

Issue #336, opened 2016-11-27
topfeatures per-document option


adamobeng:

koheiw:
It is easy to implement this, but how it should look like?
```r
topfeatures(mx)
```
or
```r
apply(mx, 1, topfeatures)
```

stefan-mueller:
Good idea! I'd prefer the first option as the structure seems to be more similar to the general **quanteda** syntax.

kbenoit:
Output would be a sorted list, much like:

```r
sink("/dev/null"); example(textstat_simil); sink()
head(as.list(s1), 2)
# $`1993-Clinton`
# 1997-Clinton   2009-Obama   2013-Obama    2001-Bush    2005-Bush   2017-Trump 
#    0.6660288    0.6288108    0.6278986    0.5333062    0.5092769    0.5033731 
# 
# $`1997-Clinton`
#   2009-Obama   2013-Obama 1993-Clinton    2001-Bush   2017-Trump    2005-Bush 
#    0.6954614    0.6826705    0.6660288    0.6434983    0.6136180    0.5535663 
```
Possibilities for syntax:
1. For real consistency, we deprecate `topfeatures()` and provide a `textstat_frequency()` function. 
2. Use the `by_document = TRUE` as @koheiw suggests
3. Use `by = docnames(x)` as a default, but allow this to be a grouping variable, including docvar names.

Could of course combine 1 and (2 or 3).

Not pressing on the groups since since we already have a `dfm_group()`, and the `by_document` option would give the same result if `dfm_group()` were run first.

PR #812, opened 2017-06-21
Issue 336

koheiw:
Solves #336


koheiw:
Codecov is complaining about this part of the code in topfeatures.R:
```r
if (is.resampled(x)) {
subdfm <- x[, order(colSums(x[,,1]), decreasing = decreasing), ]
subdfm <- subdfm[, 1:n, ]   # only top n need to be computed        return(data.frame(#features=colnames(subdfm),
            freq=colSums(subdfm[,,1]),
            cilo = apply(colSums(subdfm), 1, stats::quantile, (1 - ci) / 2),
            cihi = apply(colSums(subdfm), 1, stats::quantile, 1 - (1 - ci) / 2)))
}
```
I have never seen resampled dfm objects, and there is actually no `resample()` method for dfm in resample.R. Should I keep or delete this?

kbenoit:
`is.resampled()`  is not currently working, since I never finished it. 

just redefine the function for now (`resample.R`) to return `FALSE` and it should solve the codecov issue.

koheiw:
`topfeatures(x, groups = ...)` is one of the options that proposed, so I am perfectly fine with this.

kbenoit:
cool, please make the change then. Just wrap call dfm_groups first for a new groups argument, then use your code as if by_document = TRUE, and remove by_document from the argument list. Simple = best, and using similar arguments across functions = simpler.

Issue #813, opened 2017-06-22
optimize document grouping in dfm


koheiw:
`dfm()` has groups argument, but it creates a full-dfm to apply `dfm_compress()`. This inermidiate dfm becomes really large, slowing down construction of a dfm from large a tokens. `dfm()` should covert document IDs before making a dfm when groups are given.

Issue #498, opened 2017-01-22
what is fastest sequence of internal dfm operations?


kbenoit:
Right now, a dfm is created and then various operations use the `dfm_*` methods for applying dictionaries, feature selection, etc. Only `tolower` is applied before constructing the matrix. 

It would be better, operationally, to use `tokens_lookup` pre-dfm, rather than `dfm_lookup` post-dfm, because this will work with ngram dictionary values. The question: what are the performance differences, for large objects?

Also what are the performance implications for the `tokens_select` versus `dfm_select`?

Depending on the tests, we could alter the sequence of the dfm construction and possibly improve efficiency. 

Finally, on putting `tolower` pre-tokenisation and pre-dfm:
We decided that (many versions ago) based on the speed of lowercasing a whole text rather than lowercasing tokenzedTexts objects. But this may no longer be true now that we only need to lowercase types (but then also to recompile the hash table). Worth testing.

PR #814, opened 2017-06-22
Issue 813

koheiw:
I added `tokens_group()` to group documents before making dfm for #813. It is used when groups is given to `dfm()`)

```r
> toks <- readRDS('data_tokens_sydney_v2.RDS')
> 
> print(object.size(toks), units = 'MB')
1332.9 Mb
> 
+     new = dfm(toks, groups = docvars(toks, 'year'), tolower = FALSE),
+     old = dfm_group(dfm(toks, tolower = FALSE), groups = docvars(toks, 'year')),
+     times = 10
+ )
Unit: seconds
 expr       min        lq     mean    median       uq      max neval
  new  9.233766  9.471829 10.68980  9.831215 10.88419 16.89836    10
  old 11.011277 11.133569 11.40997 11.318823 11.47242 12.59041    10
> 
> null <- gc(reset = TRUE)
> dfm(toks, groups = docvars(toks, 'year'), tolower = FALSE)
Document-feature matrix of: 31 documents, 362,824 features (80% sparse).
> gc()
            used  (Mb) gc trigger   (Mb)  max used   (Mb)
Ncells   9770401 521.8   14442815  771.4  14442815  771.4
Vcells 117340537 895.3  559296049 4267.1 694770055 5300.7
> 
> null <- gc(reset = TRUE)
> dfm_group(dfm(toks, tolower = FALSE), groups = docvars(toks, 'year'))
Document-feature matrix of: 31 documents, 362,824 features (80% sparse).
> gc()
            used  (Mb) gc trigger   (Mb)  max used   (Mb)
Ncells   9771046 521.9   14442815  771.4   9989286  533.5
Vcells 117343054 895.3  632020472 4822.0 779315158 5945.8
```


R/dfm.R
```diff
@@ -436,14 +447,15 @@ compile_dfm.tokens <- function(x, verbose = TRUE) {
...
     if (attr(x, 'padding')) {
         types <- c("", types)
-        index <- index + 1
+        features_index <- features_index + 1
     }
+    documents_index <- rep(seq_len(length(x)), lengths(x))
     
-    temp <- sparseMatrix(j = index, 
-                         p = cumsum(c(1, lengths(x))) - 1, 
+    temp <- sparseMatrix(i = documents_index,
```
kbenoit: This is an older way we constructed the dfm, and less efficient than using the sparse row format directly (here it's row, but gets converted to sparse column format). I see no reason to go back.

R/tokens_group.R
```diff
@@ -0,0 +1,15 @@
+#' recombine documents tokens by groups
+#' 
+#' @param x \link{tokens} object
```
kbenoit: this needs 
a) more thorough documentation (which I could supply if you wish)
b) needs to be object-oriented - define a generic then an `@noRd`ed `tokens_group.tokens` method
c) needs to work identical to the dfm method, where it takes docvar names in the same way (since `tokens` objects now carry docvars too.
d) needs examples

tests/testthat/test-tokens_group.R
```diff
@@ -0,0 +1,27 @@
+context('test tokens_group.R')
```
kbenoit: need examples that group on docvars

R/dfm.R
```diff
@@ -313,11 +324,11 @@ dfm.dfm <- function(x,
     
     if (!is.null(groups)) {
```
kbenoit: if `tokens_group` + dfm is faster than dfm + `dfm_group`, then we should do that sequence internally in `dfm.tokenizedTexts()`, per issue #498.

R/dfm.R
```diff
@@ -354,17 +365,17 @@ dfm.dfm <- function(x,
...
-                     ifelse(oldNfeature - nfeature(x) != 1, "s", ""), "\n", sep = "")
+            if (n_old - nfeature(x) > 0) 
+                catm(", trimmed ", n_old - nfeature(x), " feature variant",
+                     ifelse(n_old - nfeature(x) != 1, "s", ""), "\n", sep = "")
     }
     
     # remove any NA named columns
-    if (any(naFeatures <- is.na(featnames(x))))
-        x <- x[, -which(naFeatures), drop = FALSE]
+    if (any(is.na(featnames(x))))
```
kbenoit: fine, but the old way you only needed one `is.na()` operation, whereas here you need two.

R/tokens_group.R
kbenoit: You can ignore these comments if you `tokens_group` is not exported, but still it would be nice for this to work the same way as `dfm_group`.

koheiw:
My intention to add `tokens_group()` is pre-group document in the same way as dfm does with characters with `texts()`. `tokens_group()` can be exported if we add padding at the end of the each documents to avoid false ngrams, but kept it internal only for now.

However, admittedly, use of `tokens_group()` is not the most efficient way to group documents: it is better to re-indexing groups just before making a matrix with `sparseMatrix()`. Yet, it is difficult to optimize dfm construction without major changes, because it calls dfm.dfm(). We have to simplify the process.

R/dfm.R
koheiw: I am thinking to do document grouping by re-indexing document ids, but you can keep current version.

R/dfm.R
koheiw: Just made code simpler, because I did not think this is performance critical (there should be no NA in columns). You can keep current code.

kbenoit:
@koheiw to strip tokens_group code and tests out, start a new branch and PR. Make the functionality and documentation as close to dfm_group as possible, but do not export, and mark the documentation as @keyword internal.

PR #824, opened 2017-06-29
Issue 408

kbenoit:
Adds docfreq as a way to rank features using `topfeatures()`.

- Solves #408.
- Not yet possible to use `groups` as well as `scheme = "docfreq"`, but this is easily solved.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: quanteda
-Version: 0.9.9.76
+Version: 0.9.9.77
 Title: Quantitative Analysis of Textual Data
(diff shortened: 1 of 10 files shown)


kbenoit:
@BobMuenchen your input welcome - is this the sort of behaviour you had in mind?

I'm also thinking that for consistency, `topfeatures()` might evolve into a `textstat_frequency()` function. Thoughts welcome.

kbenoit:
@HaiyanLW This was failing for some reason, until I commented out some of the sequences tests - see the change log for this PR. Not sure why, can you review and fix? 

(I think the Travis failures are a Travis problem.)

Issue #825, opened 2017-06-29
Add "textplot_frequency"


stefan-mueller:
We could add a new plotting object called, for example, `textplot_frequency()`, which is based on [this plot](http://quanteda.io/articles/pkgdown_only/plotting.html#frequency-plots) from the vignette. 

On the x-axis we would have the topfeatures, on the y-axis the term frequency. Once the `groups` option for topfeatures is implemented (see #824), we could group topfeatures by document based on `facet_wrap`. Let me know if you think it's reasonable to add this plot, and I can make a PR.

stefan-mueller:
I think we can close this as the manual of `textstat_frequency()` shows how to plot the most frequent features (also by groups).

PR #828, opened 2017-07-02
Add textplot_frequency and textplot_topfeatures

stefan-mueller:
I added the files `textplot_frequency.R` and `textplot_topfeatures.R` based on #825. I decided to write two separate functions for it, but we might think about merging both plotting functions into one (like `textplot_scale1d` for Wordfish, Wordscores and Correspondence Analysis). When `topfeatures` can return the top features by group, we should adjust `textplot_topfeatures` accordingly. 

Please add yourselves as authors (I think @adamobeng and/or @kbenoit created similar plots in the [vignette](http://quanteda.io/articles/pkgdown_only/plotting.html)) and let me know what you think.


kbenoit:
Moved to #830.

PR #844, opened 2017-07-10
Dev tbb enabled

koheiw:
Add function to check if TBB is enabled, and fix misleading start-up message.


koheiw:
I added a new condition for clang, because we need the condition for gcc. Please check if TBB is still enabled on your Mac, and `tokens_ngrams()` is working (check CPU load separately for threads).

PR #847, opened 2017-07-12
Call setThreadOptions with a given value

koheiw:
`quanteda_options()` was not doing anything with the number of thread, so I fixed this.


kbenoit:
oops! How could we (I, I am afraid) have missed putting that call into the code!? Thanks @koheiw.

Issue #836, opened 2017-07-06
tokens_lookup should not match sequences of dictionary values


kbenoit:
The *only* behaviour should be what is now matched when `multiword  = FALSE`. We should hardwire `multiword = FALSE` and remove it as an option. We don't want to allow sequences of separate values to generate matches because their order in the dictionary values should not matter. 

```r
(toks_uni2 <- as.tokens(list(d1 = c("The", "United States of America", 
                                    "China", "is", "not", "a", "country"))))
## tokens from 1 document.
## d1 :
## [1] "The"                      "United States of America" "China"                   
## [4] "is"                       "not"                      "a"                       
## [7] "country"                 

## WRONG: SHOULD BE TWO MATCHES
tokens_lookup(toks_uni2, 
              dictionary = dictionary(country = c("China", "United States of America")))
## tokens from 1 document.
## d1 :
## [1] "country"
```

This is what I would expect:
```r
## CORRECT
tokens_lookup(toks_uni2, multiword = FALSE,
              dictionary = dictionary(country = c("China", "United States of America")))
## tokens from 1 document.
## d1 :
## [1] "country" "country"
```

Since we thought - wrongly - that the alphabetical order of types might matter (it does not), I also tried this with a "Z" word. Same result so this is not something to be worried about.
```r
(toks_uni3 <- as.tokens(list(d1 = c("The", "United States of America", "Zimbabwe", "is", "not", "a", "country"))))
## tokens from 1 document.
## d1 :
## [1] "The"                      "United States of America" "Zimbabwe"                
## [4] "is"                       "not"                      "a"                       
## [7] "country"   

## WRONG: SHOULD BE TWO MATCHES
tokens_lookup(toks_uni3, 
              dictionary = dictionary(country = c("Zimbabwe", "United States of America")))
## tokens from 1 document.
## d1 :
## [1] "country"

## CORRECT
tokens_lookup(toks_uni3, multiword = FALSE,
              dictionary = dictionary(country = c("Zimbabwe", "United States of America")))
## tokens from 1 document.
## d1 :
## [1] "country" "country
```

koheiw:
If you want to remove `multiword` argument, we have to let users to wrap a dictionary with `phrase()`. To do so, I will add a `multiword` attribute to dictionary objects.

kbenoit:
[...]

PR #848, opened 2017-07-12
Issue 836

koheiw:
To solve #836, `tokens_lookup()` now finds multi-word and single-word tokens:

```r
toks <- as.tokens(list(d1 = c("The", "United States of America", "and", 
                                   "the", "United", "States", "of", "America",
                                   "are", "the", "same", "country")))
attr(toks, 'concatenator') <- ' '
tokens_lookup(toks, dictionary = dictionary(country = c("China", "United States of America")))

# tokens from 1 document.
# d1 :
# [1] "country" "country"
```
It is depends on the `concatenator` attribute of tokens. It is '_' by default.


koheiw:
We probably need to make `as.tokens()` a bit smarter in a way that it detect concatenator in the strings.

kbenoit:
or we add an argument to `as.tokens()` to let the user set it. For example:

```r
> chars <- c("united_states", "other", "word")
> as.tokens(list(doc1 = chars), concatenator = "_")
tokens from 1 document.
doc1 :
[1] "united_states" "other"         "word"  
```

Note that currently, 
```r
> str(tokens("this is a sentence"))
List of 1
 $ : chr [1:4] "this" "is" "a" "sentence"
 - attr(*, "class")= chr [1:2] "tokens" "tokenizedTexts"
 - attr(*, "types")= chr [1:4] "this" "is" "a" "sentence"
 - attr(*, "what")= chr "word"
 - attr(*, "ngrams")= int 1
 - attr(*, "concatenator")= chr "_"
 - attr(*, "padding")= logi FALSE
```
even though we never set `concatenator` through a function that uses it. Maybe we should keep it `NULL` by default, and only set it if a function for which it would apply, or the proposed new argument to `as.tokens()` above, set it.

koheiw:
It works like this 
```r
toks <- as.tokens(list(d1 = c("The", "United States of America", "and", 
                                   "the", "United", "States", "of", "America",
                                   "are", "the", "same", "country")), concatenator = ' ')
tokens_lookup(toks, dictionary = dictionary(country = c("China", "United States of America")))
```
Long time ago, the default concentrator of tokens was '' (empty string), but I change this to '_' because stirngi complains when it gets zero-length strings. It seems OK with NULL.
```r
> stringi::stri_detect_fixed(letters, ' ')
 [1] FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE
[21] FALSE FALSE FALSE FALSE FALSE FALSE
> stringi::stri_detect_fixed(letters, '')
 [1] NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA
There were 26 warnings (use warnings() to see them)
> stringi::stri_detect_fixed(letters, NULL)
logical(0)

```

koheiw:
stringi actually does nothing when NULL is given, so we need to add an extra layer of value check.
```r
stringi::stri_replace_all_fixed(letters, ' ', NULL)
```

kbenoit:
OK I like your idea of making the default concatenator a `" "` when tokens are created, but setting the default for the `concatenator` argument in `tokens_compound()` and `tokens_ngrams()` to `"_"`. This is more natural for most people with multiword pattern matches, until they have explicitly put together different elements, in which case we set it to `"_"`.

This still works so I am happy:
```r
tokens_select(tokens("a b c", n = 2), "a_b")
tokens from 1 document.
Component 1 :
[1] "a_b"
```

but shouldn't this also return `"a_b"` (and below it does not, yet):
```r
tokens_select(tokens("a b c", n = 2), dictionary(key = "a b"))
tokens from 1 document.
Component 1 :
character(0)
```

koheiw:
Now, `tokens_select()` works in either way:
```r
tokens_select(tokens("a b c", n = 2), dictionary(key = "a b"))
# > tokens_select(tokens("a b c", n = 2), dictionary(key = "a b"))
# tokens from 1 document.
# Component 1 :
#     [1] "a_b"

tokens_select(tokens("a b c", n = 2), dictionary(key = "a_b"))
# > tokens_select(tokens("a b c", n = 2), dictionary(key = "a_b"))
# tokens from 1 document.
# Component 1 :
#     [1] "a_b"
```

koheiw:
Dictionary values are white-space segmented always (internally):
```r
dict_fix <- dictionary(list(country = "united_states", 
                            law = c('law', 'constitution'), 
                            freedom = c('freedom', 'liberty')), concatenator = '_')
unlist(dict_fix)

#         country            law1            law2        freedom1        freedom2 
# "united states"           "law"  "constitution"       "freedom"       "liberty" 
```

koheiw:
In `dfm`:
```r
dfm(tokens("a b c", n = 2), dictionary = dictionary(key = "a b"))
# Document-feature matrix of: 1 document, 1 feature (0% sparse).
# 1 x 1 sparse Matrix of class "dfmSparse"
# features
# docs    key
# text1   1
dfm(tokens("a b c", n = 2), dictionary = dictionary(key = "a_b"))
# Document-feature matrix of: 1 document, 1 feature (0% sparse).
# 1 x 1 sparse Matrix of class "dfmSparse"
# features
# docs    key
# text1   1

dfm_select(dfm(tokens("a b c", n = 2)), features = dictionary(key = "a b"))
# Document-feature matrix of: 1 document, 1 feature (0% sparse).
# 1 x 1 sparse Matrix of class "dfmSparse"
# features
# docs    a_b
# text1   1
dfm_select(dfm(tokens("a b c", n = 2)), features = dictionary(key = "a_b"))
# Document-feature matrix of: 1 document, 1 feature (0% sparse).
# 1 x 1 sparse Matrix of class "dfmSparse"
# features
# docs    a_b
# text1   1
```

R/dictionaries.R
```diff
@@ -202,7 +214,7 @@ setMethod("as.list",
 #' head(dfm(data_corpus_inaugural, dictionary = mfdict))}
 #' @export
 dictionary <- function(..., file = NULL, format = NULL, 
-                       concatenator = " ", 
+                       concatenator = "_",
```
kbenoit: Why set this as a default if we always convert it to whitespace later? The purpose is not to set it, but to tell it the "input concatenator", so that we know what to convert to whitespace. We should consider renaming this to `separator`.

R/tokens.R
```diff
@@ -307,36 +307,27 @@ tokens.corpus <- function(x, ..., include_docvars = TRUE) {
...
     attr(result, "ngrams") <- 1L
-    attr(result, "concatenator") <- "_"
+    attr(result, "concatenator") <- concatenator
     attr(result, 'padding') <- FALSE
     class(result)[2] <- "tokenizedTexts"
     return(result)
 }
 
 #' @export
-#' @method as.tokens collocations
```
kbenoit: Why kill this function?

koheiw: `textstat_collocations` no longer keep the original IDs, so there is no point to return tokens-like object from the C++ side (and there is not object to coerce). If you want to keep this function, we just tokenize characters. I will do this quickly.

R/dictionaries.R
koheiw: Maybe `separator` is better.

R/tokens.R
kbenoit: We don't have the use the hashing information, so this is really just `as.tokens(x$collocation)`. But I'm not sure this ever really made sense, because it's completely unclear what a "document" should be when the input is not a list. So let's leave it out.

koheiw:
The beginning of test-dfm.R is a mess. We have to delete or tidy them up.

koheiw:
Why don't we remove `dfm.dfm()`? I do not understand why it is there.

R/dfm.R
```diff
@@ -236,18 +236,26 @@ dfm.tokenizedTexts <- function(x,
...
+        if (verbose) catm("   ... ")
+        x <- tokens_lookup(x, dictionary,
+                           exclusive = ifelse(!is.null(thesaurus), FALSE, TRUE),
+                           valuetype = valuetype,
+                           verbose = verbose)
+    }
+    
+    # use tokens_select for tokens objects
+    if (!is.null(c(remove, select))) {
+        if (verbose) catm("   ... ")
```
kbenoit: This should precede a message output in `tokens_select()` when `verbose = TRUE`. But in `tokens_select()` there is no output message. Can you please add one that will be similar to https://github.com/kbenoit/quanteda/blob/issue-836/tests/testthat/test-dfm.R#L377-L386, which you can then modify and uncomment.

tests/testthat/test-dfm.R
```diff
@@ -374,16 +374,16 @@ test_that("dfm works with relational operators", {
...
-    txt <- c(d1 = "a b c d e f g x y z",
-             d2 = "a c d x z",
-             d3 = "x y",
-             d4 = "f g")
-    expect_message(dfm(txt, remove = c('a', 'f'), verbose = TRUE),
-                   'removed 2 features and 0 documents')
-    expect_message(dfm(txt, select = c('a', 'f'), verbose = TRUE),
-                   'kept 2 features and 4 documents')
-})
+# test_that("dfm's document counts in verbose message is correct", {
```
kbenoit: See my comment above - the verbose output from this call is no longer correct since there is no message from `tokens_select()`. Needs to be added to `tokens_select()` and this test (slightly modified) restored.

tests/testthat/test-dfm.R
```diff
@@ -457,10 +457,11 @@ test_that("printing an empty dfm produces informative result (#811)", {
     
     expect_output(
         print(my_dfm),
-        "^Document-feature matrix of: 2 documents, 0 features.\\n2 x 0 sparse Matrix of class \"dfmSparse\""
+        "^Document-feature matrix of: 2 documents, 2 features \\(100% sparse\\)\\.\\n2 x 2 sparse Matrix of class \"dfmSparse\""
```
kbenoit: This is good - a more expected behaviour if you ask me. :+1:

R/dictionaries.R
```diff
@@ -153,7 +164,7 @@ setMethod("as.list",
 #'   the Linguistic Inquiry and Word Count software} \item{\code{"yoshikoder"}}{
 #'   format used by Yoshikoder software} \item{\code{"lexicoder"}}{format used
 #'   by Lexicoder} \item{\code{"YAML"}}{the standard YAML format}}
-#' @param concatenator the character in between multi-word dictionary values. 
+#' @param separator the character in between multi-word dictionary values. 
 #'   This defaults to \code{"_"} except LIWC-formatted files, which defaults to
```
kbenoit: I think we should default it to `" "`. What dictionary format do we know uses `_` for the separator?

R/dictionaries.R
```diff
@@ -202,7 +213,7 @@ setMethod("as.list",
 #' head(dfm(data_corpus_inaugural, dictionary = mfdict))}
 #' @export
 dictionary <- function(..., file = NULL, format = NULL, 
-                       concatenator = " ", 
+                       separator = "_",
```
kbenoit: Should be `" "`

kbenoit:
Re the above:

- `test-dfm.R` yes it's a mess. I prefer data objects to be defined in the testing block, but these were some of the very first unit tests we wrote. The 🍝 needs untangling...

- `dfm.dfm()` definitely needs to exist - it not only allows one to construct a dfm from a dfm input using more arguments, but also it's used in the dfm code I think. In general I like the principle of the core constructor objects being able to take their own class as input. See https://github.com/kbenoit/quanteda/issues/690.

tests/testthat/test-dfm.R
koheiw: Good

tests/testthat/test-tokens_select.R
```diff
@@ -226,19 +226,19 @@ test_that("tokens_select works as expected for bigrams selected on char, list of
 
 test_that("tokens_select works correctly with collocations objects", {
     expect_equal(
-        as.list(tokens_select(toks_uni, coll_bi)),
+        as.list(tokens_select(toks_uni, coll_bi$collocation)),
```
kbenoit: None of these tests are now testing collocations objects, since by slicing out the `$collocation` you are just returning a character vector. Please restore the test objects by removing `$collocation` so we can check the appropriate behaviour, which is to treat this as equivalent to `phrase(coll_bi$collocation)`.

tests/testthat/test-tokens_select.R
```diff
@@ -226,23 +226,24 @@ test_that("tokens_select works as expected for bigrams selected on char, list of
...
         list(d1 = character(), d2 = character())
     )
     expect_equal(
-        as.list(tokens_select(toks_bi, coll_bi)),
+        as.list(tokens_select(toks_bi, coll_bi$collocation)),
         list(d1 = c("a b", "e g", "g h"), d2 = c("a b", "e g", "g h"))
     )
-    expect_silent(
-        tokens_select(toks_bi, coll_bi)
+    expect_equal(
```
kbenoit: I added this knowing it fails, but once you fix the behaviour it will pass.

PR #853, opened 2017-07-14
Fix max load factor2

koheiw:
Optimize hash tables by setting max_load_factor in C++. In some cases, hash table construction becomes 10x faster, making lead time of `tokens_*()` much shorter.


kbenoit:
On Windows I'm still getting
```r
* checking compiled code ... NOTE
File 'quanteda/libs/x64/quanteda.dll':
  Found no calls to: 'R_registerRoutines', 'R_useDynamicSymbols'

It is good practice to register native routines and to disable symbol
search.

See 'Writing portable packages' in the 'Writing R Extensions' manual.
```
Can you make sure all functions are registered in `src/init.c`?

PR #855, opened 2017-07-15
Issue 854

koheiw:
Not yet removed the documents argument from `dfm_select()`.


R/dfm_compress.R
```diff
@@ -74,14 +79,18 @@ dfm_compress.dfmSparse <- function(x, margin = c("both", "documents", "features"
 #' # equivalent
 #' dfm(mydfm, groups = "grp")
 #' dfm(mydfm, groups = c(1, 1, 2, 2))
-dfm_group <- function(x, groups = NULL) {
+dfm_group <- function(x, groups = NULL, fill = FALSE) {
```
kbenoit: Here I've added a `fill` argument, but it doesn't do anything (yet) - you will need to review the tests and add the behaviour to match. But first: See comments on the tests below.

R/dfm_compress.R
```diff
@@ -74,14 +79,18 @@ dfm_compress.dfmSparse <- function(x, margin = c("both", "documents", "features"
...
 }
 
 #' @noRd
 #' @export
-dfm_group.dfm <- function(x, groups = NULL) {
-    dfm(x, groups = groups, tolower = FALSE)
+dfm_group.dfm <- function(x, groups = NULL, fill = FALSE) {
+    #dfm(x, groups = groups, tolower = FALSE)
+    if (is.character(groups) & all(groups %in% names(docvars(x)))) {
+        groups <- interaction(docvars(x)[, groups], drop = TRUE)
```
kbenoit: This is the line that automatically makes a factor of any input vectors, and that may not be desired since the documents are then re-ordered in the factor order. See tests below.

R/dfm_select.R
```diff
@@ -70,37 +64,32 @@
 #' txts <- c("This is text one", "The second text", "This is text three")
 #' (dfm1 <- dfm(txts[1:2]))
 #' (dfm2 <- dfm(txts[2:3]))
-#' (dfm3 <- dfm_select(dfm1, dfm2, valuetype = "fixed", padding = TRUE, verbose = TRUE))
+#' (dfm3 <- dfm_select(dfm1, dfm2, valuetype = "fixed", verbose = TRUE))
 #' setequal(featnames(dfm2), featnames(dfm3))
 #' 
-dfm_select <- function(x, pattern = NULL, documents = NULL,
```
kbenoit: As you will see, I killed the document selection from `dfm_select()`. If we want to implement document selection, I suggest `dfm_subset()` as described in https://github.com/kbenoit/quanteda/issues/854#issuecomment-315523396.

tests/testthat/test-dfm_group.R
```diff
@@ -56,3 +56,94 @@ test_that("dfm.character groups works (#794)", {
         dfm(corp, groups = "grp")
     )
 })
+
```
kbenoit: Here I've rewritten (slightly) your tests and added others. These will fail until the function is changed.

tests/testthat/test-dfm_group.R
```diff
@@ -56,3 +56,94 @@ test_that("dfm.character groups works (#794)", {
...
+test_that("test dfm_group with non-factor grouping variable, with fill", {
+    grp <- c("D", "D", "A", "C")
+    corp <- corpus(c('a b c c', 'b c d', 'a', "b d d"),
+                   docvars = data.frame(grp = grp, stringsAsFactors = FALSE))
+    testdfm <- dfm(corp)
+    expect_equal(
+        as.matrix(dfm_group(testdfm, groups = "grp", fill = FALSE)),
+        matrix(c(1,0,0,0, 0,1,0,2, 1,2,3,1), byrow = TRUE, nrow = 3, 
+               dimnames = list(docs = c("A", "B", "D"), features = letters[1:4]))
```
kbenoit: Oops, this should be `docs = c("A", "B", "D")` ...  But it's an open question as to whether this should be ordered or not. My view is that they should, since we can't assume there is any particular order in the groups assigned to the documents, simply in the way that they are assigned.

tests/testthat/test-dfm_group.R
```diff
@@ -56,3 +56,94 @@ test_that("dfm.character groups works (#794)", {
...
+    testdfm <- dfm(corp)
+    expect_error(
+        dfm_group(testdfm, groups = c(1, 1, 2, 3, 3), fill = FALSE),
+        "groups must name docvars or provide data matching the documents in x"
+    )
+    expect_error(
+        dfm_group(testdfm, groups = c(1, 1, 2, 3, 3), fill = TRUE),
+        "groups must name docvars or provide data matching the documents in x"
+    )
+    expect_error(
```
kbenoit: It should be an error to produce:
```r
dfm_group(testdfm, groups = c(1, 1, 2, 3, 4), fill = TRUE)
# Document-feature matrix of: 4 documents, 8 features (71.9% sparse).
# 4 x 8 sparse Matrix of class "dfmSparse"
#     features
# docs a B c D e BBB Aaaa cc
#    1 2 1 2 2 2   1    0  0
#    2 0 0 0 0 0   1    1  1
#    3 0 0 0 0 0   0    0  0
#    4 0 0 0 0 0   0    0  0
```
We should always check that the `groups` object is either a docvar name, or equal in length (or nrow) to the number of documents.

tests/testthat/test-dfm_select.R
```diff
@@ -161,34 +159,34 @@ test_that("test dfm_select with ngrams concatenated with whitespace", {
...
-test_that("test dfm_select with dates in document names", {
-    
-    dates <- as.character(seq.Date(as.Date('2017-01-01'), as.Date('2017-12-31'), 1))
-    txts <- paste('text', seq_along(dates))
-    names(txts) <- dates
-    paddeddfm <- dfm_select(dfm(txts)[1:10,], documents = dates, valuetype = 'fixed', padding = TRUE)
-    expect_equal(ndoc(paddeddfm), 365)
-    
-})
+# test_that("test dfm_select with documents", {
```
kbenoit: These are commented out now but not deleted, since they could form the basis for `dfm_subset()` tests.

koheiw:
When `fill = TRUE` users can specify order in factor levels, but how to sort when `fill = FALSE`? Alphabetic or reverse-alphabetic, or numeric? I think it is better to keep the original order when `fill = FALSE`. If users really want to sort, they should set `fill = TRUE` and provide a vector with the same length as dfm (then it does add anything), or just use `[]`.

kbenoit:
Alphabetical is fine - this is how `groups` has been working in other functions, and how factors are ordered when levels are not supplied manually. However, the last one below is inconsistent - I am suggesting they be made consistent.
```r

groupvec <- c("grp2", "grp2", "grp1", "grp1")
mycorpus <- corpus(c("a a b", "a b c c", "a c d d", "a c c d"), 
                   docvars = data.frame(grp = groupvec))

texts(mycorpus, groups = "grp")
##               grp1               grp2 
## "a c d d  a c c d"   "a a b  a b c c" 
dfm(mycorpus, groups = "grp")
## Document-feature matrix of: 2 documents, 4 features (25% sparse).
## 2 x 4 sparse Matrix of class "dfmSparse"
##       features
## docs   a c d b
##   grp1 2 3 3 0
##   grp2 3 2 0 2

texts(mycorpus, groups = groupvec)
##               grp1               grp2 
## "a c d d  a c c d"   "a a b  a b c c" 
dfm(mycorpus, groups = groupvec)
## Document-feature matrix of: 2 documents, 4 features (25% sparse).
## 2 x 4 sparse Matrix of class "dfmSparse"
##       features
## docs   a c d b
##   grp1 2 3 3 0
##   grp2 3 2 0 2

dfmtest <- dfm(mycorpus)
dfm_group(dfmtest, groups = "grp")
## Document-feature matrix of: 2 documents, 4 features (25% sparse).
## 2 x 4 sparse Matrix of class "dfmSparse"
##       features
## docs   a b c d
##   grp1 2 0 3 3
##   grp2 3 2 2 0
dfm_group(dfmtest, groups = groupvec)
## Document-feature matrix of: 2 documents, 4 features (25% sparse).
## 2 x 4 sparse Matrix of class "dfmSparse"
##       features
## docs   a b c d
##   grp2 3 2 2 0
##   grp1 2 0 3 3

texts(mycorpus, groups = "grp")
dfm(mycorpus, groups = "grp")

texts(mycorpus, groups = groupvec)
dfm(mycorpus, groups = groupvec)

dfmtest <- dfm(mycorpus)
dfm_group(dfmtest, groups = "grp")
dfm_group(dfmtest, groups = groupvec)
```

kbenoit:
I suggest we create a separate manual page for `groups` (similar to `valuetype` and `pattern`) and use `@inheritParams groups` in the Roxygen header for any function that uses the argument. And we make sure it's consistent across the functions where it's used.

kbenoit:
In my latest commits, I consolidated the `groups` parameter documentation in single new .Rd entry and updated the functions that use it to inherit this parameter description.

What about the idea of letting groups be a dfm for `dfm_group` only, with the behaviour:
```r
dfm1 <- dfm(c(d1 = "a b b c", d2 = "b b c d"))
dfm2 <- dfm(c(d1 = "x y z", d2 = "a b c c d", d3 = "a b c", d4 = "x x x"))

dfm_group(dfm1, groups = dfm2)
# Document-feature matrix of: 3 documents, 4 features (16.7% sparse).
# 3 x 4 sparse Matrix of class "dfmSparse"
#     features
# docs a b c d
#   d1 1 2 1 0
#   d2 1 3 3 2
#   d3 0 0 0 0
#   d4 0 0 0 0
```

kbenoit:
Actually I like the above idea better for `dfm_subset()`, since this is the document-level analog to `dfm_select()` (which as of this PR, operates only on features). `dfm_groups()` would then not need this, unless documents in the `x` dfm had duplicated names, in which case `dfm_compress()` would solve that problem. So this feature is probably not needed and arguably inappropriate in `dfm_group()`.

We can however use a similar argument `fill` in `dfm_subset()` that will create new empty "documents" based on factor levels in `groups` if and only if the grouping variable(es) are factors.

Issue #785, opened 2017-06-10
Feature suggestion: mutual information measure and keyness plot


methodds:
Hi,

I'd like to suggest adding a mutual information measure to quanteda. This could just be another measure that users are able to choose in ```textstat_keyness```. In addition, it would also be awesome to have a plot for keyness term comparisons which could look similar to this one (from [Grimmer 2010](http://web.stanford.edu/~jgrimmer/ExpAgendaFinal.pdf)):

![image](https://user-images.githubusercontent.com/7460499/27001508-fca10fc8-4dcb-11e7-9d97-af68d92a5533.png)

kbenoit:
Good suggestion! We could add mutual information to `textstat_keyness()` very easily. The plot would be a new `textplot_keyness()` that takes the return of a `textstat_keyness()` and plots it.

HaiyanLW:
Is this what you want?
![image](https://user-images.githubusercontent.com/20244806/27551702-613f40e4-5a9c-11e7-8a0a-b5b0225ea743.png)

methodds:
Yes, something like that, looks nice :) Would it also be possible to include "directions" somehow? 

What I mean by this is that the plot would support showing terms with the highest keyness scores that are much more likely (by whatever metric) in corpus 1 , but in contrast also show terms that are much more likely for corpus 2 ? An alternative would be to plot this contrast with a wordcloud, as proposed in #528.

HaiyanLW:
Here's the dual-plot: Blue dots associates with Document 1 and red dots accociated with the 2nd Document. The code is in branch `issue-785`
```r
period <- ifelse(docvars(data_corpus_inaugural, "Year") < 1945, "pre-war", "post-war")
mydfm <- dfm(data_corpus_inaugural, groups = period)
head(mydfm) # make sure 'post-war' is in the first row
```
```
Document-feature matrix of: 2 documents, 9,232 features (34.6% sparse).
(showing first 2 documents and first 6 features)
          features
docs       chief justice    , mr    . vice
  post-war    12      40 2194 25 1804   15
  pre-war     24      97 4832  6 3141    3
```
```r
result <- textstat_keyness(mydfm)
 textplot_keyness(result, showBothDoc = TRUE)
```
![image](https://user-images.githubusercontent.com/20244806/27731931-0bbd53c0-5d87-11e7-8336-607c66d4a4e0.png)

methodds:
cool thanks :) I guess the only thing missing is the label for the second y axis?

[...]

PR #860, opened 2017-07-17
Add textplot_keyness function

HaiyanLW:
Solves issue #785


kbenoit:
Looking good! A couple of things:

- looks like the words stay the same, but the bars invert, when `sort = TRUE` and `sort = FALSE`. Needs to be fixed (and manually compared to the data.frame output to be verified).
- I'd prefer the default sort for the reference words to be the highest keyness words at the bottom, not a the top of the reference group.
- do you need the two new graphics package dependencies? I could not see where they were actually used. The fewer dependencies, the better.
- Please add tests.

HaiyanLW:
Done
- [x] removed `sort` as it should be sorted for keyness
- [x] rearranged the order of the reference words
- [x] Removed **grid** and **gtable**-- forgot to remove them when I deleted the function using them.
- [x] Added test for `show_reference`

kbenoit:
Looks great @HaiyanLW! I suggest one last change:

* make the default labels for the legend not "Target" and "Reference", but rather the docnames of the categories from the dfm that are being compared. Target should always get the docname of the target row. Reference can be "Reference" if there is more than one non-target row, or the docname of the reference if it is a single row. This means that for the example plot, it would be "Trump" and "Obama" for the Target and Reference groups, respectively.

PR #861, opened 2017-07-17
Add dfm_subset, fix indexing for dfm docvars

### LINKED ISSUE #859: add dfm_subset() function
Proposal:

```r
dfm_subset(x, subset, select, fill = FALSE, ...)
```
The `select` will work on docvars or some other supplied variable equal in length to `ndoc(x)`, but *not features of the dfm `x`*. Not sure why this would be useful, but we can keep it anyway for consistency with `base::subset()`. This could make it useful in trimming docvars from a dfm. Or, we just remove the `select` argument because it is confusing.

`subset` can also be a dfm, in which case the returned dfm will conform to the documents in the dfm supplied in `subset`, similar to how we can use `dfm_select(x, pattern = someotherdfm)` to make the output dfm match the feature set of `x` to that of `someotherdfm`. It would work like this
```r
dfm1 <- dfm(c(d1 = "a b b c", d2 = "b b c d"))
dfm2 <- dfm(c(d1 = "x y z", d2 = "a b c c d", d3 = "a b c", d4 = "x x x"))

dfm_subset(dfm1, subset = dfm2)
# Document-feature matrix of: 3 documents, 4 features (16.7% sparse).
# 3 x 4 sparse Matrix of class "dfmSparse"
#     features
# docs a b c d
#   d1 1 2 1 0
#   d2 0 2 1 1
#   d3 0 0 0 0
#   d4 0 0 0 0
```

Here's how to do that, if we decide to:
```r
dfm1a <- dfm1[which(docnames(dfm2) %in% docnames(dfm1)), ]
dfm_group(dfm_trim(dfm1a), 
          groups = factor(docnames(dfm1a), levels = docnames(dfm2)),
          fill = TRUE)
# Document-feature matrix of: 4 documents, 4 features (62.5% sparse).
# 4 x 4 sparse Matrix of class "dfmSparse"
#     features
# docs b c a d
#   d1 2 1 1 0
#   d2 2 1 0 1
#   d3 0 0 0 0
#   d4 0 0 0 0
```
Note that here, the feature order is changed, but this will be solved once we resolve #858.

See also:
* the discussion starting here https://github.com/kbenoit/quanteda/pull/855#issuecomment-315606312
* the discussion in #854.

kbenoit:
Fixes #859 

Also fixes a problem in dfm row indexing that was not also slicing out docvars.


kbenoit:
The fill didn't make sense so I left that for `dfm_groups()` only.

koheiw:
I notice that `select` argument is not working, so I added a test for that.

koheiw:
I also could not understand why `dfm_select` is implemented such a strange way. I would rather use base functions:
```r
x <- dfm(data_corpus_inaugural)
vars <- docvars(x)

x_sub <- subset.matrix(x, vars[['Year']] > 1980 & vars[['Year']] < 2018)
vars_sub <- subset.data.frame(vars, vars[['Year']] > 1980 & vars[['Year']] < 2018, select = 'President')

attr(x_sub, 'docvars') <- vars_sub
str(x_sub) # still dfmSparse
```
I think we should use `subset.data.frame` on docvars everywhere instead of `[]`.

koheiw:
Yes, `dfm_subset`.

kbenoit:
You'd rather subset the dfm using row slicing using `"["`, which is fine and still completely possible. The advantage of `dfm_subset()` is 

a) implement the exact subset behaviour we prefer, if we need a slightly different behaviour from `"["`. This is why we use `dfm_select()` rather than `"["` on the columns, for instance, because it gives us valuetype matches and other options (even future ones). For instance, the NSE for subsetting conditions is for docvars, not for column elements of the matrix. With `subset.data.frame` this would operate on the columns (features) not the docvars.

b) it's consistent with `corpus_subset()`. Remember (if you can!) that we started with overloading the `subset()` generic for **quanteda** objects, until we did the Big Rename.

I'm happy to remove `select` from the `dfm_subset()`, but I think it does little harm.

koheiw:
The example is for the internal working of `dfm_subset`. I agree that it is a good idea to have `dfm_subset`. The code should have been like this:
```r
dfm_subset(x, subset, select) {
    vars <- docvars(x)
    x <- base::subset.matrix(x, subset)
    vars <- base::subset.data.frame(vars, subset, select = select)
    attr(x, 'docvars') <- vars
    return(x)
}
```
This obviously does not work in subsetting by docvars but my point here is to use base functions for subsetting to make `dfm_subset` (`corpus_subset` as well) consistent with functions in other packages.

koheiw:
Now I know where the code of `dfm_subset()` is from, but this should do the same thing:
```r
dfm_subset <- function(x, subset, select) {
    vars <- docvars(x)
    l <- eval(substitute(subset),vars,parent.frame())
    if (!is.logical(l)) l <- subset
    x <- base::subset.matrix(x, subset = l)
    attr(x, 'docvars') <- base::subset.data.frame(vars, subset = l, select = select)
    return(x)
}

dfm_subset(x, Year > 1990, 'President')
dfm_subset(x, Year > 2000, 'President')
dfm_subset(x, rep(TRUE, ndoc(x)), 'President')
docvars(dfm_subset(x, rep(TRUE, ndoc(x)), 'President'))
```

kbenoit:
It's tricky with non-standard evaluation, because calling that internally won't evaluate the variables unless quoted, and then it's not the same as subset. It's consistent now because the code is taken straight from `subset.data.frame.()`.

I think we just need to fix select, and we are good. I'm trying to do that now.

koheiw:
`eval(substitute())` the greatest and also the weirdest thing in R.

kbenoit:
I just fixed it, and it works with both `select = "President"` and `select = President`, so is consistent now with the base `subset` functions, except that for `dfm_select()` select works on docvars not features. I prefer doing it at this level because we adapt the subset code rather than rely on it.

`eval(substitute())` gets really strange when you send it variables defined in the environments. Would the approach you propose work with both of these?
```r
selvar <- "President"
docvars(dfm_subset(dfmtest, select = selvar))
#              President
# 1981-Reagan     Reagan
# 1985-Reagan     Reagan
# 1989-Bush         Bush
# 1993-Clinton   Clinton
# 1997-Clinton   Clinton
# 2001-Bush         Bush
# 2005-Bush         Bush
# 2009-Obama       Obama
# 2013-Obama       Obama
# 2017-Trump       Trump
tempfun <- function() {
    selvar2 <- "President"
    docvars(dfm_subset(dfmtest, select = selvar))
}
tempfun()
#              President
# 1981-Reagan     Reagan
# 1985-Reagan     Reagan
# 1989-Bush         Bush
# 1993-Clinton   Clinton
# 1997-Clinton   Clinton
# 2001-Bush         Bush
# 2005-Bush         Bush
# 2009-Obama       Obama
# 2013-Obama       Obama
# 2017-Trump       Trump
```

koheiw:
No, it doesn't. We need a trick for `select` argument as well.
```r
x <- dfm(data_corpus_inaugural)

selvar <- "President"
docvars(dfm_subset(x, select = selvar))

# Error in eval(substitute(subset), vars, parent.frame()) : 
#     argument is missing, with no default 

tempfun <- function() {
    selvar2 <- "President"
    docvars(dfm_subset(x, select = selvar2))
}
tempfun()

# Error in eval(substitute(subset), vars, parent.frame()) : 
#     argument is missing, with no default 
```

kbenoit:
It does work - I think you haven't built with my last commit.

kbenoit:
@koheiw If you're happy with it now, approve and merge (or I can merge).

PR #865, opened 2017-07-20
Added Yates correction and min_count to testplot_keyness

### LINKED ISSUE #785: Feature suggestion: mutual information measure and keyness plot

HaiyanLW:
fix  #785

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -41,12 +41,12 @@ qatd_cpp_fcm <- function(texts_, n_types, count, window, weights, ordered, tri, 
     .Call(`_quanteda_qatd_cpp_fcm`, texts_, n_types, count, window, weights, ordered, tri, nvec)
 }
 
-qatd_cpp_sequences_old <- function(texts_, words_, types_, count_min, len_max, nested, ordered = FALSE) {
-    .Call(`_quanteda_qatd_cpp_sequences_old`, texts_, words_, types_, count_min, len_max, nested, ordered)
-}
-
 qatd_cpp_sequences <- function(texts_, types_, count_min, sizes_, method, smoothing) {
     .Call(`_quanteda_qatd_cpp_sequences`, texts_, types_, count_min, sizes_, method, smoothing)
+}
+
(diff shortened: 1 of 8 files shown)


kbenoit:
Sorry for the extra work, but I think it made more sense before either to simply automatically apply Yates (and Williams) for Chi2 (and G2), or to provide the option for correction TRUE or FALSE.

HaiyanLW:
Currently, we offer option for user to set the value of correction, and the default is Yates (`0.5`); and for each word the correction is applied only if its smallest expected frequency is less than `5` -- which means that the correction may still not be applied to each word. Since there's an argument about whether `Yates` over-corrects, user can choose not applying correction by set it to `0` or a smaller number. 
So you prefer to provide the option for `correction = TRUE` instead of a free value setting? And if so, do we still check the smallest expected frequency? 
And for `pmi`?

kbenoit:
how about:

```r
correction = c("default", "yates", "williams", "none")
```
where we explain in the details that "default" is yates for chi2, williams for lr, and none for pmi?

These options allow the user to override. So the main change you would need to is change this from being numeric to being (back to) the character label, and to set the conditions. And to check whether williams for chi2 makes any sense. I don't think yates or williams for pmi is applicable.

HaiyanLW:
Agree. 
Didn't find any reference of corrections for `pmi`, though we could add `normalised pmi` in `measure`.

kbenoit:
normalised as in `tf(x, "prop")`? Sounds reasonable.

HaiyanLW:
As `npmi≡pmi/(−logp(x,y))`, so pmi is normalized between [-1,1]

HaiyanLW:
William's works for `chi2` as well, see http://www.biostathandbook.com/small.html

kbenoit:
even better

Issue #507, opened 2017-01-24
Need a complete regex functionality audit


kbenoit:
We want to make sure that all of our regex flavours are the same. In particular, we want to move exclusively to **stringi** for pattern matching, rather than using the base R `grep` family. The difference: base R does not recognise Unicode shortcuts such as the character classes.

Example:

```r
toks <- as.character(quanteda::tokens("This: a sentence, with several words."))
## toks
## [1] "This"     ":"        "a"        "sentence" ","        "with"     "several"  "words"    "."       
which(stringi::stri_detect_regex(toks, "\\p{P}"))
[1] 2 5 9
```
but
```r
grep("\\p{P}", toks)
## Error in grep("\\p{P}", toks) : 
##   invalid regular expression '\p{P}', reason 'Invalid contents of {}'
```

To test this we should:

1.  Try using Unicode pattern matches for every function that takes a `valuetype` argument.  
2.  Search the whole source code to make sure we are not calling anything from the `grep` family internally, even when not linked to `valuetype`.

PR #873, opened 2017-07-25
issue #507

koheiw:
Use string instead of grep() and gsub() except in code that will be moved soon.


koheiw:
codecov is complaining about the lack of tests for `textstat_wordscore()`.

kbenoit:
I'll add tests

PR #878, opened 2017-07-25
Issue 649

koheiw:


R/kwic.R
```diff
@@ -64,7 +64,7 @@ kwic.character <- function(x, pattern, window = 5, valuetype = c("glob", "regex"
 #' @export 
 kwic.corpus <- function(x, pattern, window = 5, valuetype = c("glob", "regex", "fixed"), 
                         case_insensitive = TRUE, join = FALSE, ...) {
-    kwic(texts(x), pattern, window, valuetype, case_insensitive, join, ...)
+    kwic(tokens(x, ...), pattern, window, valuetype, case_insensitive, join)
```
kbenoit: This bypasses the checks for deprecated arguments in `kwic.character()`, and means we call the same thing twice, from two different functions. Better with the original.

R/kwic.R
```diff
@@ -159,9 +158,18 @@ print.kwic <- function(x, ...) {
 #' @export
 #' @method as.tokens kwic
 as.tokens.kwic <- function(x, ...) {
-    result <- attr(x, 'tokens')
-    names(result) <- x$docname
-    docvars(result) <- data.frame('_docid' = attr(x, 'docid'),
-                                  '_segid' = attr(x, 'segid'))
+    vars <- attr(x, 'docvars')
```
kbenoit: This not only makes no real sense to me, but also violates the idea that docnames should be unique. What makes more sense is to expand the docnames as you do with the results of 

```r
head(rownames(attr(as.tokens(testkwic), "docvars")))
## [1] "1789-Washington.1" "1789-Washington.2" "1797-Adams.1"      "1797-Adams.2"      "1797-Adams.3"      "1805-Jefferson.1" 
```

But is it meaningful to coerce kwic objects to tokens?  What is the use case?

tests/testthat/test-kwic.R
```diff
@@ -396,3 +396,32 @@ test_that("deprecated keywords argument still works", {
         "fox"
     )
 })
+
+test_that("test docvar is passed through kwic()", {
+    
+    # kwic only copy original docvars
+    testkwic <- kwic(data_corpus_inaugural, 'american')
+    expect_equal(attr(testkwic, 'docvars'),
```
kbenoit: If we do attach docvars in this way to kwic objects, should there be a `docvars.kwic()`, or are you just storing the docvars in order to get them when you create `as.tokens()`?

R/kwic.R
koheiw: This is to pass docvars to kwic, we cannot do this kwic.character().

R/kwic.R
koheiw: Fix this.

tests/testthat/test-kwic.R
koheiw: Unless users want to see docvars in concordance view, I don't think we need docvars.kwic(). I get docvars through as.tokens(), which multiplies rows of docvars.

R/kwic.R
kbenoit: OK I just fixed this in e43a822.

tests/testthat/test-kwic.R
kbenoit: I suggest you write an internal `docvars.kwic()` and use that to do the multiplication of rows of docvars in `as.tokens()`. 

We could use this for the `corpus.kwic()` as well but we would need to add options to make it possible to do it as one keyword per document with the pre/keyword/post as a document, versus the current approach of having 2*(number of keywords) as corpus documents, with a variable for pre v. post.

PR #881, opened 2017-07-26
Issue 373

koheiw:
Make sure that `nrow(docvars(x)) == ndoc(x)` and `all(rownames(docvars(x)) == docnames(x))` are always true for any corpus, tokens and dfm.


kbenoit:
Merged into `issue-690` (PR#880)

PR #887, opened 2017-07-27
Dev tokens

koheiw:
Adds `tokens.tokens()`. Also gives much greater control of types of tokens to be preserved, while increasing efficiency.


R/corpus-methods-quanteda.R
```diff
@@ -208,7 +208,11 @@ docnames <- function(x) {
 docnames.corpus <- function(x) {
     # didn't use accessor documents() because didn't want to pass
     # that large object
-    rownames(x$documents)
+    if (is.null(rownames(x$documents))) {
```
kbenoit: This seems slightly dangerous, since it means we could have a corpus object with no document names, but that `docnames()` will make it seem like it does have document names. Have you tested all of the functions that rely on there being actual document names? If they all call `docnames.corpus()` then we should be OK, since this will create the names.

tests/testthat/test-dfm.R
```diff
@@ -8,10 +8,10 @@ test_that("oldest dfm test", {
                               taxation="taxation",
                               taxregex="tax*",
                               country="united_states"))
-    dictDfm <- dfm(mycorpus, dictionary = mydict, valuetype = "glob")
-    dictDfm[1:10, ]
-    thesDfm <- dfm(mycorpus, thesaurus = mydict, valuetype = "glob")
-    thesDfm[1:10, (nfeature(thesDfm)-8) : nfeature(thesDfm)]
+    dictDfm <- dictDfm <- dfm(mycorpus, dictionary = mydict, valuetype = "glob")
+    dictDfm <- dictDfm[1:10, ]
+    dictDfm <- thesDfm <- dfm(mycorpus, thesaurus = mydict, valuetype = "glob")
```
kbenoit: I'm not sure what the point of these assignments are, since they keep overwriting `dictDfm`. Shouldn't there be some `expect_*()` calls around these to make them tests?

tests/testthat/test-dfm.R
```diff
@@ -21,13 +21,13 @@ test_that("oldest dfm test", {
                     remove_punct = TRUE)
     
     
-    dfm(txt, dictionary = mydict, verbose = FALSE)
-    dfm(txt, thesaurus = mydict, verbose = FALSE)
-    dfm(txt, thesaurus = mydict, verbose = FALSE)
+    dictDfm <- dfm(txt, dictionary = mydict, verbose = FALSE)
+    dictDfm <- dfm(txt, thesaurus = mydict, verbose = FALSE)
```
kbenoit: Likewise, not sure what the point of these are, esp since one line is repeated.

tests/testthat/test-dfm.R
```diff
@@ -21,13 +21,13 @@ test_that("oldest dfm test", {
...
+    dictDfm <- dfm(txt, thesaurus = mydict, verbose = FALSE)
     
-    (txtDfm <- dfm(txt, verbose = FALSE))
-    dfm_lookup(txtDfm, mydict, valuetype = "glob") 
-    dfm_lookup(txtDfm, mydict, exclusive = FALSE, valuetype = "glob", verbose = FALSE) 
+    txtDfm <- dfm(txt, verbose = FALSE)
+    dictDfm <- dfm_lookup(txtDfm, mydict, valuetype = "glob")
```
kbenoit: ditto re above

tests/testthat/test-dfm.R
```diff
@@ -40,10 +40,10 @@ test_that("oldest dfm test", {
     
     ## need to be carefully inspected!
     txt <- "The tall brown trees with pretty leaves in its branches."
-    dfm(txt)
-    dfm(txt, stem = TRUE)
-    dfm(txt, remove = stopwords("english"))
-    dfm(txt, stem = TRUE, remove = stopwords("english"))
+    txtDfm <- dfm(txt)
```
kbenoit: here too - no tests here

tests/testthat/test-dfm_trim.R
```diff
@@ -8,15 +8,15 @@ test_that("dfm_trim", {
     nfeature(dfm_trim(preDictDfm, min_count = 7))
     nfeature(dfm_trim(preDictDfm, min_count = 0.001))
     
-    expect_equal(nfeature(dfm_trim(preDictDfm, min_count = 0.001)), 1045)
```
kbenoit: Why did these numbers change? Is it because of changes in `tokens()`?

tests/testthat/test-docnames.R
```diff
@@ -0,0 +1,12 @@
+context("test docnames")
+
+test_that("docnames always return names even if there aren't", {
+    
+    corp <- corpus(c('aaa', 'bbb', 'ccc'))
+    expect_equal(length(docnames(corp)), ndoc(corp))
+    
+    toks <- as.tokens(list('aaa', 'bbb', 'ccc'))
```
kbenoit: I think that definitionally, we should never have `corpus` objects, and certainly never `tokens` objects, with no docnames. A better method would be to define them as an S4 object with a validator that requires docnames to be non-NULL and unique.

tests/testthat/test-tokens.R
```diff
@@ -400,21 +400,65 @@ test_that("tokens works for strange spaces (#796)", {
...
+expect_equal(tokens(chars[4], remove_url = TRUE),
+             tokens(toks4, remove_url = TRUE))
+
+expect_equal(tokens(chars[5], ngrams = 1:2),
+             tokens(toks5, ngrams = 1:2))
+
+expect_equal(tokens(chars[5], ngrams = 2, skip = 1:2),
+             tokens(toks5, ngrams = 2, skip = 1:2))
+
+# This fails because hyphnated words are concatenated in toks
```
kbenoit: The behaviour of `remove_hyphens` is a bit inconsistent now, and not what the option in the Rd file describes.

from `master`:
```r
> tokens("One two-three four.", remove_hyphens = FALSE)
tokens from 1 document.
text1 :
[1] "One"   "two"   "-"     "three" "four"  "."    

> tokens("One two-three four.", remove_hyphens = FALSE, remove_punct = TRUE)
tokens from 1 document.
text1 :
[1] "One"       "two-three" "four"     

> tokens("One two-three four.", remove_hyphens = TRUE)
tokens from 1 document.
text1 :
[1] "One"   "two"   "three" "four"  "."    

> tokens("One two-three four.", remove_hyphens = TRUE, remove_punct = TRUE)
tokens from 1 document.
text1 :
[1] "One"   "two"   "three" "four" 
```

from `dev-tokens`:
```r
> tokens("One two-three four.", remove_hyphens = FALSE)
tokens from 1 document.
text1 :
[1] "One"   "two"   "-"     "three" "four"  "."    

> tokens("One two-three four.", remove_hyphens = FALSE, remove_punct = TRUE)
tokens from 1 document.
text1 :
[1] "One"   "two"   "three" "four" 

> tokens("One two-three four.", remove_hyphens = TRUE)
tokens from 1 document.
text1 :
[1] "One"   "two"   "three" "four"  "."   

> tokens("One two-three four.", remove_hyphens = TRUE, remove_punct = TRUE)
tokens from 1 document.
text1 :
[1] "One"   "two"   "three" "four" 
```

In the examples above, it's the **second** that behaves unexpectedly in `dev-tokens`, since it does remove hyphenated words, even when it is not supposed to.

R/tokens.R
```diff
@@ -178,16 +178,67 @@ tokens.character <- function(x, ...) {
 #' @rdname tokens
 #' @export
 #' @noRd
-tokens.corpus <- function(x, ..., include_docvars = TRUE) {
+tokens.corpus <- function(x, ..., hash = TRUE, include_docvars = TRUE) {
```
kbenoit: Interesting - I thought we would get errors in the CHECK if not all the signatures are the same.

R/tokens.R
```diff
@@ -178,16 +178,67 @@ tokens.character <- function(x, ...) {
...
+    regex <- c()
+    if (remove_numbers)
+        regex = c(regex, "^[\\p{N}]+$")
+    if (remove_punct)
+        regex = c(regex, "^[\\p{P}\\p{S}]+$")
+    if (remove_symbols)
+        regex = c(regex, "^[\\p{S}]+$")
+    if (remove_separators)
+        regex = c(regex, "^[\uFE00-\uFE0F\\p{Z}\\p{C}]+$") 
+    if (remove_url)
```
kbenoit: with the exception of the `remove_url`, the others are char_class and might be faster implemented that way (through `stringi::stri_remove_charclass()` - although I realise that `tokens_remove()` does not currently provide this option. Something to note for the future? Or, we could examine the regex in `tokens_remove()` and if it matches a charclass, we call that instead of the equivalent regex function. We should test the performance differences before investing time in this.

R/tokens.R
```diff
@@ -197,9 +248,18 @@ tokens.corpus <- function(x, ..., include_docvars = TRUE) {
 #' @param x object to be coerced or checked
 #' @param concatenator character between multi-word expressions, default is the underscore character.  See Details.
 #' @return \code{as.tokens} returns a quanteda \link{tokens} object.
-#' @details The \code{concatenator} value is (TO BE COMPLETED).
+#' @details The \code{concatenator} is used to automatically generate dictionary values for multi-word expressions
```
kbenoit: Please provide a bit more detail, and an (prose) example.

R/tokens.R
```diff
@@ -208,13 +268,14 @@ as.tokens <- function(x, concatenator = '_') {
 #' @export
 as.tokens.list <- function(x, concatenator = '_') {
     result <- structure(tokens_hash(x),
-                        "class" = c("tokens", "tokenizedTexts"),
-                        "what" = "word",
-                        "ngrams" = 1L,
-                        "skip" = 0L, 
-                        "concatenator" = concatenator,
-                        "padding" = FALSE)
-    docvars(result) <- data.frame(row.names = docnames(result))
```
kbenoit: Wouldn't all this be easier if we had an S4 class for this with a prototype to set these slots?

R/tokens.R
```diff
@@ -472,9 +549,12 @@ tokens_internal <- function(x, what = c("word", "sentence", "character", "fastes
     for (i in seq_along(x)) {
         
         if (verbose) catm("...tokenizing", i, "of" , length(x), "blocks\n")
-        if (what %in% c("word", "fastestword", "fasterword")) {
```
kbenoit: This look might be a natural thing to parallelise (in R).

R/corpus-methods-quanteda.R
koheiw: Documents should have names always, but really old corpus object might not. This is a safeguard for errors in subsequent processes. If we were to give up backward compatibility, better to raise errors here.

tests/testthat/test-dfm.R
koheiw: These are the oldest test for dfm that were not isolated. I do not see whey they are there either. I assigned output values to variables only because I do not want to set dfm to be printed to my console.

tests/testthat/test-dfm.R
koheiw: I think we should delete all the code in `test_that("oldest dfm test", {})`.

tests/testthat/test-docnames.R
koheiw: We need to have some kind of integrity check for all the `corpus` and `tokens`. S4 could be one, but could be done in S3. The main point of corpus overhaul.

tests/testthat/test-tokens.R
koheiw: '-' is remove in the second case, because `stringi::stri_detect_regex('-', "^[\\p{P}]$") == TRUE`. It seems counter intuitive to make the effect of `remove_hyphens` conditional on `remove_punct`. 

It should be like
```r
tokens("One two-three four.", remove_hyphens = TRUE)
tokens from 1 document.
text1 :
[1] "One"   "two"   "-"     "three" "four"  "."    

> tokens("One two-three four.", remove_hyphens = FALSE)
tokens from 1 document.
text1 :
[1] "One"       "two-three" "four"   "."    
```

R/tokens.R
koheiw: I don't think there will be a much performance gain, but `tokens_remove(toks, 'P', 'charclass')` is certainly much easier than `tokens_remove(toks, "^[\\p{P}]$", 'regex')`.

R/tokens.R
koheiw: Done

koheiw:
Since I separated special handling of hyphens and Twitter tags etc, `tokens_word()` became almost nothing. We can use **tokenizers** anytime.

R/tokens.R
kbenoit: I was thinking we would stick to the **stringi** syntax, not create an interpretative layer in between. So it would be similar to
```r
> stri_detect_charclass("Unicode symbol £", "[\\p{S}\\p{P}]")
[1] TRUE
> stri_detect_charclass("Unicode puncutation !", "[\\p{S}\\p{P}]")
[1] TRUE
```

tests/testthat/test-tokens.R
kbenoit: I agree, this is how I would expect it to work too.

tests/testthat/test-dfm.R
kbenoit: I'll take a look at that.

R/corpus-methods-quanteda.R
kbenoit: OK

kbenoit:
Some dfm print tests are failing because the tokenisation behaviour has changed. Before we just rewrite the test expectations, should we investigate the differences so we are sure we understand them?

koheiw:
I added `old` argument to tokens to compare with the verson:
```r
old <- unlist(tokens(data_char_ukimmig2010, old = TRUE), use.names = FALSE)
new <- unlist(tokens(data_char_ukimmig2010, old = FALSE), use.names = FALSE)
new2 <- unlist(tokens(data_char_ukimmig2010, remove_hyphen = TRUE, old = FALSE), use.names = FALSE)

setdiff(old, new)
setdiff(new, old)

setdiff(old, new2)
setdiff(new2, old)

> setdiff(old, new)
[1] "abiding"        "non"            "ill"            "educated"       "knee"           "jerk"           "300,000"       
[8] "represented"    "term"           "re"             "building"       "soft"           "touch"          "breakers"      
[15] "Abiding"        "sponsored"      "ethnicity"      "28"             "decision"       "creating"       "cross"         
[22] "boundary"       "E"              "longer"         "discriminatory" "quality"        "No"             "traffickers"   
[29] "co"             "ordinated"      "wide"           "point"          "scoring"        "recently"       "skill"         
> setdiff(new, old)
[1] "law-abiding"           "decades-long"          "non-British"           "ill-educated"          "knee-jerk"            
[6] "300,000-500,000"       "over-represented"      "Non-Immigration"       "non-immigration"       "long-term"            
[11] "re-building"           "second-generation"     "soft-touch"            "law-breakers"          "Law-Abiding"          
[16] "re-entry"              "government-sponsored"  "ethnicity-specific"    "28-style"              "second-class"         
[21] "decision-making"       "nation-building"       "civilisation-creating" "non-EU"                "cross-boundary"       
[26] "E-borders"             "longer-term"           "non-discriminatory"    "high-quality"          "Australian-style"     
[31] "points-based"          "police-level"          "public-sector"         "No-one"                "people-traffickers"   
[36] "high-priority"         "co-ordinated"          "EU-wide"               "point-scoring"         "recently-arrived"     
[41] "high-skill"            "five-year"             "time-limited"          "non-work"              "face-to-face"         
> setdiff(old, new2)
character(0)
> setdiff(new2, old)
character(0)
```

kbenoit:
OK, nice. So it looks like most of the difference is coming from hyphens being preserved (correctly) in the new code, while being split (incorrectly) in the old code.

koheiw:
It seems to be the case in Japanese texts too.

koheiw:
This is just a confirmation:
```r

    old = tokens(corp, old = TRUE),
    new = tokens(corp, remove_hyphen = TRUE, old = FALSE),
    times = 10
)


Unit: seconds
 expr       min       lq     mean   median       uq      max neval
  old 14.301439 14.73900 15.08327 15.02283 15.65775 15.80192    10
  new  9.725151 10.22395 10.31369 10.41880 10.43882 10.86264    10
```

Issue #496, opened 2017-01-22
dfm_lookup needs "non-matched" column option


kbenoit:
For computing relative frequencies in a dfm after applying a dictionary, what is usually needed is the dictionary count relative to all features. But `dfm_lookup` only keeps features found. To get relative frequencies of all tokens requires some workarounds, but it would be better to have directly. (An earlier version of `dfm()` did behave this way.)

```r
require(magrittr)

txt <- c(d1 = "a b c a a b d",
         d2 = "a a b d", 
         d3 = "a b d D d D d")
toks <- tokens(txt)

mydict <- dictionary(list(A = "a", BC = c("b", "c")))

# how to get frequencies of non-dictionary items?
dfm(toks, tolower = FALSE, dictionary = mydict) %>% 
    dfm_weight("relFreq")
## Document-feature matrix of: 3 documents, 2 features (0% sparse).
## 3 x 2 sparse Matrix of class "dfmSparse"
##     features
## docs         A        BC
##   d1 0.5000000 0.5000000
##   d2 0.6666667 0.3333333
##   d3 0.5000000 0.5000000

# not idea since cannot distinguish keys from original features
dfm(toks, tolower = FALSE) %>% 
    dfm_lookup(dictionary = mydict, exclusive = FALSE) %>% 
    dfm_weight("relFreq")
## applying a dictionary consisting of 2 keys
## Document-feature matrix of: 3 documents, 4 features (16.7% sparse).
## 3 x 4 sparse Matrix of class "dfmSparse"
##            d         D         A        BC
## d1 0.1428571 0         0.4285714 0.4285714
## d2 0.2500000 0         0.5000000 0.2500000
## d3 0.4285714 0.2857143 0.1428571 0.1428571



# workaround
tokens_select(toks, unlist(mydict), padding = TRUE) %>%
    tokens_lookup(dictionary = mydict, exclusive = FALSE) %>%
    dfm(tolower = FALSE) %>%
    dfm_weight("relFreq")
## Document-feature matrix of: 3 documents, 3 features (0% sparse).
## 3 x 3 sparse Matrix of class "dfmSparse"
##     features
## docs                   A        BC
##   d1 0.1428571 0.4285714 0.4285714
##   d2 0.2500000 0.5000000 0.2500000
##   d3 0.7142857 0.1428571 0.1428571
```

koheiw:
I think this is related to  #479.

kbenoit:
> I think this is related to #479.

[...]

PR #889, opened 2017-07-28
Dev ntoken

koheiw:
`original` argument is added to `ntoken()` for #496.


kbenoit:
I see the idea here, but I think the only time this will be needed is when performing `dfm_lookup()` (or using a dictionary in `dfm()`). So what if we just add an attribute to a `tokens` object when someone has applied `tokens_lookup()`, and just compute this on the fly (if needed) in `dfm_lookup()`?

I'm not keen on complicating the `ntoken()` (or any `n*` function) with additional options.

No matter what way we adopt, we will need to add an option to both `dfm()` and `dfm_lookup()` called (something like `show_unmatched = TRUE)` to provide a column for non-matched features. It's still not perfectly compositional, by the way, since some features might be counted in multiple keys.

koheiw:
It can be done in different ways but the goal should be normalizing dfm, not just showing unmatched counts, as in the examples for `dfm_lookup()`:
```r
myDict <- dictionary(list(christmas = c("Christmas", "Santa", "holiday"),
                          opposition = c("Opposition", "reject", "notincorpus"),
                          taxglob = "tax*",
                          taxregex = "tax.+$",
                          country = c("United_States", "Sweden")))
myDfm <- dfm(c("My Christmas was ruined by your opposition tax plan.", 
               "Does the United_States or Sweden have more progressive taxation?"),
             remove = stopwords("english"), verbose = FALSE)
myDfmDict <- dfm_lookup(myDfm, myDict, valuetype = "glob")

# normalize dictionary count
myDfmDict / ntoken(myDfmDict, original = TRUE)

# Document-feature matrix of: 2 documents, 5 features (50% sparse).
# 2 x 5 sparse Matrix of class "dfmSparse"
# features
# docs    christmas opposition   taxglob taxregex country
# text1 0.1666667  0.1666667 0.1666667        0     0  
# text2 0          0         0.2000000        0     0.4
```
It could be:
```r
dfm_weight(myDfmDict, 'relFreqAll')`

myDfmDict / rowSums(myDfmDict[,,all = TRUE]) # looks crazy though
```

kbenoit:
If a user wants that, I suggest simply:
```r
myDfmDict / ntoken(myDfm)
# Document-feature matrix of: 2 documents, 5 features (50% sparse).
# 2 x 5 sparse Matrix of class "dfmSparse"
#        features
# docs    christmas opposition   taxglob taxregex   country
#   text1 0.1428571  0.1428571 0.1428571        0 0        
#   text2 0          0         0.1666667        0 0.3333333
```
This will always work because the token


Dividing by the original number of tokens doesn't solve the compositionally issue for multiple dictionary matches, since:

```r
myDict <- dictionary(list(one = letters[1:3], 
                          two = letters[2:4],
                          three = letters[3],
                          OTHER = "e"))
mydfm <- dfm_lookup(dfm(c("a b c d", "b c d e")), myDict)

mydfm / ntoken(mydfm)
# Document-feature matrix of: 2 documents, 3 features (0% sparse).
# 2 x 3 sparse Matrix of class "dfmSparse"
#        features
# docs          one       two     three
#   text1 0.4285714 0.4285714 0.1428571
#   text2 0.3333333 0.5000000 0.1666667

mydfm / ntoken(mydfm, original = TRUE)
# Document-feature matrix of: 2 documents, 3 features (0% sparse).
# 2 x 3 sparse Matrix of class "dfmSparse"
#        features
# docs     one  two three
#   text1 0.75 0.75  0.50
#   text2 0.50 0.75  0.75
```

I am proposing a method that will always produce a properly row-"normalised" matrix (even if that's slightly wrong, it's wrong in a better way):

```r
# (here I added a category to simulate functionality not yet added)
myDict2 <- dictionary(list(one = letters[1:3], 
                          two = letters[2:4],
                          three = letters[3],
                          "_unmatched_" = "e"))
mydfm2 <- dfm_lookup(dfm(c("a b c d", "b c d e")), myDict2) #, show_unmatched = TRUE)
mydfm2
# Document-feature matrix of: 2 documents, 4 features (12.5% sparse).
# 2 x 4 sparse Matrix of class "dfmSparse"
#        features
# docs    one two three _unmatched_
#   text1   3   3     1           0
#   text2   2   3     1           1

tf(mydfm2, "prop")
# Document-feature matrix of: 2 documents, 4 features (12.5% sparse).
# 2 x 4 sparse Matrix of class "dfmSparse"
#        features
# docs          one       two     three _unmatched_
#   text1 0.4285714 0.4285714 0.1428571   0        
#   text2 0.2857143 0.4285714 0.1428571   0.1428571
```

### Advantages

1. We don't make `ntoken` more complicated, with an argument that only applies to certain object classes.  
2. The dfm from a lookup operation is the same, plus one more column consisting of unmatched category of token counts.
3. We can use existing dfm weighting tools, and they will behave the way they should.
4. A user will _still_ have the option to do it the way you show above, simply by using

koheiw:
Users can just do this to get unmatches
```r
myDict2 <- dictionary(list(one = letters[1:3], 
                           two = letters[2:4],
                           three = letters[3]))
mydfm <- dfm(c("a b c d", "b c d e"))
mydfm2 <- dfm_lookup(mydfm), myDict2)
mydfm2[,'unmatch'] <- rowSums(dfm_remove(mydfm, myDict2))
```

kbenoit:
Not possible to assign a new column of the dfm that way, but this works. 

(Looks like a bit of code golf, because we don't have operators defined for `+` of different sparse matrix objects with a dfm, but it works.  The + 1 is to make sure we get zero-length documents too.)

```r
myDict2 <- dictionary(list(one = letters[1:3], 
                           two = letters[2:4],
                           three = letters[3]))
mydfm <- dfm(c("a b c d", "b c d e"))

nunmatched <- ntoken(dfm_remove(mydfm, myDict2))
cbind(dfm_lookup(mydfm, myDict2),
      new("dfmSparse", as(dfm(as.tokens(lapply(nunmatched+1, function(x) rep("_unmatched_", x)))) - 1, "dgCMatrix")))
# Document-feature matrix of: 2 documents, 4 features (12.5% sparse).
# 2 x 4 sparse Matrix of class "dfmSparse"
#        features
# docs    one two three _unmatched_
#   text1   3   3     1           0
#   text2   2   3     1           1
```

koheiw:
It looks like time to add an assignment accessor to dfm for advanced users.

kbenoit:
Assignment already works fine:
```r
> mydfm[, 5] <- c(4,5)
```
but matrix indexing generally in R does not allow indexing non-existent dimensions: (unlike columns of a data.frame or a more general list):
```r
> mydfm[, 6] <- c(4,5)
Error in intI(i, n = di[margin], dn = dn[[margin]], give.dn = FALSE) : 
  index larger than maximal 5
> as.matrix(mydfm)[, 6] <- c(4,5)
Error in `[<-`(`*tmp*`, , 6, value = c(4, 5)) : subscript out of bounds
```

koheiw:
If rowSums() returns a dfm, we can use cbind() to add a column, like rowSums(x, colname = "nomatch").

kbenoit:
`rowSums()` returns a named integer vector, as it should. This is the same as `ntoken.dfm()`. We just need to add a signature for `cbind` for `("dfm", "numeric")` to allow adding the integer vector to the dfm as a new column.

Since it's not S4 at the moment (it's S3 in `dfm-classes.R` it will be easy, but we need to check the type and then coerce it into a dfm, then call the cbind again. (or reimplement as S4 methods!)

kbenoit:
@koheiw I undid the changes to `ntoken()`, and added `show_unmatched` to `dev_lookup()`. Still needs:

- [ ] add `show_unmatched` argument to `dfm()`
- [ ] the ability to show unmatched if the dfm is built from tokens that have applied `tokens_lookup()`. Here I kept the new metadoc variable you added to make this possible.
- [ ] consider a better label for the unmatched tokens than `_unmatched_` - could be `UNMATCHED` or anything better you might think of
- [ ] add tests

Ultimately we will also improve `cbind.dfm()` but for now I went with the solution I used above.

koheiw:
We should go to the `cbind()` solution straight away. I do not think it is a good idea add extra arguments to only for no matches. I notice that it is as easy as this:

```
myDict2 <- dictionary(list(one = letters[1:3], 
                           two = letters[2:4],
                           three = letters[3]))
txts <- c("a b c d", "b c d e") 

mydfm <- dfm(txts)
mydfm2 <- dfm_lookup(mydfm), myDict2)
cbind(mydfm2, as.dfm(cbind("NOMATCH" = ntoken(dfm_remove(mydfm, myDict2)))))

mytok <- tokens(txts)
mytok2 <- tokens_lookup(tokens(txts), myDict2)
cbind(dfm(mytok), as.dfm(cbind("NOMATCH" = ntoken(tokens_remove(mytok, myDict2)))))
```
Extending `dfm.cbind()` can be deferred until overhauling **dfmSparse**. In that case, add `as.dfm.vector()`, which only does `as.dfm(cbind(x))`, now.

I can withdraw this PR, separating important changes related to other functions to other branches.

koheiw:
Alternatively we could add `nomatch = TRUE` (like `match()` function) to `dfm_lookup()` and `tokens_lookup()`. On dfm, just rename features and run `dfm_compress()`. On tokens, record unmached like this:
```r
list(text1= c('KEY1', 'NOMATCH', 'KEY2', 'NOMATCH'))
```
`nomatch = TRUE` can only be used `exclusive = TRUE`.

kbenoit:
Let's discuss Monday.

kbenoit:
@koheiw I think this is ready to go now, but have a look. I did the following:
- implemented `show_unmatched` in `dfm_lookup()` only. Didn't want to change any other arguments to any other functions. If someone wants this now, they can either use the new and improved `cbind.dfm()` "manually", or use `dfm_lookup()` in a second stage after creating the dfm.
- cleaned up `dfm.R` generally - once I was in there I decided to straighten out the code. It's non-repetitive now and very consistent.

PR #892, opened 2017-07-30
Improve cbind functions for dfm and conformable objects

kbenoit:
Adds methods for:
- cbind: dfm + numeric, numeric + dfm, dfm + matrix, matrix + dfm

Adds `quanteda_options`: `docnames_stem` and `featnames_stem` to centralise the sub names for documents and features that are unassigned. (I noticed that what is done in `corpus()` and `dfm()` was different from `as.dfm()`.)

Still needed: 
- [x] make the new methods applicable to all cbind objects passed in the `...`
- [x] add tests

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -41,12 +41,12 @@ qatd_cpp_fcm <- function(texts_, n_types, count, window, weights, ordered, tri, 
     .Call(`_quanteda_qatd_cpp_fcm`, texts_, n_types, count, window, weights, ordered, tri, nvec)
 }
 
-qatd_cpp_sequences <- function(texts_, types_, count_min, sizes_, method, smoothing) {
-    .Call(`_quanteda_qatd_cpp_sequences`, texts_, types_, count_min, sizes_, method, smoothing)
-}
-
 qatd_cpp_sequences_old <- function(texts_, words_, types_, count_min, len_max, nested, ordered = FALSE) {
     .Call(`_quanteda_qatd_cpp_sequences_old`, texts_, words_, types_, count_min, len_max, nested, ordered)
+}
+
+qatd_cpp_sequences <- function(texts_, types_, count_min, sizes_, method, smoothing) {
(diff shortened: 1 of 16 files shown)


koheiw:
To achieve the level of flexibility R's base functions have, it should be like this... not cool, but works.
```r

cbind <- function(...){
    args <- list(...)
    names <- names(args)
    
    x <- args[[1]]
    if (is.dfm(x)) {
        cat('x is a dfm\n')
    } else if (is.matrix(x)) {
        cat('x is a vector\n')
        x <- as.dfm(x)
    } else if (is.numeric(x)) {
        cat('x is a vector\n')
        x <- as.dfm(matrix(x, ncol = 1, nrow = length(x), dimnames = list(NULL, names[1])))
    }
    
    y <- args[[2]]
    if (is.dfm(y)) {
        cat('y is a dfm\n')
    } else if (is.matrix(y)) {
        cat('y is a vector\n')
        y <- as.dfm(y)
    } else if (is.numeric(y)) {
        cat('x is a vector\n')
        y <- as.dfm(matrix(y, ncol = 1, nrow = length(y), dimnames = list(NULL, names[2])))
    }
    
    quanteda:::cbind_dfm_dfm(x, y)  
}


cbind(mydfm, "ALL" = ntoken(mydfm))
cbind(mydfm, ntoken(mydfm))

kbenoit:
I'm happy to rename the new options to `base_docname` and `base_featname` (or transposed if you think that's better).

I think the existing code works with names for vectors already. Also the loop in bind.dfm over `...` will allow joining any length sequence of compatible objects. It uses S4 method signatures instead of the conditions in the code above. And, it assigns a (uniwue) feature name when none is present, since no dfm should have unnamed features. However, a better way to do this might be to form the dfm, and then send it to a `dfm_check()` internal routine that sequentially names any unnamed features.

kbenoit:
Updated now, all of the above should work, with functionality identical to `base::cbind()` except that it will add featnames if none exist. I also changed the option names to `base_featname` and `base_docname`.

PR #897, opened 2017-08-01
Add textstat_frequency() function, tests

kbenoit:
- Infrastructure for #825 

From this, @stefan-mueller you should be able to build a (single) `textplot_frequency()` with a `n` argument, some group options, or a target frequency. We should think about whether we want this to be something in the function, or whether the user should use `subset.data.frame()`, etc to target individual features. Happy to discuss.


stefan-mueller:
Thanks Ken! Some thoughts: You might add an argument `n` (how many top features should be returned based on `rank`) and `decreasing = TRUE/FALSE` to return the most/least frequent features. 

At the moment, `textstat_frequency()` returns all words, but it's helpful have an option for the number of features, and whether to show the most or least frequent ones (of course this can be done by the user by subsetting the data frame, but the arguments above could be time-savers). Also, it would be great to do this by group: `textstat_frequency(my_dfm, n = 10, groups = "GROUP")` should return the 10 most frequent words for _each_ group, while  `textstat_frequency(my_dfm, n = 10)` returns the 10 most frequent words of the entire dfm. After these changes, I think we can push `textstat_frequency()`, and I will issue a separate PR for `textplot_frequency()`.

stefan-mueller:
Just wondering whether it makes sense to add the weight options from `dfm_weight()` to `textstat_frequency()`. It seems that `textstat_frequency()` only returns the absolute frequencies. Of course, it's not possible to first weight a `dfm` and apply `textstat_frequency` afterwards:

```r
ie_dfm <- dfm(data_corpus_irishbudget2010, 
               remove = stopwords("english"),
               remove_punct = TRUE)

ie_dfm_weight <- dfm_weight(ie_dfm, type = "relfreq")

frequency_weight <- textstat_frequency(ie_dfm_weight)

# Error in tf.dfm(x, "boolean") : 
#   this dfm has already been term weighted as:propNULLNULL
```

Is it reasonable to add a `type` option as in `dfm_weight()` to `textstat_frequency()`?

kbenoit:
I prefer to weight beforehand. I can fix that warning.

kbenoit:
Fixed in f245426.

stefan-mueller:
I was wondering whether an extra column in the `textstat_frequency` data frame is necessary/possible when `textstat_frequency` is applied to a weighted dfm? To label `textplot_frequency` plots properly (the correct frequency type) we need a way to store information about the frequency type applied if the values aren't whole numbers. Otherwise, the user would need to do this manually with `xlab()`.

To it it automatically, we probably need an extra slot in the `dfm_weight` object with information about the weighting type applied. And this information then could be transferred to an additional column in `textstat_frequency`. What do you think about this, @kbenoit?

stefan-mueller:
Or what about changing the `frequency` column in the `textstat_frequency`data frame if `textstat_frequency`is based on `dfm_weight` object? So the column could currently have one of these values: "frequency", "relfreq", "relmaxfreq", "logfreq",  "tfidf". With the new column name, it would be easier to see *which* frequency was actually calculated, and it makes it very easy to label `textplot_frequency` plots correctly.

Issue #536, opened 2017-02-01
Redesign and extend the tokens object 


kbenoit:
**For discussion**

A new, augmented tokens object should have the following improvements:

*  Be an S4 object that contains a list, rather than an S3 object, because list operations tend to obliterate the attributes and hence require special handling.  S4 objects, on the other hand, tend to preserve the slots even when executing methods for object types contained in the S4 class.

    This could involve a hierarchy of several classes:

    1.  a `tokens` super class, which contains a list, and slots corresponding to the current `tokens` attributes.

    2.  a `tokens-tagged` class that contains tokens, that also contains information about the parts of speech of each type (see below).  

    3.  a `tokens-compounded` class that contains `tokens`, to record compounded tokens (multi-word expressions), including the scoring and methods used to detect them.  

    Existing methods would be extended as required for the special subclasses.

*  `tokens` with parts of speech (the proposed `tokens-tagged` class) could record this information in one of the following ways:

    - with an additional slot parallel to `types`, which records a part of speech for each type.  This would distinguish homograph types otherwise treated as identical if they were different parts of speech, e.g. `program/NOUN` and `program/VERB`, for the term "program".  We could have these part of speech types mapped to an integer lookup table of their own, for known tagging schemes (Penn, Google).

    - in a list of length *V* as present (where *V* represents the length of the types vector), but with each element in this list a two-element vector consisting of an index to the type, and an index to the POS

    - we could combine POS information with the type to make a unique combined type*pos, so that the `types` table contain separate single entries for `program/NOUN` and `program/VERB`.  This is probably the simplest since it could be done within the existing object framework without modification (except for what is input), but it would make downstream selection of tokens based on POS less efficient.  

    - as a data table, consisting of documentID, type, and POS.  (Any or all of these could be indexes or just values.)

*  A similar scheme could be used to record parallel information on scoring for MWEs, although this might point to just the token index locations that are MWEs, rather than to unigrams as well.  

*  We could consider having an additional `types-stemmed` index corresponding to the stemmed types, so that we could preserve the original types but also include the stemmed variants.  This would make stemming an indexing operation rather than one of transforming words.  (I don't think we need this for lower and upper case, since these operations are already extremely fast and therefore can be executed on demand.)

*  New methods required:

    For constructing a special tokens object:  
    - `tokens_tagpos` - inputs a simple tokens object and outputs a `tokens-tagged` class object.  (#435)
[...]

Issue #63, opened 2015-11-20
Implement settings for corpus and downstream functions


kbenoit:
A tokenized text object, and a dfm object, should have settings that record the processing options applied to the texts or corpus from which they were created. These provide a record of what was done to the text, and where it came from.  Examples are `tolower`, `stem`, `removeTwitter`, etc.  They also include any objects used in feature selection, such as dictionaries or stopword lists.

kbenoit:
Closing because we've moved on, but partly this is addressed by #1845.

PR #898, opened 2017-08-01
Issue 882

koheiw:


koheiw:
We can use ngram and skip arguments in `dfm.tokens` now. I think `tokens.tokens` should have the same signature as `dfm.tokens` (It is more like `tokens.characters` currently).

kbenoit:
I'm  not sure what you mean by `tokens.tokens` having the `dfm.tokens` signature.

koheiw:
I don't know why I said "tokens.tokens should have the same signature as dfm.tokens", but it is out my concern regarding different outputs from `tokens.tokens` and `tokens.chars` as shown below. 
```r
txt = "abc be-fg hi 100kg 2017"
tokens(tokens(txt, remove_hyphens = FALSE), remove_hyphens = TRUE)
# [1] "abc"   "be fg" "hi"    "100kg" "2017"

tokens(txt, remove_hyphens = TRUE)
# [1] "abc"   "be"    "-"     "fg"    "hi"    "100kg" "2017"


tokens(tokens(txt, remove_hyphens = FALSE), remove_hyphens = TRUE, ngrams = 2)
# [1] "abc_be"     "be_-"       "-_fg"       "fg_hi"      "hi_100kg"   "100kg_2017"

tokens(txt, remove_hyphens = TRUE, ngrams = 2)
# [1] "abc"   "be"    "-"     "fg"    "hi"    "100kg" "2017"

```
Tokenization is irreversible transformation from texts to tokens, but `tokens.tokens` gives an impression that we can redo tokenenization after the first tokenization. I think it is better to keep it only internal and use only in `dfm.tokens` to trap extra arguments (`ngrams` and `dictionary`).

kbenoit:
I see the point - you cannot restore something by setting `tokens(x, remove_something = FALSE)` if the `x` was already created by `tokens(x, remove_something = TRUE)`. 

When we implement the settings functionality (described in #536 and #63) we will be able to check this and issue warnings. I don't think many people will be using `tokens.tokens()` anyway, but let's leave it exposed and I will add a caution note to the documentation.

PR #899, opened 2017-08-02
Update correspondence analysis in textplot_scale1d

stefan-mueller:
- Add example for correspondence analysis.
- Make `geom_pointrange()` conditional on model to be plotted. (So far), `textstat_ca()` does not return confidence intervals. Therefore, `geom_pointrange()` returned a warning when used with a fitted ca-textmodel.


kbenoit:
I changed this slightly, so that if in the internal function `textplot_scale1d_documents(x, se, ...)`, `se` is `NA` (as it is for a `textmodel_ca_fitted`) then it just makes the SEs zero - so they are not plotted - but also there is no **ggplot2** warning. This is more general than looking for the `textmodel_ca_fitted` class.

PR #900, opened 2017-08-03
Add textplot_frequency.R

stefan-mueller:
Following #897, this is the first version of `textplot_frequency()` which works on the document/group level as well as for features. I based the `margin` function on the one from `textplot_scale1d()`.
@kbenoit, please have a look at the code and the examples. 

So far `textplot_frequency()` does not work for only one feature and several groups. All other examples from the man should work.

```r
ie_dfm <- dfm(data_corpus_irishbudget2010[1:6], 
               remove = stopwords("english"),
               remove_punct = TRUE)

ie_groups <- apply(docvars(data_corpus_irishbudget2010, c("name", "party")), 
                 1, paste, collapse = " ")[1:6]

# does not work
textplot_frequency(ie_dfm, features = "tax", groups = ie_groups, margin = "features")

# works (but facet_wrap(~ group) needs to be added
textplot_frequency(ie_dfm, features = c("tax", "Ireland"), groups = ie_groups, margin = "features")
```
Also tests need to be added and the manual improved.

### DIFF
--- a/R/textplot_frequency.R
+++ b/R/textplot_frequency.R
@@ -1,0 +1,151 @@
+#' plot the frequency of a feature across documents
+#' 
+#' Plots the absolute or relative frequency of a feature in all documents of a 
+#' document-feature matrix. Given that this returns a ggplot object, you can 
+#' modify the plot by adding ggplot layers (see example).
+#' @param x a dfm or weighted dfm object
+#' @param margin "documents" to plot the most frequent features across documents or groups
+#'   or "features" to plot the occurence of one or more features
+#' @param features the feature or features whose frequency will be plotted
+#' @param group a vector of names to group the dfm; if left \code{NULL} (the
+#'   default) the feature frequencies will be estimated across all documents
+#' @param n how many top features should be returned (only relevant for margin = "documents")
+#'
+#' @return a \pkg{ggplot2} object
+#' @author Stefan Müller and Kenneth Benoit
+#' @seealso \code{\link{textstat_frequency}}, \code{\link{dfm}}, \code{\link{dfm_weight}}
+#' @examples 
+#' \dontrun{
+#' # transform to dfm and get grouping doclabels
+#' ie_dfm <- dfm(data_corpus_irishbudget2010[1:6], 
+#'               remove = stopwords("english"),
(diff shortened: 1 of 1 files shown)


kbenoit:
I'd be aiming for something more like http://tidytextmining.com/tfidf.html. There is no need for a groups argument, since that's taken care of in `textstat_frequency()`, and because it's already in long format, it will be very easy to use the **ggplot2** functions as in the link above.

I also think we should let the user do any selection on the input object, rather than providing wrappers for this. It's a standard data.frame so they can use `subset()`, or `dplyr::filter()`, or whatever.

stefan-mueller:
Ok, I see. So far, I replicated and extended the frequency plots from the **quanteda** plotting [vignette](http://quanteda.io/articles/pkgdown_only/plotting.html#frequency-plots). 

Just some questions for clarification: Do you want the `x`-input to be a `dfm` or a `textstat_frequency()` data frame? If we use the second option, I agree that we don't need the `group` argument. But if we use a `dfm` as the input, don't we need an argument to specify the group?

And which of the plot types from http://tidytextmining.com/tfidf.html would you like to incorporate into `textplot_frequency()`?

kbenoit:
Input should be a `textplot_frequency` output - that was the whole point of adding that function.

As for which plots, I was thinking 3.4 onward, with the groups (or documents) determining the different plots. So, we might need a `groups` argument.

stefan-mueller:
Perfect, that makes sense. I'll redesign `textplot_frequency` based solely on `textstat_frequency` and get back to you.

stefan-mueller:
Sorry that I didn't find time to implement this. Yet, as `textstat_frequency` returns a data frame in long format, it should be easy for users to plot frequencies (and facet by group) with **ggplot2**.  The frequency plots from the [Text Mining with R](http://tidytextmining.com/tfidf.html#the-bind_tf_idf-function) book (sections 3.3-3.4) can be reproduced easily with our `textstat_frequency` data frame.

kbenoit:
That's no problem, and we can always reopen the PR once it's done. 

Feel free in the meantime to add a plot example using the long format and **ggplot2** to either the plotting vignette or to `textstat_frequency.Rd` (or both).

PR #901, opened 2017-08-03
Re-implement textstat_collocations()

kbenoit:
Addresses, and will eventually solve, #803.

This is a pretty substantial rewrite of `textstat_collocations()`, that does the following:

- moved all of collocation detection code from `sequences.R` into `textstat_collocations.R`
- removed `sequences_old.R`
- removed tests based on old sequences
- rewrote `textstat_collocations()` significantly
    * removed dice, LFMD, and gensim until we get these working
    * made it possible to select a single measure, or  all by default
    * added option show_counts to output observed and expected counts
    * added new computation of G2, chi2, and pmi based on new observed and expected counts
    * added methods for character and corpus objects, which are tokenized in a way that segments on punctuation, so that no collocation is formed as a collocation that includes punctuation, but punctuation tokens are not counted in the marginal totals
- updated tests that relied on `sequences()` or older implementations of `textstat_collocations()`


kbenoit:
Possibilities for getting expected counts in C/C++:
* (C) `apop_data()` from the [Apophenia open statistical library](http://apophenia.info/group__all__public.html#ga830250966bff3fb314141512ee88c688)
* (R) `mipfp::Ipfp()` from https://cran.r-project.org/web/packages/mipfp/mipfp.pdf

PR #904, opened 2017-08-05
Dev nomatch

koheiw:


koheiw:
We might need to define a system-reserved type range beyond zero, but let's not do that now. I found an easy way to append a no-match key to types in tokens, whereby it will appear in the left most column in dfm.

PR #917, opened 2017-08-08
Add remove_delimiter argument

koheiw:


kbenoit:
How about instead of adding an argument to the segment functions, you simply implement a "tag" option to `tokens_segment()`? Adding `remove_delimiter` makes an existing problem worse, that the behaviour of one argument influences the behaviour of a second argument - here, `what` and `remove_delimiter` - and that some combinations are incompatible, such as `what = "tag"` and `remove_delimiter = FALSE`.

Principle of minimal intervention/programmer's Hypocratic oath: can't we simply add a "tags" option to `tokens_segment()` to solve the issue in https://github.com/kbenoit/quanteda/issues/902#issuecomment-320932109. Let's stick to what we decided yesterday and improve the `*_reshape`/`*_segment` functions later, and keep their existing behaviours and signatures as they are for v0.9.9000.

koheiw:
`remove_argument` is something will will need in tokens_segment, so just keep it internal for now. When I was scrutinizing the code, I got a sense that we really need to separate tag extraction function.

kbenoit:
That's fine, keep it in tokens_segment() internal, but restore the old versions of `char_segment()` and `corpus_segment()`.

PR #921, opened 2017-08-09
Typo correction in textstat_dist() argument

t0tem:
I got an error 
`"Error in match.arg(type) : 'arg' should be one of “frequency”, “relFreq”, “relMaxFreq”, “logFreq”, “tfidf”`
while going through vignette. So I think `relfreq` should be `relFreq`


kbenoit:
Thanks but this is already `"relfreq"` in the `master` branch. The package has undergone a lot of updates recently, with a CRAN update coming tomorrow. Suggest you always pull from the upstream repo before submitting a PR.

But thanks anyway!!

Issue #911, opened 2017-08-08
Create as.dfm for dgCMatrix


koheiw:
Just do `new("dfmSparse", x)`, but users may not find this.

kbenoit:
Solved in #930.

Issue #55, opened 2015-10-07
preserve settings for dfm objects


kbenoit:
From Issue #46 @zachmayer:

"I need to remember the settings I passed to the training set dfm. This isn't too bad, but it'd sure be convenient to include them inside dfmSparse in case I pass the model on to someone else."

kbenoit:
Same as #63.

PR #930, opened 2017-08-21
Dev dfm

koheiw:
Addresses #494 and #911


koheiw:
Strangely, textmodels are failing on AppVeyor only. There is no such errors on my Windows 10.

kbenoit:
We need to be very careful once we start changing the object classes, since here, we could break functionality with the objects that users have created previously. I'm not sure there is an easy solution to this, except perhaps to add to each `dfm_*` function a call to a function that checks for older objects, and if they are older, runs `dfm(x)` on the object to construct the newer version from the older. We should also add a slot to hold the value of `packageVersion("quanteda")` for when the object was created.

`as.dfm()` should still work, but `dfm.dfmSparse()` and `dfm.dfmDense()` can be defined and called after checking for older classes. This would mean you could uncomment the code you added to (e.g.) `tests/testthat/test-textmodel_wordscores.R` that first calls `as.dfm()` on the input object, since it would still work. But the check-and-convert-if-necessary could also issue a deprecation warning.

kbenoit:
And we can delete `tests/testthat/Rplots.pdf` - I'm not sure which test creates this but we should stop if from doing so!

koheiw:
I wrapped all the dfm inputs by `as.dfm()` in exported functions for backward compatibility. I think it is a good idea to always wrap complex input objects (dfm, tokens, corpus, dictionary etc.) that would change their internal structure.

kbenoit:
@kohei this looks good but I still see:
```r
dfm(c(d1 = "a a b c", d2 = "b b c c c d"))
# Document-feature matrix of: 2 documents, 4 features (25% sparse).
# 2 x 4 sparse Matrix of class "dfmSparse"
#     features
# docs a b c d
#   d1 2 1 1 0
#   d2 0 2 3 1
```

Shouldn't that print as `dfm` instead of `dfmSparse`? I could not figure out why it's still classed as `dfmSparse`.

koheiw:
I am getting this:
```r
> dfm(c(d1 = "a a b c", d2 = "b b c c c d"))
Document-feature matrix of: 2 documents, 4 features (25% sparse).
2 x 4 sparse Matrix of class "dfm"
    features
docs a b c d
  d1 2 1 1 0
  d2 0 2 3 1
```
`2 x 4 sparse Matrix of class "dfm"` is printed by `Matrix::printSpMatrix2()`, so your dfm is actually old one.

kbenoit:
How does it work with the package version as an integer, when our versions are not integers? True, the standard comparison operators won't work for something like 0.99.13 > 0.99.1, but there is some way the R is doing this for comparing updates to packages or enforcing versions in the `Imports:` section of `DESCRIPTION`.

kbenoit:
BTW @koheiw I'm letting this PR "ripen" since I want to us to talk through how to roll out the infrastructure changes (the ones the kill off deprecated functions and older structures), and keep these separate for now from the changes that fix bugs or add minor new features (e.g. `corpus_extract()`). 

Changing a core object is definitely an infrastructural change. Maybe we should start a 1.0 branch and make all major changes to that, while keeping it synced for minor changes with `master`? I'm just concerned about what happens if we want to refresh the CRAN v0.99.x for bug fixes (which we probably already should) but don't want to introduce the major infrastructural changes at the same time.

In any case let's be very careful not to blend in the dfm/corpus/tokens and deprecation removal branches with the minor fixes and updates, for now.

koheiw:
Please try to check version of the package, which is recorded in `@version` of a dfm.

kbenoit:
Got it now!

koheiw:
It is a vector of three integers `c(0, 99, 13)` produced by `packageVersion("quanteda")`. Its ouput is actually a list, but I though it is too much:
```r
> str(packageVersion("quanteda"))
Classes 'package_version', 'numeric_version'  hidden list of 1
 $ : int [1:3] 0 99 1
```

koheiw:
I agree that we should very careful not to mix major infrastructural changes. How about keeping 0.99 only for maintenance releases, while we develop 1.0?

kbenoit:
If we add `as.dfm()` to the beginning of all of dfm methods, and `as.dfm.dfmSparse()` converts to this to a plan "dfm", then we should be able to merge this into master and still not have backwards compatibility issues, right?

koheiw:
I had done this. `as.dfm()` is everywhere.

kbenoit:
Great - I'd thought so. Should we merge this into `master`, or take the more conservative strategy we've discussed before with a parallel `1.0` branch?

koheiw:
Other branches do not make changes in dfm, so we do not need to do merge this yet.

koheiw:
We could work on #55 (one of the issues in the v1.0 milestone) in this branch.

kbenoit:
Agreed. Let's branch a `v1.0` from `master`, and then merge `dev-dfm` to `v1.0`. We can keep `v1.0` synced to any minor changes we make to `master`.

PR #931, opened 2017-08-21
Issue 928

### LINKED ISSUE #928: Error when using purrr to map the dfm() function to a list
The map() function in the purrr package throws an error when passed a list to convert to dfms. The error appears whether the input is a list of characters or corpora.

```{r}
a <- "a"
b <- "b"
c <- list(a, b)
map(c, dfm)

# Error in if (who_called_me_first(sys.calls(), "dfm") == "corpus") { : 
missing value where TRUE/FALSE needed
```

```{r}
a <- corpus("a")
b <- corpus("b")
c <- list(a, b)
map(c, dfm)
# Error in if (who_called_me_first(sys.calls(), "dfm") == "corpus") { : 
  missing value where TRUE/FALSE needed
```

This issue does not occur with other quanteda functions, e.g. corpus().

```{r}
a <- "a"
b <- "b"
c <- list(a, b)
map(c, corpus)
# [[1]]
# Corpus consisting of 1 document and 0 docvars.

# [[2]]
# Corpus consisting of 1 document and 0 docvars.
```

kbenoit:
Fixes #928.

- decrease reliance on who_called_me_first()
- solve #928 purrr::map() problem
- consolidates print messaging in dfm() generic
- adds tests for `dfm()` messages with `verbose = TRUE`
- changes how unused argument warning is given in tokens()


kbenoit:
Note to @koheiw and @HaiyanLW: Appveyor is failing on this temporarily because of https://github.com/RcppCore/RcppArmadillo/issues/161, which is fixed on CRAN just recently but only in the source so far. This is why Travis passes but Appveyor does not, because Appveyor wants to install the Windows binaries and they are not yet updated.

koheiw:
I got it.

Issue #919, opened 2017-08-09
Need to disallow or reprogram `[<-` and `[[<-` for tokens objects


kbenoit:
An alternative would be to update the types and tokens tables to make this possible, but it is ever something we want to allow?

Currently it wreaks havoc:
```r
toks <- tokens(c(d1 = "a b c d", d2 = "c d e"))
toks
# tokens from 2 documents.
# d1 :
# [1] "a" "b" "c" "d"
# 
# d2 :
# [1] "c" "d" "e"

toks[[1]] <- c(toks[[1]], "f")
toks
# tokens from 2 documents.
#
# Error in y + 1 : non-numeric argument to binary operator 

toks <- tokens(c(d1 = "a b c d", d2 = "c d e"))
toks[[1]] <-  c(toks[[1]], 6)
toks
# tokens from 2 documents.
#
# Error in y + 1 : non-numeric argument to binary operator 

toks <- tokens(c(d1 = "a b c d", d2 = "c d e"))
toks[1] <- list(c(1, 1, 1, 2))
toks
# tokens from 2 documents.
# d1 :
# [1] "a" "a" "a" "b"
# 
# d2 :
# [1] "c" "d" "e"

toks[1] <- list(1:6)
# tokens from 2 documents.
# d1 :
# [1] "a" "b" "c" "d" "e" NA 
# 
# d2 :
# [1] "c" "d" "e"
```

PR #933, opened 2017-08-22
Issue 919

koheiw:
Disable assignment operators for #919


R/tokens.R
```diff
@@ -493,6 +493,22 @@ print.tokens <- function(x, ...) {
     x[[i]]
 }
 
+#' @method "[<-" tokens
+#' @export
+#' @noRd
+"[<-.tokens" <- function(x, i, value) {
+    warning('assignment to tokens objects is not allowed', call. = FALSE)
```
kbenoit: Shouldn't this be `stop` instead of warning?

R/tokens.R
```diff
@@ -493,6 +493,22 @@ print.tokens <- function(x, ...) {
...
+    x
+}
+
+#' @method "[[<-" tokens
+#' @export
+#' @noRd
+"[[<-.tokens" <- function(x, i, value) {
+    warning('assignment to tokens objects is not allowed', call. = FALSE)
```
kbenoit: Ditto

PR #939, opened 2017-09-04
Issue 908

koheiw:
This branch adds `corpus_extract()` to remove `what = tag` from `corpus_segment()`.


koheiw:
```r
corp2 <-
corpus(c("##INTRO This is the introduction.
          ##DOC1 This is the first document.  Second sentence in Doc 1.
          ##DOC3 Third document starts here.  End of third document.",
         "##INTRO Document ##NUMBER Two starts before ##NUMBER Three."))
corp2_seg <- corpus_segment(corp2, "other", delimiter = "##",
                            valuetype = "fixed", position = "before")
corp2_seg <- corpus_extract(corp2_seg, '##[A-Z0-9]+', 'tag', remove = TRUE)
head(corp2_seg)

# text1.1                                                   text1.2 
# " This is the introduction." " This is the first document.  Second sentence in Doc 1." 
# text1.3                                                   text2.1 
# " Third document starts here.  End of third document."                                               " Document" 

head(docvars(corp2_seg))

# tag
# text1.1  ##INTRO
# text1.2   ##DOC1
# text1.3   ##DOC3
# text2.1  ##INTRO
# text2.2 ##NUMBER
# text2.3 ##NUMBER
```

kbenoit:
Nice. How about changing `remove = TRUE` to `keep_tag = FALSE` since that's more descriptive, and because `trim` and `remove` are basically synonyms? I'm not 100% happy with `trim` but it is the established term for removing whitespace pads (in **stringi** for instance).

Also should/could we use the standard `pattern` and `valuetype` pair here too, rather than just fixing `pattern` to be a regular expression?

koheiw:
We have to decide what are the best argument names. As I have modified the above code, what extracted by `corpus_extract` is not always tags, so `keep_matched` is more suitable.

Regarding `patterns` and `valuetypes`, do you think `fixed` patterns is useful here? Glob also needs (strange) conversion to avoid * to match white spaces. `corpus_segment` accepts glob patterns, but we need to rethink.

It only removes whitespaces around extracted values using `stri_trim_both()` when `trim` is true, but I am note sure if we need this option in the first place.

kbenoit:
How about: `keep_pattern = FALSE`?

And we remove the `trim` argument and just always trim whitespace at both ends? If someone really really wants this feature added back, we can consider it. (After making fun of them and throwing things.)

I agree it's odd to consider other pattern types, but perhaps we could add whitespace or boundaries automatically to them. Most people would find it much easier to use a glob `pattern = "##*"` than the scary looking regex `"##\\w+\\b"`. We could make it clear that only regex value type pattern matches that can (but also must) include whitespace. What do you think?

kbenoit:
I've been thinking, on the heels of our pre-v0.99 discussion of what to do with the overlap between `corpus_reshape()` and `corpus_segment()`. I like the functionality of `corpus_extract()`, but am not sure about the name or whether instead of simplifying the existing situation (which we postponed), we are not making it worse.

How about this:

*  `corpus_reshape()`: is used as it currently is, to reorganise documents from document, to paragraph or sentence, or (to add) fixed length segments. The basic point is to (as it does now) change the unit of the document to a lower or higher unit preserving meta-data that make this possible. We would add to this the argument `use_docvars = TRUE` currently in `corpus_segment()`. 

    We _might_ try to think of a better name for `corpus_reshape()`, since it does not reshape the width of the object (as does `stats::reshape()` or the **reshape2** package). Maybe the deprecated, original name `changeunits`, to make a `corpus_changeunits()`?

*  `corpus_segment()` takes the form of the current `corpus_extract()`, but with `omit_empty = TRUE`, `use_docvars = TRUE`, and `position = c("after", "before")` added to the function.

    This means we remove the ability to segment to sub-document units, which will be exclusive now to `corpus_reshape()`.  

    It also means we do not add a new function `corpus_extract()`, which is not really an accurate description either, since it suggest that a subset is being extracted, when in fact it's being segmented through pattern matches, with (only the) the pattern really being "extracted".

This proposed scheme solves the overlap between `corpus_reshape()` and `corpus_segment()`, and does not require a third function.

kbenoit:
OK, new idea:

* keep `corpus_reshape()` as is, and this will be for "canned" units such as sentences, paragraphs, and (to be added) fixed-length token sequences.

* `corpus_segment()` will provide finer user control to split documents based in boundaries identified by `pattern`, and give the option to remove them. This is similar to `tokens_segment_by_punctuation` which splits on punctuation and removes the punctuation, before sending this to `textstat_collocations()`. The position argument here tells us that if we do not remove the pattern, then should it be appended to the beginning, or to the end, of the new "document." For punctuation as a delimiter, we would typically append it to the end of the new document. (If `remove_pattern = TRUE`, then this does not matter and has no effect.)
    ```r
    corpus_segment(x, pattern, valuetype = c("glob", "regex", "fixed"), 
        remove_pattern = FALSE, position = c("after", "before"), 
        use_docvars = TRUE, ...)
    ```
* new function for extracting text before or after a user-defined tag:
    ```r
    corpus_extracttags(x, pattern, valuetype = c("glob", "regex", "fixed"), 
        use_docvars = TRUE, position = c("after", "before"))
    ```
but this needs a better name!!

Some examples of how this should work:
```r
txt <- "Sentence one.  Sentence two, sentence three!  Yes."
corp <- corpus(txt)
texts(corpus_segment(corp, valuetype = "regex", pattern = "\\p{P}"))
##         text1.1           text1.2           text1.3           text1.4 
## "Sentence one."   "Sentence two," "sentence three!"            "Yes." 

# why not having the end punctuation does not matter
txt2 <- "Sentence one.  Sentence two, sentence three!  Yes"
corp2 <- corpus(txt2)
texts(corpus_segment(corp2, valuetype = "regex", pattern = "\\p{P}"))
##         text1.1           text1.2           text1.3           text1.4 
## "Sentence one."   "Sentence two," "sentence three!"            "Yes" 


corpus_extracttags(corp, valuetype = "regex", pattern = "\\p{P}", position = "before")$documents
#                  texts  tag _document _docid _segid
# text1.1   Sentence one    .     text1      1      1
# text1.2   Sentence two    ,     text1      1      2
# text1.3 sentence three    !     text1      1      3
# text1.4            Yes    .     text1      1      4

corpus_extracttags(corp, valuetype = "regex", pattern = "\\p{P}", position = "after")$documents
#                  texts  tag _document _docid _segid
# text1.1   Sentence two    .     text1      1      1
# text1.2 sentence three    ,     text1      1      2
# text1.3            Yes    !     text1      1      3
# text1.4                   .     text1      1      4  ## OPTIONAL?

## (but these are not the typical use cases for corpus_extracttags() !)

## similar to existing:
## corpus_segment(corp, what = "tags", delimiter = "\\p{P}", 
##                valuetype = "regex", position = "before")$documents
## (although this apparently contains a bug for first document)
```

kbenoit:
Here's a refinement of the above proposal: 

We keep only a single function, called `corpus_segment()`, that segments a text between the boundaries, where boundaries are defined as `pattern` _and the beginning and end of the documents_. We keep the `position` argument, and change `remove_pattern` to `extract_pattern = FALSE`, which if `TRUE`, removes the pattern text from the new "document", but saves it to a new docvar called `pattern` (or something like that). 

So the equivalent to the `corpus_extracttags()` above would be 

```r
corpus_segment(x, pattern = "##*", extract_pattern = FALSE, position = "after")
```
where the first "pattern" recorded would be `BOF` (beginning of file). If the `position = "before"` then the _last_ tag would be `EOF`. (Alternatively, these could be `NA` - I'm open-minded.)

So we would _not_ add a third function, and the new signature for `corpus_segment()` would be:

```r
corpus_segment(x, pattern, valuetype = c("glob", "regex", "fixed"), 
    extract_pattern = FALSE, position = c("after", "before"), 
    use_docvars = TRUE, ...)
```

Simpler, and consistent as long as we make clear that the "delimiter" is the `pattern` AND the beginning and end of each document.

koheiw:
I updated `corpus_segment()`. Please check if it is working (and update the tests, if possible).

koheiw:
I made `corpus_segment` to remove empty documents, but it started breaking other tests. I am not sure what is the desired behavior. 

I also think that `pattern_remove` should be `remove_pattern` in **quanteda**'s naming rules.

kbenoit:
Make sure you're working with the latest commits, since I fixed many of the tests, except the ones for the tag extraction. In fact all but one now passes.

I'm happy if you want to globally search and replace `pattern_remove` with `remove_pattern`. 

We need this to work:

```r
corpus_segment(corpus("#oneA #one text one #two text three #end"), "#*", 
               valuetype = "glob", pattern_remove = TRUE)
# Error in `[<-.data.frame`(`*tmp*`, field, value = c("#oneA", "#one", "#two",  : 
#  replacement has 4 rows, data has 5 
```

koheiw:
My local rep is updated. I fixed this kind problem:
```r
corp <- corpus_segment(corpus("#oneA #one text one #two text three #end"), "#*", 
                       valuetype = "glob", pattern_remove = TRUE)
texts(corp)
docvars(corp)

> texts(corp)
     text1.1      text1.2 
  "text one" "text three" 
> docvars(corp)
        pattern
text1.1    #one
text1.2    #two
```
but it is falling even in the first test in `test-corpus_segment.R` now :(

kbenoit:
I think the original `omit_empty` argument (now just in the internal function) was designed to allow "empty" documents where we extracted the tag, but left the "document" as empty. This makes no real sense if we define a boundary _also_ as the end or beginning of the file, _and_ do not extract tags, but it makes sense when we use tags and have either a `##TAG1 ##TAG2 not empty` or a `#STARTTAG text #ENDTAG` situation.

Maybe with `remove_pattern`, we should extract ALL matches, and record the empty document? But not do this if `remove_pattern = FALSE`?

koheiw:
If users are interested in tags, it makes sense to extract even if documents are empty. We can keep empty documents when `remove_pattern = FALSE`, but this argument name implies patterns are not important. Why not naming it `save_pattern` or `extract_pattern`?

kbenoit:
Do you mean an added argument name, or are you taking about the new docvar name? I thought we'd just call it`pattern` since that is what is extracted, and only save it if `remove_pattern = TRUE`.

koheiw:
I just mean that `remove_pattern` does not sound like a option to save patterns. The docvar name 'pattern' is fine.

kbenoit:
Should we call it `extract_pattern`?

koheiw:
If keeps empty documents, when `extract_pattern = TRUE` like this:
```r
corp1 <- corpus_segment(corpus("#oneA #one text one #two text three #end"), "#*", 
                       valuetype = "glob", extract_pattern = TRUE)
texts(corp1)
docvars(corp1)

# > texts(corp1)
# text1.1      text1.2      text1.3      text1.4 
# ""   "text one" "text three"           "" 
# > docvars(corp1)
# pattern
# text1.1   #oneA
# text1.2    #one
# text1.3    #two
# text1.4    #end

corp2 <- corpus_segment(corpus("#oneA #one text one #two text three #end"), "#*", 
                       valuetype = "glob", extract_pattern = FALSE)
texts(corp2)
docvars(corp2)

# > texts(corp2)
# text1.1           text1.2           text1.3           text1.4 
# "#oneA"   "#one text one" "#two text three"            "#end" 
# > docvars(corp2)
# data frame with 0 columns and 4 rows
```
but I am feeling that it is good to add `remove_pattern`. When it is `TRUE`, `corpus_segment` only split at the pattern (empty documents will be removed unlike the first example):
```r

corp3 <- corpus_segment(corpus("#oneA #one text one #two text three #end"), "#*", 
                        valuetype = "glob", extract_pattern = FALSE, remove_pattern = TRUE)
texts(corp3)
docvars(corp3)

# > texts(corp3)
# text1.1         text1.2 
# "text one" "text three"
# > docvars(corp3)
# data frame with 0 columns and 4 rows

```

SIBenoit:
OK, so change `extract_pattern` to `remove_pattern`, but only for `char_segment()`? Otherwise this works great now!

I'll approve the PR and you can make the modification for `char_segment()` and merge it.

kbenoit:
OK, so change extract_pattern to remove_pattern, but only for char_segment()? Otherwise this works great now!

I'll approve the PR and you can make the modification for char_segment() and merge it.

Issue #941, opened 2017-09-05
Support hierarchical LIWC


leezu:
[read_dict_liwc](https://github.com/kbenoit/quanteda/blob/b913e0f89369d88788c57db282f4067467068d66/R/dictionaries.R#L530) does not read the hierarchical LIWC format correctly (i.e. LIWC2015_English.dic).

Consider this partial excerpt from the categories:
```
30	affect (Affect)
	31	posemo (Positive Emotions)
	32	negemo (Negative Emotions)
		33	anx (Anx)
		34	anger (Anger)
		35	sad (Sad)
```

Instead of the `posemo` and `negemo` columns the final dataframe only contains a `Emotions)` column.

koheiw:
Thank you for reporting the issue. We were not aware that there is a new LIWC format, and developed `read_dict_liwc` only for the 2007 format. Once we obtained the new dictionary, we will upgrade the importer.

koheiw:
@kbenoit , I am working on this issue, but I do not understand this. Can you explain?
https://github.com/kbenoit/quanteda/blob/88009a4b3ffed07d32997c7282c0ed6ae3d06bc8/R/dictionaries.R#L275-L280

kbenoit:
I'm not sure what that means now! I could probably figure it out later this afternoon, when my meetings are done, by getting back into that code.

koheiw:
The conversion of dictionary to list on line 284 destroyer the structure that allows a key to have both values and sub-keys. For example, in the LIWC 2015 format, 'past' has a value *liked* as well as sub-keys 'cogmech' and 'insight'. Internally, `read_dict_liwc()` constructs a list of lists, but `as.list()` coverts the the element to a list of lists and a character vector named '__'. Dictionary should always be a list of lists, always. 
```
%
11  verb
12  misc
	121 whatever
	122	family
	123	friend
	124	humans
	125	affect
	126	posemo
13  past
	131	cogmech
	132	insight
		133	cause
		134	discrep
		135	tentat
else 
	140 whatever2
	253	time
	464	filler
nothing
%
kin	121	122
kind	<of>131/125	<of>135/126
kinda	131	135
light	140
like	(02 134)125/464	(02 134)126	(02 134)126	253
likeab*	125	126
liked	11	13	125	126 721

```

```r
dictionary(file = "../data/dictionaries/liwc_hierarchical.dic")

# Dictionary object with 4 primary key entries and 3 nested levels.
# - verb:
#     - liked
[...]

Issue #945, opened 2017-09-05
Change dictionary class to dictionary


kbenoit:
To `dictionary` as the S4 class, from the current `dictionary2`, which was created as a transitional object during a period of reprogramming the dictionary class.

koheiw:
I found that `dictionary` collides with `dictionary` in **rlang**, so I kept the current name. If we were to change the name, should be `quanteda_dictionary`.

PR #955, opened 2017-09-11
Issue 941

koheiw:
Adds support of LIWC 2015 format to solve #941.


koheiw:
A hierarchical dictionary prints like this:
```r
Dictionary object with 4 primary key entries and 3 nested levels.
- verb:
  - liked
- misc:
  - whatever:
    - kin
  - family:
    - kin
  - affect:
    - likeab*, liked
  - posemo:
    - like, likeab*, liked
- past:
  - liked
  - cogmech:
    - kinda
  - insight:
    - tentat:
      - kinda
- else:
  - whatever2:
    - light
  - time:
    - like
```
but it is not clear what are keys and what are values in the current format. It would be better to make the distinction clear:
```r 
Dictionary object with 4 primary key entries and 3 nested levels.
- [verb]:
  - liked
- [misc]:
  - [whatever]:
    - kin
  - [family]:
    - kin
  - [affect]:
    - likeab*, liked
  - [posemo]:
    - like, likeab*, liked
- [past]:
  - liked
  - [cogmech]:
    - kinda
  - [insight]:
    - [tentat]:
      - kinda
- [else]:
  - [whatever2]:
    - light
  - [time]:
    - like
```

koheiw:
I can also address  #945, if it is only to change the names of the class. I am not entirely sure 
 why we need to name dictionary 'dicitonary2'.

kbenoit:
`dfm_lookup(x, dict, exclusive = FALSE)` should drop features whose counts have been matched to a dictionary key and counted in that key. If a pre-existing zero-count feature did not match a dictionary entry, then it should be kept at its original value (zero, in this case), just like every other non-matching value.

So more nuanced than just dropping zero-count features.

PR #957, opened 2017-09-11
Dev v1.0.0

koheiw:
A branch for v1.0. You can protect this branch by Github settings.


kbenoit:
I've branched a `v99.9.x` from `master`, and we can use this for any bug maintenance for the current CRAN version. I'll now merge this into `master` and we will start using that for the 1.0 milestone.

PR #962, opened 2017-09-12
Add textstat_frequency plots to manual

stefan-mueller:
Add two examples for plotting word frequencies based on a textstat_frequency data frame (see #900). Preserving the order within facets (second plot) is a bit tricky, so I tried to solve this with base R. If there is a better way of doing it, please let me know (I come across this problem quite often). Once this PR has been edited/approved, I extend the plotting vignette.


kbenoit:
@amatsuo is there a more direct way to order the words by facet, in the example?
(See https://github.com/kbenoit/quanteda/pull/962/files#diff-1943127aa3d223c2b89ed6bc92639970R57)

amatsuo:
I was thinking for a while but could not come up with a better way to do it. By the way, the original example had some minor bugs (`dplyr` was not required although chaining was used, calling non-existent data.frame etc). In the last commit, I fixed them.

PR #976, opened 2017-09-16
Issue 960

koheiw:
This branch contains numerous changes to improve **quanteda**'s performance.  Due to the poor implementation of `cbind.dfm()`, `dfm_lookup()` was unusably slow. It also fixes critical but in `regex2id()`. After merging branch, CRAN version needs to be update.

Both `tokens_lookup()` and `dfm_lookup()` are much faster in this branch:
```r
quanteda_options(threads = 8)

load("/home/kohei/Documents/Brexit/Analysis/data_tokens_guardian.RData")
toks <- data_tokens_guardian
mx <- dfm(toks)

dict15 <- dictionary(file = '/home/kohei/Documents/Dictionary/LIWC/LIWC2015_English_Flat.dic')
dict07 <- dictionary(file='/home/kohei/Documents/Dictionary/LIWC/LIWC2007_English.dic')

    tokens_lookup(toks, dict15), 
    tokens_lookup(toks, dict07), 
    times = 10)

# Unit: seconds
# expr      min       lq     mean   median       uq      max neval
# tokens_lookup(toks, dict15) 73.63517 74.93640 76.79295 76.41722 79.22653 80.77356    10
# tokens_lookup(toks, dict07) 59.45824 60.11967 61.62623 61.17731 62.56146 64.67022    10

    dfm_lookup(mx, dict15), 
    dfm_lookup(mx, dict07), 
    times = 10)

# Unit: seconds
# expr      min       lq     mean   median       uq      max neval
# dfm_lookup(mx, dict15) 21.19515 22.45237 23.65662 22.79392 25.21239 28.90048    10
# dfm_lookup(mx, dict07) 18.84198 19.14635 19.80817 19.89353 20.37558 20.78381    10

```
On windows, both of them compete analysis of 10,000 news articles (subset of Guardian corpus) in around 9 sec, while LIWC takes around 47 sec.


R/dfm.R
```diff
@@ -200,17 +199,39 @@ dfm.tokenizedTexts <- function(x,
...
+                       verbose = quanteda_options("verbose"), 
+                       ...) {
 
     valuetype <- match.arg(valuetype)
     # set document names if none
     if (is.null(names(x))) {
         names(x) <- paste0(quanteda_options("base_docname"), seq_along(x))
     } 
-
-    if (who_called_me_first(sys.calls(), "dfm") %in% c("tokens", "tokenizedTexts")) {
```
kbenoit: This was designed to print a unique message identifying the input object from the call stack. It's far from perfect, but the old message did prevent this:
```r
> dfm(tokenize(data_char_sampletext), verbose = TRUE)
Creating a dfm from a tokenizedTexts input...
Creating a dfm from a tokens input...
   ... lowercasing
   ... found 1 document, 241 features
   ... created a 1 x 241 sparse dfm
   ... complete. 
Elapsed time: 0.018 seconds.
Document-feature matrix of: 1 document, 241 features (0% sparse).
```
Can you fix? (even though `tokenizedTexts` will soon be removed...

R/dfm.R
```diff
@@ -200,17 +199,39 @@ dfm.tokenizedTexts <- function(x,
...
-        check_dfm_dots(list(...), permissible_args = names(formals(tokens)))
```
kbenoit: For arguments not in the signature of `dfm()`, with this change we don't get the error message within `dfm()`, but rather from `tokens()`. Seems like the former behaviour was clearer to the user.

R/dictionaries.R
```diff
@@ -70,11 +70,11 @@ print_dictionary <- function(entry, level = 1) {
 
 
 # Internal function for special handling of multi-word dicitionary values
```
kbenoit: Could you please document these and the other dictionary functions? OK to make this a separate PR.

R/quanteda_options.R
```diff
@@ -1,5 +1,5 @@
 # implement default options
-QUANTEDA_OPTIONS <- list(threads = max(1L, floor(RcppParallel::defaultNumThreads() / 2)),
+QUANTEDA_OPTIONS <- list(threads = max(1L, RcppParallel::defaultNumThreads() - 1),
```
kbenoit: We previously had n-1 cores, where n = all available, but I change it to 1/2 for a reason that now escapes me. Can we think of any reason not to take all but one of a user's cores?

We also want to test that if this is set in `.Rprofile` then it stays set. (This is not possible with **testthat** and therefore needs to be done manually.)

R/tokens.R
```diff
@@ -252,7 +252,7 @@ tokens.tokens <-  function(x, what = c("word", "sentence", "character", "fastest
     
     if (length(regex))
         x <- tokens_remove(x, paste(regex, collapse = '|'), valuetype = 'regex', padding = FALSE)
-    if (!identical(ngrams, 1L))
+    if (!identical(ngrams, 1L) || !identical(skip, 0L))
```
kbenoit: 👍

R/utils.R
```diff
@@ -224,7 +224,7 @@ code <- function(texts){
 
 # convert various input as features to a vector used in tokens_select, 
 # tokens_compound and kwic.
-features2id <- function(features, types, valuetype, case_insensitive, 
+pattern2id <- function(features, types, valuetype, case_insensitive,
```
kbenoit: More documentation please.

kbenoit:
Does `rbind.dfm()` need similar optimisation?

R/quanteda_options.R
koheiw: I set n - 1, because sustained 100% CPU load can make system unstable - sometimes even not able to open task manager. There is a test that directly compare the option value and the environment variable in the test:
```r
expect_equal(
        quanteda_options("threads"),
        as.numeric(Sys.getenv('RCPP_PARALLEL_NUM_THREADS'))
    )
```

R/dfm.R
koheiw: Done

R/dictionaries.R
koheiw: I will do separately

R/dfm.R
koheiw: Added warnings of unused argumetns  within `dfm()`

Issue #69, opened 2015-11-21
Complete Jockers vignette


kbenoit:
With explanation that amounts to both a tutorial and a replication. May involve adding new functionality, such as the `plot.kwic()` added in 0.8.7-9 to produce dispersion plots.

PR #982, opened 2017-09-19
Update LitVignette

stefan-mueller:
Done:
- Change `char_segment()` arguments after update of function (otherwise, vignette won't knit).
- Update text with regard to **readtext** package.
- Switch from `topfeatures()` to `textplot_frequency()`.
- Change first two plots to **ggplot2** objects.

To do?
- Object names are a wild mixture of camelCase and snake_case. Shall we adjust this?
- The last plots (section 4.4 onwards) are not **ggplot2** objects. I didn't change this yet, because base R plotting might be easier for non-`data.frame` objects (but I'm a fan of consistency, so either we use **ggplot2** or base R throughout the vignette).

Let me know what you think and I make the necessary changes.


kbenoit:
Thanks @stefan-mueller that's very useful. We have an open issue for completing this vignette, which is close but not exactly a reproduction of the entire Jockers book. If you are interested in completing this document, that would be great (and would solve #69).

stefan-mueller:
Good idea, @kbenoit, I'm happy to go ahead with a replication of the book. This might take a while, but I will work on this continuously. The author line also sounds good. If we replicate all chapters, we might consider using [**bookdown**](https://bookdown.org) for this project. It will be a very long document (too long for a single article) and **bookdown** offers great functions (e.g., chapters, separate URL, search within book, download book as PDF/epub/mobi).

kbenoit:
Great. Most of it is already done, so it's not likely to take too long. There is a TOC at the beginning of the document, with each section corresponding to a chapter of the book. The idea is to show how **quanteda** can replicate the book in with just a bit of code, so I think **bookdown** is overkill. (Plus it would not then make a vignette-style article for the website.)

PR #988, opened 2017-09-22
Issue 986

koheiw:


kbenoit:
One more change pls: I think the first below should return invisible `TRUE`.  The rest of the returns are correct.
```r
> tmp <- quanteda_options(threads = 4)
> tmp
$threads
[1] 4

> tmp <- quanteda_options(initialize = TRUE)
> tmp
[1] TRUE
> tmp <- quanteda_options(reset = TRUE)
> tmp
[1] TRUE
> tmp <- quanteda_options()
> tmp
$threads
[1] 3

$verbose
[1] FALSE

$print_dfm_max_ndoc
[1] 20

$print_dfm_max_nfeature
[1] 20

$base_docname
[1] "text"
```

koheiw:
I made `quanteda_options()` to return a list when it sets option values, because 
```r
str(options(something = 100))
List of 1
 $ something: num 100
```
but I am happy to change it to invisible `TRUE`.

kbenoit:
OK that's fine, let's make it consistent with `options()`.

PR #990, opened 2017-09-23
Update and extend vignette

stefan-mueller:
Following #69, I almost finished replicating the Jockers book.

Done: 
- Change to snake_case
- Plot with **ggplot2**
- Remove some depreciated functions
- Add chapters 6, 7, 8, 9, 11, 13

To do/discuss: 
- [ ] Didn't know the easiest way to replicate the classification from chapter 12. Could you have a look, @kbenoit.
- [ ] I used built-in data and data from **quantedaData** for chapters 11 and 13 as the `XMLAuthorCorpus` used in the book is only available on the author's website as a .zip-file (http://www.matthewjockers.net/wp-content/uploads/2013/08/TextAnalysisWithR.zip). Yet, even with the State of the Union addresses and Irish debate shows how topic modelling and clustering works.
- [ ] After adding chapter 12 and reviewing my changes, the website needs to be rebuilt.


koheiw:
#1035 could be addressed in this PR.

PR #1005, opened 2017-10-04
Dev lsa

HaiyanLW:


R/textmodel_lsa.R
```diff
@@ -0,0 +1,87 @@
...
+    
+    # return the LSA space
+    return ( space )
+}
+
+#' @rdname textmodel_lsa
+#' @param newX new matrix to be transformed into the lsa space
+#' @param LSAspace previously fitted lsa space
+#' @export
+transform_lsa <- function( newX, LSAspace ) {
```
kbenoit: Do we need a separate function for this, or can this be part of the `textmodel_lsa()` return object?

HaiyanLW: Yes, this function transfers query into the pre-generated LSA space.

HaiyanLW:
I removed the `tf-idf` from within the function because theoretically LSA can be applied to any term-document matrix, `tf-idf` is one of them. 
Do you think we need to make it "LSAspace" class to be compatible with "lsa" package?

kbenoit:
If there are additional functions defined in **lsa** that work with their class of return object, then it would make sense. We did this for the `textstat_simil()` for instance so that functions defined for the `dist` class would work with our output. I'm not sure if this is the case for **lsa**.

HaiyanLW:
There are some other functions defined in 'lsa', however, I think we have similar such as 'textstat_simil()', 'tf-idf()'

kbenoit:
Add a short tutorial/vignette in `vignettes/pkgdown-only/` and add to example in the man page for `textmodel_lsa()`. You can take the example from http://www1.se.cuhk.edu.hk/~seem5680/lecture/LSI-Eg.pdf.

koheiw:
Here is my comparison between **irlba** and **RSpectra**. The latter looks good in speed as well as accuracy.
http://rpubs.com/koheiw/330986

kbenoit:
So the good news is that we are ok sticking with Rspectra?

koheiw:
I am happy with **RSpectra** now. I will keep eyes on **irlba**.

Issue #890, opened 2017-07-28
Add character class valuetype for pattern


koheiw:
This should be easily done through `rege2id`. It would look like this:

```r
tokens_remove(toks, pattern = c('P', 'N'), valuetype = 'charclass')

tokens_remove(toks, pattern = c('Kana', 'Han'), valuetype = 'charclass')
```

koheiw:
This issue needs some discussion to specify its behavior. I think, for example,  `
tokens_select(toks, pattern = 'L', valuetype = 'charclass')` should select token types comprised of ONLY letters. If so, we should only generate a regex pattern `'^[\\p{L}]+$'` and set `valuetype = regex` internally.

`stri_detect_charclass` is faster than `stri_detect_regex` but we cannot select types ONLY with a letters with it:
```r
require(stringi)

toks <- tokens(data_corpus_inaugural)
types <- attr(toks, 'types')

microbenchmark::microbenchmark(
stri_detect_charclass(types, '\\p{L}'),
stri_detect_regex(types, '\\p{L}')
)

# Unit: microseconds
# expr     min       lq      mean   median       uq      max neval
# stri_detect_charclass(types, "\\\\p{L}")  95.544  461.511  591.5141  491.404  578.286 4984.433   100
# stri_detect_regex(types, "\\\\p{L}") 834.743 1753.994 1992.5693 1889.486 2186.172 4506.719   100

stri_detect_charclass('abc', '\\p{L}') # TRUE
stri_detect_regex('abc', '\\p{L}') # TRUE

stri_detect_charclass('abc.', '^[\\p{L}]+$') # Error 
stri_detect_regex('abc.', '^[\\p{L}]+$') # FALSE
```

kbenoit:
We've decided not to do this, since charclass is not really for word matches, and since it can be built into regex matches easily.

PR #1013, opened 2017-10-09
Issue 890

koheiw:
Adds `valuetype = charclass` to solve #890.

### DIFF
--- a/R/dfm_select.R
+++ b/R/dfm_select.R
@@ -7,7 +7,13 @@
 #' @param x the \link{dfm} or \link{fcm} object whose features will be selected
 #' @inheritParams pattern
 #' @param selection whether to \code{keep} or \code{remove} the features
-#' @inheritParams valuetype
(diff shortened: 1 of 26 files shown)


kbenoit:
Thinking about this more, I'm not convinced we should add a `charclass` valuetype, the main reasons being:

1.  A character class matches individual characters, whereas the other value types match multi-character patterns (words). It's therefore not clear what happens when a charclass tries to match terms.

2. Character classes can be nested within the `regex` valuetype. If we wanted the relative efficiency gains of character class matching, why not do what we do with glob -> fixed, and detect when a `regex` can be considered equivalent to a character class, because it contains only a `\\p{}` or a `[ ]` expression? That way we avoid the indeterminacy of 1) above.

koheiw:
I was aware of the problem as in #890, so did not used `string_detect_charclass()`. It simply generates '^[pattern]$' and use `string_detect_regex()`. It is just a short cut to 'regex', which beginners might find useful.

Issue #991, opened 2017-09-25
make deprecated functions defunct


kbenoit:
This includes: 
* anything deprecated functions, e.g. `subset.corpus()` etc.
* deprecated arguments
* older object classes, e.g.
    - `dfmDense`
    - `tokenizedTexts`
    - etc

We should proceed **very carefully** with this however, especially the object classes, making a series of incremental commits and multiple PRs, rather than putting this all into one.

PR #1016, opened 2017-10-11
Remove all the deprecated functions

kbenoit:
Well, almost all - there are still some older non-hashed tokens objects in there, but I promised @koheiw he could have the pleasure of killing them off personally.

Solves #991.


kbenoit:
And are there any C++ functions no longer used, that we should also remove?

koheiw:
Is there any reason to keep `tokens-old.R`?

kbenoit:
I initially deleted `tokens_old.R` entirely, but restored it because `tokens(x, old = TRUE)` calls one of its functions, and removing that stuff seemed part of the tokenizedText transition that we have put aside for now.

PR #1026, opened 2017-10-18
Dev tokens replace

koheiw:
I decided not supporting `regex` and `glob` patters, because they slow down the function. I was also not sure why we need them in `tokens_replace()`. Let's keep this function as simple as possible until someone to request those. Names of `from` and `to` arguments can be better.


kbenoit:
I'm not in favour of exporting `types()`, since it's not grammatically consistent with similar functions, such as `tokens()`. Why expose it?

Also, what creative solutions can we adopt for the arguments (besides, obviously, `x`)? If it were `tokens_replace(x, pattern, replacement, valuetype = c("fixed"))` it would be consistent with **stringi** and with our other functions, and leaves open the possibility of implementing other valuetypes. 

But I'd also like to allow pattern to be a dictionary, where the keys are lemmas and the values the inflected forms, and then replacement would be unnecessary. (This also provides a faster `tokens_lookup()` when all values are single terms and all valuetype = "fixed"). 

We should think think through carefully however since we want `pattern` to be (mostly...) consistent across the functions that use it. Here, it could be:
* a character vector, with a length-matching `replacement` character vector. We would need to think very carefully whether we wanted to allow recycling. In other words, could you supply stopwords for `pattern`, and then replace them al with a single value? This could be allowed, but not other recycling, which would probably result in unanticipated consequences for users who do not understand recycling. (And even for some who do!)  
* a named list of character vectors; if named, we could use the name as the replacement, and the elements as the patterns;
* a dictionary, which would replace the values with the patterns, but this would not operate for token sequences as `tokens_lookup` does.

`dfm` and `collocations` classes would not be valid `pattern` inputs.

koheiw:
We can add `featnames()` to tokens instead of exporting `type()`. If we do not provide type extractor, users will just use `attr(x, 'types')`.

It is more natural to have two arguments in replacement functions like this:
```r
stri_replace_all_fixed(str, pattern, replacement, opts_collator = list())
```
We could create a function that wraps `as.data.frame.dictionary2()` to use a dictionary with `tokens_replace()`.

koheiw:
Strictly speaking, a dictionary object is not a pattern but a pair of pattern (values) and replacement (keys). If we accept this, we make `values()` and `keys()` for a dictionary object, and use them like:
```r
tokens_replace(toks, values(dict), keys(dict))
tokens_select(toks, values(dict), valuetype = "glob")
tokens_select(toks, values(keys), valuetype = "fixed")
kwic(toks, values(dict))
```

kbenoit:
Decided:

1) no recycling of `replacement`, at all
2) if `pattern` is a dictionary, then replace top-level keys with all values in that top-level key and any keys nested within it.
```r
> dict <- dictionary(list(one = c("a", "b"), two = list(two1 = "c", two2 = c("d", "e"))))
> dict
Dictionary object with 2 primary key entries and 2 nested levels.
- [one]:
  - a, b
- [two]:
  - [two1]:
    - c
  - [two2]:
    - d, e
```
    means that a,b -> "one" and c, d, e -> "two"
3) Do nothing with the named character vector idea.

Issue #1025, opened 2017-10-17
Fix testthat 2.0 issues


kbenoit:
Got this email today:

> Dear Kenneth Benoit,
> 
> This is an automated email to let you know that:
> 
> * A new version of testthat is ready to go to CRAN. testthat is
>  currently at version 1.0.2.9000 and will become 2.0.0 upon release.
> 
> * quanteda uses testthat and has problems with the new version.
> 
> * We plan to submit testthat to CRAN on Nov 13.
> 
> I need your help to keep quanteda and testthat working together
> smoothly. In the next month, can you please:
> 
> 1. Read about the changes to testthat at
>   https://github.com/r-lib/testthat/blob/master/NEWS.md#testthat-1029000.
>   This page includes a list of breaking changes, the reasoning behind
>   them, and to how to update your code.
> 
> 2. Carefully inspect the failing checks listed at the bottom of this email.
> 
> 3. For each failing check, either update your package, or tell me
>   that I have a bug. If you have made changes to your package, please
>   submit an update to CRAN before Nov 13.
> 
> If you have discovered a bug in testthat, please file an issue
> (ideally with a small reprex that illustrates the problem) at
> https://github.com/r-lib/testthat/issues. If you're not sure whether
> or not you've found a bug, please an issue and we'll help you figure
> it out.
> 
> Thanks,
> 
> Hadley Wickham

```
== CHECK RESULTS ========================================

*   checking tests ...
    ERROR
   Running the tests in ‘tests/testthat.R’ failed.
   Last 13 lines of output:
     3: compare.default(act$val, exp$val, ..., check.attributes = FALSE)
     4: all.equal(x, y, ...)
     5: all.equal.list(x, y, ...)
     6: all.equal(target[[i]], current[[i]], check.attributes =
check.attributes, use.names = use.names,
            ...)
     7: all.equal.numeric(target[[i]], current[[i]], check.attributes
= check.attributes,
            use.names = use.names, ...)
     8: stop("'tolerance' should be numeric")

     testthat results
================================================================
     OK: 1586 SKIPPED: 18 FAILED: 1
     1. Error: nsyllable works as expected (@test-tokens.R#12)

     Error: testthat unit tests failed
     Execution halted
```

PR #1027, opened 2017-10-21
expect_equivalent will not take multiple test objects

koheiw:
Prepare for the release of testthat 2.0  #1025


koheiw:
I will make the same change to `v0.99.x`.

PR #1038, opened 2017-11-01
Fix a huge bottle neck when keep_acronyms = TRUE

koheiw:
I noticed this because `keep_acronyms = TRUE` by default in  `textstat_collocation()`. The only way to disable this in the function is `tolower = FALSE` at the moment, but I think some people want `tolower = TRUE` and `keep_acronyms = FALSE`. 

We could take out `tolower` and let users do
```r
textstat_collocation(tokens_tolower(x))
# or 
textstat_collocation(tokens_tolower(x, keep_acronyms = TRUE))
```


R/casechange-functions.R
```diff
@@ -17,7 +17,14 @@ tokens_tolower <- function(x, keep_acronyms = FALSE, ...) {
 #' @noRd
 #' @export
 tokens_tolower.tokens <- function(x, keep_acronyms = FALSE, ...) {
-    types(x) <- char_tolower(types(x), keep_acronyms = keep_acronyms, ...)
+    type <- types(x)
+    if (keep_acronyms) {
+        is_acronyms <- stri_detect_regex(type, "^\\p{Uppercase_Letter}{2,}$", ...)
```
kbenoit: Could be refined for the condition `\\p{Uppercase_Letter}(\\p{Uppercase_Letter}|\\d)+` to also get `B2B`, `EC2`, etc.

PR #1044, opened 2017-11-02
Dev network

koheiw:
It seems **igraph**'s error has gone.


kbenoit:
Closer now, and I fixed the errors, but only by commenting out the tests, since the function does not actually run yet.

koheiw:
https://github.com/sctyner/geomnet seems good.

kbenoit:
Questions:

1) The option is a list, where the first value is `type = "python_executable`. Is there another type?

    Why not simplify this to:
    ```
    spacyr_python_executable
    ```

2) Do we still need the option `source_bash_profile`? It looks like the value is being taken from `.Rprofile`, which is always sourced the R is started.

    Probably we could go with an .onLoad() function in a `zzz.R`, so that it initialises the settings to the defaults iff the option is not already set (which it would be if set in `.Rprofile`). The argument in `spacy_initialize()` can then be:

    ```
    spacy_initialize(model = "en", python_executable = getOption("spacyr_python_executable"), ...)
    ```

    The default for this option would be `NULL`, so that it searches for it unless it has been set by the user.

koheiw:
This is ready for inspection. I implemented without (abandoned) `ggnetwork` or `geonet`.

PR #1048, opened 2017-11-03
Add min_nchar and max_nchar to tokens_select()

koheiw:


kbenoit:
@koheiw any idea why my test of `as.dfm.dfmSparse()` fails on Travis (only)? I full admit it has nothing to do with this branch, but I noticed it was a red item on codecov so added it. I think the antipodal gravity here in Australia has diverted blood flow from my brain, because I am not figuring it out. (And probably should never have added it to this PR to begin with!)

PR #1052, opened 2017-11-05
Issue 1051

koheiw:
This branch adds `nested` argument with minimal change in the current logic. If `nested = FALSE`, output do not contain nested collocations.

This also makes computation faster even when `nested = TRUE`:

New
```
Unit: seconds
 nested 1.877418 1.890013 1.910341 1.899149 1.926853 1.965251    10
```
Old
```
Unit: seconds
 nested 2.363261 2.484385 2.634432 2.667844 2.769005 2.923769    10
```


kbenoit:
Speed improvement is always good!

Here's a test I ran:
```r
txt <- c("a b c . . a b c . . a b c . . . a b c",
         "a b . . a b . . a b . . a b . a b")
toks <- tokens(txt) %>% tokens_keep(c("a", "b", "c"), padding = TRUE)
toks
# tokens from 2 documents.
# text1 :
# [1] "a" "b" "c" ""  ""  "a" "b" "c" ""  ""  "a" "b" "c" ""  ""  ""  "a" "b" "c"
# 
# text2 :
# [1] "a" "b" ""  ""  "a" "b" ""  ""  "a" "b" ""  ""  "a" "b" ""  "a" "b"

textstat_collocations(toks, size = 2:3, nested = TRUE)
#   collocation count length    lambda         z
# 1         a b     9      2  6.876265  3.377673
# 2         b c     4      2  3.970292  2.520030
# 3       a b c     4      3 -3.891820 -1.201886

textstat_collocations(toks, size = 2:3, nested = FALSE)
#   collocation count length    lambda         z
# 1         a b     9      2  6.876265  3.377673
# 2       a b c     4      3 -3.891820 -1.201886
```

Based on that, here are a few comments on the nested part:

1. I don't really like `nested` as an argument, since it conveys what it is about but is very hard to figure out from the logical value what it does. Maybe something like `recursive` instead? But we want something whose behaviour is clear from the name.

2. Above, the first example is the current behaviour. This makes sense, since it first scores trigrams, then bigrams, and the `a b` shows up 5 times from `a b` and another four within `a b c`. From `a b c`, `b c` occurs 4 times. This is the same as pasting two separate collocations scoring for `size = 2` and `size = 3`, respectively.

3. For `nested = FALSE`, I would expect that `a b` should be counted just 5 times, not 9, since (as with `b c`) its occurrence with `a b c` is not counted separately. You probably intended this, so my guess is that this outcome is unintended and you will correct it - but correct me if I got this wrong.

4. The judgment of what constitutes a longer sized collocation would really need to depend on a threshold value (*p*-value). This might be in your code, but I could not find it. We would probably want to let the user change this setting, however, to decrease or increase the sensitivity of what constitutes a longer collocation. 

5. I could not explain this behaviour, which seems to be the same for both values of `nested`:

```r
txt2 <- c(". . . . a b c . . a b c . . .",
         "a b . . a b . . a b . . a b . a b",
         "b c . . b c . b c . . . b c")
toks2 <- tokens(txt2) %>% tokens_keep(c("a", "b", "c"), padding = TRUE)

textstat_collocations(toks2, size = 2:3, min_count = 1, nested = FALSE)
#   collocation count length    lambda         z
# 1         a b     7      2  4.966833  3.173907
# 2         b c     6      2  4.572417  2.947850
# 3       a b c     2      3 -6.676832 -2.233106
textstat_collocations(toks2, size = 2:3, min_count = 1, nested = TRUE)
#   collocation count length    lambda         z
# 1         a b     7      2  4.966833  3.173907
# 2         b c     6      2  4.572417  2.947850
# 3       a b c     2      3 -6.676832 -2.233106

textstat_collocations(toks2, size = 2:3, min_count = 2, nested = FALSE)
#   collocation count length    lambda         z
# 1         a b     7      2  4.966833  3.173907
# 2         b c     6      2  4.572417  2.947850
# 3       a b c     2      3 -6.676832 -2.233106
textstat_collocations(toks2, size = 2:3, min_count = 2, nested = TRUE)
#   collocation count length    lambda         z
# 1         a b     7      2  4.966833  3.173907
# 2         b c     6      2  4.572417  2.947850
# 3       a b c     2      3 -6.676832 -2.233106
```

kbenoit:
There is one more issue, in that by letting `textstat_collocations()` not count nested, lower-sized occurrences, then we are modifying the counts and therefore the stats. So the behaviour in 3 above was to count `a b` just 5 times (not 9) then we would be changing the stats in a way that produced a different score when this function was run alone with `size = 2`, versus with `size = 2:3, nested = TRUE`. This unsettles me a bit. (So maybe your default behaviour that I thought was wrong was actually right?)

koheiw:
I changed the argument to `recursive`, because it is exactly what it is. The funny collocation counts when `recursive = FALSE` was simply due to an bug in making a data.frame. 

`recursive = FALSE` and `recursive = TRUE` produces the same lambda values, because I did not change the estimation part. There was a single vector, but I separated it into two, for presentation (returned to users) and estimation (internal purpose only). Those two are more ore less the same when `recursive = TRUE`, but the presentation vector only contain those surrounded by boundaries when `recursive = TRUE`. When the presentation vector is shorter, computation becomes quicker. The estimation vector always contain all the sub-sequences to make statistics correct.

koheiw:
There are a lot of unused veriables `dice, pmi, logratio, chi2, gensim, lfmd` in the C++ code. Should clean them up?

kbenoit:
not yet, since we may use those in the paper on collocations (but they will not be exposed in a published release).

kbenoit:
The bug is gone, but I am concerned about the statistics.

We cannot have different counts, but the same stats. Let's get Jouni's opinion on this.

```r
textstat_collocations(toks2, size = 2:3, min_count = 2, recursive = TRUE)
#   collocation count length    lambda         z
# 1         a b     7      2  4.966833  3.173907
# 2         b c     6      2  4.572417  2.947850
# 3       a b c     2      3 -6.676832 -2.233106

textstat_collocations(toks2, size = 2:3, min_count = 2, recursive = FALSE)
#   collocation count length    lambda         z
# 1         a b     5      2  4.966833  3.173907
# 2         b c     4      2  4.572417  2.947850
# 3       a b c     2      3 -6.676832 -2.233106
```

kbenoit:
Consider two collocation sequences, "Trinity College Dublin" (TC) and "Trinity College" (TC). The purpose of this PR is to detect "TCD" and "TD" separately, but not count "TD" and "CD" from "TCD" when they occur within the trigram.

Probably the best pipeline is:

1. Score trigrams, and compound the ones for which p >= .05. This would detect and compound "Trinity College Dublin" as "Trinity_College_Dublin".
2. Score the compounded tokens resulting from 1, which would capture "Trinity College" separately.
3. The stopping point for this would be bigrams, since there is no lower compound than that.

koheiw:
What I need is nothing do with estimation actually. Only involving presentation, the argument can be callled `include_only_nested`. When it is FALSE, nested-only collocations (`c d`) are hidden.

```r
txt <- c(". . . . a b c . . a b c . . . c d e",
         "a b . . a b . . a b . . a b . a b",
         "b c d . . b c . b c . . . b c")
toks <- tokens(txt) %>% tokens_keep(c("a", "b", "c", "d", "e"), padding = TRUE)
textstat_collocations(toks, size = 2:3, include_only_nested = TRUE)

#  collocation count length    lambda         z
#1         b c     6      2  3.559572  3.440955
#2         a b     7      2  5.079628  3.248098
#3         c d     2      2  3.831054  2.355034
#4       a b c     2      3 -5.669191 -2.055946

textstat_collocations(toks, size = 2:3, include_only_nested = FALSE)

#  collocation count length    lambda         z
#1         b c     6      2  3.559572  3.440955
#2         a b     7      2  5.079628  3.248098
#3       a b c     2      3 -5.669191 -2.055946
```
Alternatively, we can do the same thing by added a column called `count_nested`, and subset by a condition `count != count_nested`. The column is easy to add, but very informative.
```r
#  collocation count count_nested length    lambda         z
#1         b c     6            3      2  3.559572  3.440955
#2         a b     7            2      2 5.079628  3.248098
#3         c d     2            2      2 3.831054  2.355034
#4       a b c     2            0      2 -5.669191 -2.055946 
```

jounikuha:
Unless I am missing something, I agree with Ken. What should happen when recursive=FALSE  should match what would happen if we did this by hand, step by step (and, by the way, I also agree that it is not obvious which setting should be FALSE and which TRUE here)

Consider two such manual steps, one for trigrams and one for bigrams. Suppose that in the trigram step we decide that "Trinity College Dublin" is a three-word expression, For the bigram step it is then compounded ("Trinity_College_Dublin") and treated as a single word. For example, "... students in Trinity College Dublin protested about..." becomes "...students in %%% protested about..." for the bigram step, where "%%%" represents any previously compounded expression. In the bigram step any %%% will be treated as a single word for counting purposes, but two-word expressions involving %%% (such as "in %%%" and "%%% protested" here) will not be considered as candidate multiword expressions and not included in the count tables (because that would be a four-word expression, which would have been detected and compounded in the 4-gram step if it was real). 

Because of the compounding, the bigram step uses in effect a different text, and thus all the statistics will change. This is not just for the bigrams included in %%% such as "Trinity College" and "College Dublin", but the total number of bigrams will change. In the example above they go down by 2 since we go from {"in Trinity", "Trinity College", "College Dublin", "Dublin protested"} to {"in %%%", "%%% protested"}. This changes the counts and statistics even for "Oxford University", because there are now 2 fewer two-word sequences where the first word is not "Oxford" and the second word is not "University". 

In general, what recursive=FALSE is trying to do here is quite elaborate, because it is trying to replicate this sequence of analyses at different lengths, and the transitions between them. This includes not only the re-counting of words at each step but also, as Ken noted, the decision of which expressions should be compounded at each step. This can be automated (e.g. based on a p-value) but this would probably be only a rough proxy for decisions which we would really want to make with more deliberation. So it may be that recursive=FALSE is really trying to do a bit too much. One option might be to have it do only the counts (those still seem useful) but not the statistics.

kbenoit:
@koheiw Would an option be to produce the full list of collocations (as with `recursive = TRUE`, above) and then run a command on those collocations to rearrange the list? This would solve your presentation needs, but not run afoul of the count and statistical issues that worry Jouni and me.

koheiw:
Thank you @jounikuha for your thought, but the issue is not tokenization. What I am trying to achieve here is making a list of collocations as a summary of texts. This list therefore should only include substantively meaningful collocations. Since statistics from different documents are incomparable, such a multi-pass method does not help.

Further, even for tokenization purpose, iterative approach does not seem to work in the way as we hoped. Please have a look at [my experiment](http://rpubs.com/koheiw/329712) (with 10,000 news articles). You can easily notice that collocations of rare collocations are over-represented from the second run.

I think the root problem is the way we count the nested sequences. When we are interested in three types 'A', 'B', 'C' (e.g. Trinity College Dublin or New York Times), `B C` appear in at least three ways `_ B C`, `* B C` or `A B C`, where `_` is ineligibles (punctuations), `*` is anything but `A` `B` or `C`. Only the first two are the natural collocations and the last one is artificial, but we treat all those `B C` as the same.

In the one-pass method (at the end of the experiment), you can see 'York Times`, which is a purely artificial collocation. When sizes of collocations are larger than 2, nested sequences appear higher position than nesting sequences ('New York Times') almost always because they are more frequent. We know that there is not such newspaper in the British city of York, but it can be hugely misleading in other cases. For example, occurrences of 'Stop Brexit Campaign' will promoter 'Brexit Campaign' always, and can not be more prominent than its rival group in the collocation analysis even if so in the real world.

A simple solution is to show the nested frequencies separately from the total as in my last post. The other simple solution is hide collocations that appear only as a sub-sequence. More fundamental solution is to treat nested collocations completely differently by adding a special marker that indicate adjacency to boundaries (denoted by `_`): `'', 'A', 'B', 'C', ''` entails `_ A B C _`, `_ A B`, `B C _`; `'', 'B', 'C', ''` does `_ B C _`. This means we test up to n + 2 interactions for lambda (B&J actually did this).

kbenoit:
Discussed, and decided:

* don't add additional arguments, since this seemed somewhat confusing.
* OK to add a new column to the collocations output, called `count_nested`, which is always <= `count`, that counts the sequence for which it occurs, when it occurs within a larger sequence _in the output_. This means that nested counts of bigrams for `size = c(2, 4)` will only count the bigrams nested within the 4-grams that are identified (and not also trigrams that did not occur within a counted bigram).
* Following the above point, we should consider whether to count the nested lower-grams for higher-grams that did not occur in the output because they did not meet `min_count`, such as the `c d e` and `c d e` above.
* when `size` is length 1, `count_nested = 0`.

koheiw:
I noticed that it is nearly impossible to count nested collations correctly (defined as elements of collocations with a larger size) in the current design, because collocations are counted separately for each size.

The advantage of this design is that it does not store collocation counts for different sizes, so memory consumption is smaller. However, it turned out that the memory usage is rather small (for 10,000 documents from the Guardian corpus), so I think we can change this. 
```
Size: 2
Memory: 22594KB
Size: 3
Memory: 50769KB
Size: 4
Memory: 72443KB
Size: 5
Memory: 78953KB
```

kbenoit:
I'm not sure I understand what you mean here, which part refers to the current system versus the PR. "Change this"...?

koheiw:
I haven't made the change to this branch yet.

The current master's collocation looks like (in pseudo-code):

```r

for (size in sizes) {
    col_count <- NULL
    for (document in documents) {
        col_count <- count(toks)
        estimate(col_count, size)
    }
}
```
I am proposing to change to:
```r
col_count <- NULL
for (document in documents) {
    for (size in sizes) {
        col_count <- count(toks)
    }
}
for (size in sizes) {
    estimate(col_count, size)
}
```

kbenoit:
As long as the behaviour of the existing columns are unchanged - we decided [above](https://github.com/kbenoit/quanteda/pull/1052#issuecomment-344336703) - then that's fine.

jounikuha:
Delayed response on Kohei's comment above. Two points:

* My earlier comment was about what happens if a single function call tries to approximate an iterative/recursive procedure of the kind discussed in my next point below. This is inevitably a bit hazy because it requires automatic decision rules for which (say) three-word sequences are set aside when considering two-word expressions, and so on. But if it is done, then the statistics must change depending on which longer sequences have previously been set aside, i.e. the set of counts will be different (say) for 3-word than for 2-word expressions. 

* More generally, I do think an iterative procedure, done manually, must work at least in a certain sense. This is when at each stage of it we consider *only* expressions of a given number of words, but not shorter ones. This would mean, for example, that from the 3-word round we may conclude that "New York Times" is a 3-word expression. In this case it is then treated as one word in the 2-word round, so that, for example, no appearances of "York Times" will be counted in that round (unless they actually appear without the "New").

koheiw:
The C++ code is almost completely rewritten to count nested elements correctly. It is still a bit faster than the master:
```r
toks <- data_tokens_guardian[1:1000]
toks <- tokens_remove(toks, stopwords("english"), padding = TRUE)
toks <- tokens_remove(toks, "\\p{S}", "regex", padding = TRUE)

    textstat_collocations(toks, size = 2:5, min_count = 10),
    times = 10
)
```
This branch:
```
Unit: seconds
                                                    expr      min       lq    mean   median       uq
 textstat_collocations(toks, size = 2:5, min_count = 10) 10.66965 10.73168 10.9334 10.81152 10.90579
      max neval
 11.99334    10
```
Master
```
Unit: seconds
                                                    expr      min       lq     mean   median       uq
 textstat_collocations(toks, size = 2:5, min_count = 10) 12.02498 12.11577 12.23151 12.17733 12.30681
      max neval
 12.66076    10
```

koheiw:
Its output looks like this:
```r
> head(out, 20)
        collocation count count_nested length   lambda        z
1         last year   317          230      2 5.713661 73.78798
2   chief executive   201           93      2 8.996196 57.35800
3         years ago   118          114      2 7.302860 49.87405
4     david cameron    97           48      2 8.175249 48.69433
5    prime minister   189          114      2 9.580599 48.40240
6        last month    98           84      2 5.712537 43.61346
7        last night   102           92      2 6.351813 43.25848
8     civil service    63           16      2 7.807473 41.17707
9       social care    53           35      2 6.842488 38.94888
10     central bank    57           49      2 6.977028 38.27789
11        last week    80           56      2 4.980032 38.14488
12       five years    67           40      2 5.660842 37.56095
13   private sector    48           21      2 6.781269 37.23363
14      high street    48           33      2 6.799871 37.09539
15     human rights    54           40      2 9.393044 36.75163
16       six months    50           29      2 6.644396 36.58431
17     three months    56           40      2 5.579267 36.01129
18 stephen lawrence    38           14      2 8.162991 35.10497
19        two years    72           44      2 4.534831 34.63617
20      three years    64           47      2 4.862295 34.41250

> head(out[out$count_nested / out$count < 0.5,], 20)
         collocation count count_nested length    lambda        z
2    chief executive   201           93      2  8.996196 57.35800
4      david cameron    97           48      2  8.175249 48.69433
8      civil service    63           16      2  7.807473 41.17707
13    private sector    48           21      2  6.781269 37.23363
18  stephen lawrence    38           14      2  8.162991 35.10497
21        first time    70           21      2  4.544450 34.29173
25     phone hacking    37           12      2  8.829746 33.24626
27     rick santorum    33           16      2  9.113432 32.29827
28  northern ireland    30           10      2  8.723266 32.05791
29    european union    33           14      2  6.828570 31.98504
30         make sure    40            9      2  6.706920 31.78001
32       ed miliband    28            5      2  8.422060 31.43097
33       white house    33           10      2  7.109945 31.34186
34      young people    52           16      2  5.217777 31.27099
36      PIP implants    29           13      2  7.531617 30.96870
37 local authorities    31            9      2  6.773138 30.93990
38        royal bank    36            4      2  6.145207 30.87909
42   leveson inquiry    29            3      2  7.674808 29.91743
44     executive pay    34           16      2  5.779558 29.58478
45 liberal democrats    33            7      2 10.043530 29.38251
```

PR #1054, opened 2017-11-06
Define ARMA_DONT_PRINT_OPENMP_WARNING to stop printing warnings

### LINKED ISSUE #1021: textmodel_wordfish() crashes
This bug was not detected in CI, but happens on my machine:
```r
ie2010dfm <- dfm(data_corpus_irishbudget2010)
wfm_d <- textmodel_wordfish(ie2010dfm, dir = c(6,5), sparse = FALSE) # Works
wfm_d <- textmodel_wordfish(ie2010dfm, dir = c(6,5), sparse = TRUE)  # Crashes

*** caught segfault ***
address 0x300000018, cause 'memory not mapped'
```
@HaiyanLW please check.

koheiw:
This may not fix #1021, but stops the annoying warnings printed out multiple times. I think it is also worth experiment with cpp14.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: quanteda
-Version: 0.99.9003
+Version: 0.99.9004
 Title: Quantitative Analysis of Textual Data
 Description: A fast, flexible, and comprehensive framework for 
     quantitative text analysis in R.  Provides functionality for corpus management,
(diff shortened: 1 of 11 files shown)


kbenoit:
You're right, that mitigated the warnings but did not eliminate them. 

I briefly tried CXX14 on this branch but reverted it since it only worked on clang4/macOS, not on stock Ubuntu 16.04 or on Windows 10 with RTools. I read that RTools does not support C++14 yet. So we should stick with C++11.

PR #1058, opened 2017-11-08
Add article for comparison of packages

stefan-mueller:
Changes:
- Add article for docs.quanteda.io that compares packages for text analysis.

Problems:
- csv file needs to be moved to quanteda repo (which folder?)
- Layout of the table needs improvement (alternative to `kable()`?)


kbenoit:
@stefan-mueller Thanks a lot for this. I made a few formatting changes, and saved a copy of the table locally. It's now integrated into the articles for the **pkgdown** site.

PR #1062, opened 2017-11-10
As as.tokens() method for spacyr_parsed objects

kbenoit:
This adds POS information in the simplest possible way: by concatenating them (optionally) as `/POS`. This allows us (for now) to work completely within existing functions, and to select tokens from POS using  e.g. pos selection using 

```r
toks <- spacy_parse("This is a test.") %>% as.tokens(include_pos = "pos")
toks
# tokens from 1 document.
# text1 :
# [1] "This/DET"  "is/VERB"   "a/DET"     "test/NOUN" "./PUNCT"  

tokens_select(toks, "*/DET")
# tokens from 1 document.
# text1 :
# [1] "This/DET" "a/DET" 
```


kbenoit:
@koheiw Let me know what you think of the argument syntax (could be improved...) and whether this should (also?) be a `tokens()` method.

PR #1065, opened 2017-11-11
Fix linguist

koheiw:
I notice that you are using `linguist-vendored` but did you try `linguist-documentation`


kbenoit:
Not showing as fixed immediately but maybe it takes some time for the GitHub cache to update? Fingers crossed.

Issue #1067, opened 2017-11-14
dfm(x, select = ...) does not take a dfm for selection


kbenoit:
Works in stages, but not directly. This should work fine, and used to.

```r
txt1 <- c("one two three three", "one two two four")
txt2 <- dfm("two four five")

dfm(txt1) %>% dfm_select(dfm(txt2))
# Document-feature matrix of: 2 documents, 3 features (50% sparse).
# 2 x 3 sparse Matrix of class "dfm"
#        features
# docs    two four five
#   text1   1    0    0
#   text2   2    1    0

dfm(txt1, select = dfm(txt2))
# Error in pattern2id(pattern, type, valuetype, case_insensitive, attr(x,  : 
#   dfm cannot be used as pattern 
```

kbenoit:
This is a trickier issue since we call `tokens_select()` as part of the dfm construction process in https://github.com/kbenoit/quanteda/blob/master/R/dfm.R#L252-L261.

koheiw:
It is tricky, but I am not sure if we change this, because it is natural that selection by dfm is available only in `dfm.dfm()`.

```r
txt1 <- c("one two three three", "one two two four")
txt2 <- "two four five"
dfm(dfm(txt1), select = dfm(txt2))

Document-feature matrix of: 2 documents, 3 features (50% sparse).
2 x 3 sparse Matrix of class "dfm"
       features
docs    two four five
  text1   1    0    0
  text2   2    1    0
```

kbenoit:
Fair enough - but let's add a note to the output error message explaining this, if someone tries `dfm.tokens/corpus/character(x, select/remove = any_dfm)`.

koheiw:
Yes, we have to add value check.

PR #1070, opened 2017-11-17
Issue 1067

koheiw:
For #1067. I also wonder if we should change the argument name from `select` to `keep`.


kbenoit:
We don't use `keep` as an argument, although it is one of the values of `selection` in e.g. `tokens_select()`. I think our usage of `select` is consistent, since we use "select" v. "remove" everywhere else.

PR #1078, opened 2017-11-22
Dev text2vec

koheiw:
Job done.


koheiw:
Many of the functions that uses **armadillo** are failing becasue ARMA_64BIT_WORD is (correctly) defined before linking to the header. They are all OK if ARMA_64BIT_WORD is disabled, but I think it should be enabled.

```
error: arma::memory::acquire(): out of memory
..........
test quanteda_options: ........S....
test regex2fixed.R: .....................
......test spacyr methods: ........
test stopwords.R: ...
tests of textmodel_mixfit: ..
Testing textmodel-ca.R: .........2
error: arma::memory::acquire(): out of memory
.........3..
error: arma::memory::acquire(): out of memory
..
Testing textmodel-lsa.R: W...W..
test textmodel_nb: ..........................................
Testing textmodel-wordfish.R: S456
error: arma::memory::acquire(): out of memory

error: arma::memory::acquire(): out of memory

error: arma::memory::acquire(): out of memory
7
error: arma::memory::acquire(): out of memory
8
error: arma::memory::acquire(): out of memory
9
test textmodel_wordscores: 
error: arma::memory::acquire(): out of memory
..W..............W......
```

koheiw:
I also think `threads` of `textmodel_ca()` should be removed, because we have `quanteda_options(threads)` now.

kbenoit:
On `textmodel_ca()`, we should change the default to `threads = quanteda_options("threads")`, as per `textmodel_wordfish()`.

koheiw:
`textstat_wordfish()`'s `threads` argument actually doing nothing, so it should be remove as well. C++ code (TBB) will read the setting via environment variable.

kbenoit:
Very nice, it's now faster than `text2vec::tcm` and more efficient:

```r
# text2vec 
t1 <- proc.time()
vectorizer = vocab_vectorizer(vocab)
# use window of 5 for context words
tcm <- create_tcm(it, vectorizer, skip_grams_window = 5L)
print(proc.time() - t1)
##   user  system elapsed 
## 18.303   0.446  18.802 

# quanteda
system.time(qtcm <- fcm(wiki_toks, context = "window", count = "weighted", weights = 1/(1:5), tri = TRUE))
##  user  system elapsed 
## 3.822   1.073   4.910 
```
and smaller because it's a `dgCMatrix`:
```r
format(object.size(tcm), units = "MB")
## [1] "221.2 Mb"
format(object.size(qtcm), units = "MB")
## [1] "168 Mb"
```
:+1:

PR #1079, opened 2017-11-22
Fix GitHub language detection

chmue:
The `html` files not covered by `vignettes/*`were a few files in `tests` and `inst/doc/quickstart.html` which are automatically generated from `*.Rmd`.

The count gets heavily inflated because GH counts file size and RMarkdown bundles a lot of JS and CSS in the generated html files which makes a short Rmd-generated html ~300KB which is a lot more than even the longest plain text code file.

---

I should have added that I checked that new stats with the local version of GitHub Linguist (the Ruby gem they provide) which looks like this:

```
quanteda-fork [master] % ~/.gem/ruby/2.4.0/bin/linguist
86.99%  R
13.00%  C++
0.01%   C
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,8 @@
-vignettes/*	linguist-documentation
-docs/*	linguist-documentation
-man/*	linguist-documentation
+# Fix GitHub Linguist language detection
+# docs/ and man/ are automatically excluded
+vignettes/*       linguist-documentation
+tests/**/*.html   linguist-generated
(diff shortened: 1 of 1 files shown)


chmue:
I should have added a flag for excluding it from CI as there are no actual changes in the code 🙁

kbenoit:
Brilliant! And thanks for the explanation!

chmue:
Great to see it working on the actual website now!

Issue #1077, opened 2017-11-21
Create internal textstat_collocations() with testing options


kbenoit:
For our research into collocation scoring with @jounikuha (and @koheiw and @HaiyanLW) we currently have additional options for `textstat_collocations()` in the branch `dev-collocations`, but this has falling behind `master` in development. I propose we create an internal function called `textstat_collocationsdev()` that is like `textstat_collocations()` but with the additional options for other methods and for returning the diagnostic results (actual and expected counts). Then they will not get out of sync.

We can mark it `@keywords internal` so it does not show up in the function index.

Note that this will incorporate the code that performs #924 and then you can close that issue.

PR #1083, opened 2017-11-23
Solved issue #1077

HaiyanLW:


kbenoit:
Perhaps just to avoid the big hit on codecov, you could copy the tests for collocations into a new test file, and search-replace the calls to `textstat_collocations()` with `textstat_collocationsdev()`? Beyond the obvious desire to keep codecov high, the tests are always a good thing! We ought to compare the core results for the two functions just to make sure they are identical (omitting the extra stuff that comes from `textstat_collocationsdev()` of course).

PR #1086, opened 2017-11-24
Add margin argument to textmodel_lsa

koheiw:


koheiw:
Codecov is failing only because of the experimental code in `collocations_mt.cpp`

PR #1089, opened 2017-11-26
Fix convert

koheiw:
- Improve handling of empty documents
- Hide dfm2* functions from users
- Add the empty_omit option that is accessible via dots


koheiw:
I noticed that the `topicmodels::LDA` errors with this branch when a dfm is weighted, because `LDA` accepts only unweighted counts. In the master branch, it `LDA` works because `convert(x, 'topicmodels`) floor all the weighted infrequences. Is this an intended behaviour?

```
require(quanteda)
count_dfm <- dfm(data_corpus_inaugural)
tfidf_dfm <- dfm_weight(count_dfm, 'tfidf')
tfidf_lda <- convert(tfidf_dfm, 'topicmodels')

table(tfidf_dfm@x > 0) # 43723
table(floor(tfidf_dfm@x) > 0) # 18860
table(tfidf_lda$v > 0) # 18860
```

koheiw:
Convert functions has been internally `omit_empty = TRUE` because some models do not accept matrix with empty documents. It really inconvenient to have  `ndoc(converted) != ndoc(original)`, so I actually wanted to set `omit_empty = FALSE`.

What do you think about automatic flooring by `convert`? This is a serious issue, because it removes a lot of features.

kbenoit:
Ah yes I remember why I added omit_empty now. We might need to revive that.

There should be no automatic flooring, and if a dfm whose term weight is not “unary” is converted to one of the topic model objects, we should stop with an error. New issue?

PR #1094, opened 2017-11-29
Update pkgdown

koheiw:
This PR makes the pkgdown site much more attractive.


koheiw:
I removed vignette from `/inst`, because of the search engine issue. CRAN pages is the last place people should land. There are links to quanteda.io on CRAN page. If you still think we need a vignette, let's make very short document like this, and insert links to docs.quanteda.io.
https://cran.r-project.org/web/packages/tidyverse/vignettes/manifesto.html

It is easy to incorporate the new demo.

kbenoit:
I'm happy to replace the vignette with something very minimal, or even remove it altogether, if that causes more pointing to the website instead of the CRAN page. (I think this is a function of links, and we can add those to other sites to get the page rank higher.)

What's the "best practice" advice re vignettes, if we have all that and more elsewhere?

PR #1098, opened 2017-11-30
Update pkgdown2

koheiw:


kbenoit:
I just did a reorg on the articles, and restored a vignette in the form of the explanation of the package design. I will cut this down later. Please review.

Also: Can we bring back a few badges for the home page? How about just:
* CRAN
* Total downloads
* (maybe) DOI

PR #1100, opened 2017-11-30
Fix formatting issue in README.md

stefan-mueller:


kbenoit:
Thanks @stefan-mueller . I went ahead and fixed this directly in the `master`. Remember, we only edit the .Rmd,the .md is produced from knitting. 

Also I added you to the contributor list!

stefan-mueller:
Oh that's great, thanks for adding me officially. I'll keep the `.md/.Rmd` difference in mind.

Issue #1117, opened 2017-12-05
feature request: function for network data as prestep for a network plot


methodds:
Hi,

related to @kbenoit 's suggestion in  #782, it would be very useful to have a function, e.g. `textstat_network()` which stores data for semantic networks, either in a df or in a network object. The current function `textplot_network()` could then take this as input for creating network plots. I think this has two advantages:
- it would probably be easier to implement alternative ways of creating networks in the future.
- users who would like to do more than just plotting a network could use the object from `textstat_network()` for their needs.

koheiw:
I will add `as.network` method for fcm. I was thinking of doing this on request.

PR #1118, opened 2017-12-05
Add as.network.fcm for #1117

koheiw:


kbenoit:
I don't like the idea of using `min_freq` in both `textplot_network()` and `as.network()`, since we have `dfm_trim()`. It would be straightforward to add an `fcm_trim()` and then require the dfm/fcm to be transformed before calling the network function. 

Not only does the current function duplicate the `dfm_trim()` argument, but also has a different name. (If we keep the argument in the network functions, we should at least change it to `min_count`.)

koheiw:
I am trying to avoid generating a dense matrix in `as.network.fcm`, which turned out to be not easy...

`textplot_network`'s `min_freq` does not trim the matrix, but fill cells lower than that threshold with zeros to make network clusters distinct. Its behavior is more like flooring instead of trimming, so it might be good to have `dfm_floor` and `fcm_floor`.

kbenoit:
Right. dfm_trim removes the features, you want to keep them but with zero. That will not take any space if the matrix is sparse, but will if dense.

It’s more like dfm_weight where features are weighted zero subject to a threshold. We could adapt that function to do what is needed.

On 6 Dec 2017, at 10:09, Kohei Watanabe <notifications@github.com<mailto:notifications@github.com>> wrote:


I am trying to avoid generating a dense matrix in as.network.fcm, which turned out to be not easy...

textplot_network's min_freq does not trim the matrix, but fill cells lower than that threshold with zeros to make network clusters distinct. Its behavior is more like flooring instead of trimming, so it might be good to have dfm_floor and fcm_floor.

PR #1137, opened 2017-12-13
Issue 1133

koheiw:
It is almost done, but there are two issues:

1. Check raises error for `nfeature` (should be easy but I could not how to suppress)
2. `head.dfm` breaks if I rename `nfeature` (although I don't think we need this argument)

https://github.com/kbenoit/quanteda/blob/0cc1fc0ce72057c4ca8d8dc69c1dbae684b0380b/R/dfm-print.R#L123-L128


kbenoit:
OK, thanks. I'll fix the nfeat/nfeature issue.

PR #1141, opened 2017-12-14
Add as.keras.tokens()

koheiw:


kbenoit:
I think a more natural method would be to add a new output type for `convert()`, since the input to the keras system is more logically a dfm. We should find someone who uses this system and ask them.

koheiw:
I am no expert but it seems that orders of words should be preserved for Keras. This is why I did not extend `convert()`.

kbenoit:
For any embedding or context-dependent model, that would be absolutely right!

FarzanT:
I can confirm that `keras::texts_to_sequences()` does indeed preserve the word order, and basically replaces the word with its _descending frequency index_ if it is in its internal dictionary. Maybe the quanteda tokenizer can also take in a dictionary and replace the words with their `synonyms` (indices in this case)? Maybe there already is such a function, but unfortunately, I'm not aware of that.

koheiw:
@FarzanT thank you for the input. I think you are right that the indices are frequency ranks of the tokens (1 is the most frequent). Do you know if it makes any difference if we use random integers?

FarzanT:
@koheiw For the models I have seen, it does not really matter; what matters is that the word order is kept after translation of tokens to indices. It'd be also really important if some sort of a mapping was stored somewhere, so that new tokenized sequences can be translated using a unified vocabulary (e.g. only test data tokens that are seen in the training tokens are kept, and are translated to indices of the __training vocabulary__).

koheiw:
@kbenoit We could add types argument that takes vocabulary from other tokens objects so that the same word get the same index.

spotcapkonradsemsch:
Is there eventually any recommended solution for converting a dfm into a keras object after applying texts_to_sequences? I checked the discussion here and also https://github.com/quanteda/quanteda/issues/1145 but in the end I'm not sure what the recommended approach is.

konradsemsch:
Is there eventually any recommended solution for converting a dfm into a keras object after applying texts_to_sequences? I checked the discussion here and also #1145 but in the end I'm not sure what the recommended approach is.

kbenoit:
OK...

#### for a dfm

This will work, although I am not sure how robust it is to ties in the frequencies or how **keras** handles this:

```r
library("keras")
txt <- c(d1 = "a a b b b c", d2 = "a b c c c c")
tokenizer <- text_tokenizer(num_words = 4) %>%
    fit_text_tokenizer(txt)

dfm_to_kerasmatrix <- function(x) {
    mat <- dfm_sort(x, margin = "features") %>%
        as.matrix()
    dimnames(mat) <- NULL
    cbind(rep(0, nrow(mat)), mat)
}
    
identical(
    texts_to_matrix(tokenizer, txt, mode = "count"),
    dfm_to_kerasmatrix(dfm(txt))
)
## TRUE
```

#### for tokens

This would involve us remapping the integer values to our types vector internally for tokens, so that their order corresponds to decreasing total frequency. Again, we would need to test for ties, but this should do it:

```r
tokens_to_kerassequences <- function(x) {
    x <- unclass(x)
    attributes(x) <- NULL
    x_freqs <- table(unlist(unclass(x)))
    x_freqs_ordered <- temp[order(temp, decreasing = TRUE)]
    lapply(x, function(y) as.integer(names(x_freqs_ordered))[y])
}

identical(
    texts_to_sequences(tokenizer, txt),
    tokens_to_kerassequences(tokens(txt))
)
## TRUE
```

Please test those behaviours and let us know if they match your needs, and we can think about how to (re)name them and integrate them into **quanteda**.

PR #1147, opened 2017-12-15
Quickstart ja

koheiw:
This is the final version.


koheiw:
Be careful that the I changed the location of Japanese quick start. Better to leave this unmarged for a few days.

PR #1152, opened 2017-12-18
Issue 1150

koheiw:
Not only address 1150, but also make fcm more robust


kbenoit:
Oops @koheiw looks like we crossed wires. I fixed this in c53a9c8. Can you take the other bits (`fcm()` improvements) and make those only part of this new PR?

PR #1163, opened 2017-12-21
Issue 1153

koheiw:


kbenoit:
I took off the `@noRd` tag and let it create a help page, but kept this internal only. 

I tweaked the documentation a bit, and imported the parameters - see `?textstat.subset` and `?dfm_subset` and look at the roxygen header. It's smart enough to import only the parameters not already defined in the importing function. This is a nice way to ensure consistency across similar functions.

But what I did on `subset` and `select` is not right, I just realised, because we are not selecting documents at all, but rather features. Overall, this function is a bit of an oddball `subset()` method that combines the logical expressions for the data.frame column subsetting and selection, which are not needed here, versus the pattern matching of one of our `select_*()` methods.

Maybe we should call this internally, `textstat_select()` instead?

koheiw:
`textstat_select()` is the original name, so it is fine for me. We could keep it internally, but don't you think it is really handly to select rows of large textstat table by `glob` patters?

kbenoit:
Yes it's handy, so let's rename it to `textstat_select()`. 

Before we defined a textstat class object, it was not consistent, but we have that now.

koheiw:
Renamed by it complains that 
```
Error in textstat_select(keyness, "america*") : 
  could not find function "textstat_select"
```
What is missing? `quanteda:::textstat_select(keyness, 'america*')` works.

kbenoit:
Just fixed it - there was some leftover stuff like the `@method` statement that needed removing.

kbenoit:
See my notes in 77f1ee4

koheiw:
I agree that `textstat_select()` is misleading, but how can users find this function? `textstat_*()` should actually be `textstat_calculate_*()`. If this is too long, we could wrap them by `textstat(, measure = c('keyness' 'collocations'))` just like old `textmodel()`.... :(

kbenoit:
I’ve been thinking about this, and we should not confuse the syntax by adding `textstat_select().`. Here’s a better idea: we write a method for `dplyr::filter()`. This is flexible - see the generic in https://github.com/tidyverse/dplyr/blob/master/R/manip.r - so could add our arguments. 

Using the `filter()` and `filter_()` syntax we can also get the NSE if we need it. (My not.)

This would work nicely for all of our textstat functions that return data.frames.

koheiw:
If we move functions to other package's namespace, consistency problems are solved easily. This is one solution, but it just transfers the cost to users by forcing them to learn **dplyr** syntax (they probably should).

kbenoit:
Not the **dplyr** syntax, just the filter command, and it will have an set of formals identical to our existing draft `textstat_select`. To include its generic, we simply need to 

```r
#' export
dplyr::filter

#' @export
filter.textstat <- function(.data, 
                            pattern = NULL, 
                            selection = c("keep", "remove"), 
                            valuetype = c("glob", "regex", "fixed"),
                            case_insensitive = TRUE, ...) {
    [code]
}

**Comment by koheiw** [comment]
The best way to avoid contaminating the namespace is not naming a all:
```r
col <- textstat_collocation(something)
col[pattern = 'America*', case_insensitive = FALSE]
```
Since we have disabled `drop` argument,
```r
col[,,'America*']
```
works.

This is similar to **data.table**'s design.

kbenoit:
If a user wants to use **data.table** indexing, let them create a data.table from the return object.  I don't want to confuse things by overloading a fundamental base operator on a specially tagged data.frame.

I think the `filter` approach is the way to go.

koheiw:
Other than `textstat`, on which **quanteda** objects could we use `filter()`?

kbenoit:
None that I can think of, but that doesn't really matter. `filter.data.frame` will already work on the textstat objects fine, but by extending to the textstat classed data.frames, we can make it possible to select on the features in the convenient way that you implemented in what is now named `textstat_select()`.

koheiw:
We should decide what to do with `textstat` subsetting later, because Japanese examples do not work without this branch to be merged.

PR #1165, opened 2017-12-23
Remove collocations_dev

koheiw:
I separated `collocations_dev` to a new branch, and removed its code. C and C++ can co-exist, but C++ compiler raises warnings on the C code. We can suppress that warning unless we set complier flags, but we cannot do this for the CRAN check.


kbenoit:
Thanks. I think I will just separate `collocationsdev` into a separate GitHub-based package that Depends on *quatenda*.

Issue #1167, opened 2017-12-24
textplot_network font family


amatsuo:
As all of us know well, we need to assign font family in order to show the plot of non-ascii characters appropriately (e.g. Chinese and Japanese). At the moment `textplot_network` does not have that option, and I need it for Japanese text demo. 

I am working on it and I will send a PR soon.

PR #1168, opened 2017-12-24
add option family to textplot_network

amatsuo:
for #1167


koheiw:
Am I right that this is primarily for Mac, because I had not problem with CJ characters on Windows or Linux? How about naming the argument `vertex_labelfont`? I can see you calling geom functions with or without `family`, but it should be enough to override NULL by "" (empty string) to use default font.

amatsuo:
I think everything @koheiw pointed out has been addressed.

Issue #1169, opened 2017-12-24
textstat_simil() fails when a dfm has padding and selection is given


koheiw:
```r

toks <- tokens(data_char_ukimmig2010, remove_punct = TRUE)
toks <- tokens_remove(toks, stopwords(), padding = TRUE)
mt <- dfm(toks)
topfeatures(mt)
#      immigration     british      people      asylum     britain          uk      system  population     country 
# 2596          66          37          35          29          28          27          27          21          20
textstat_simil(mt, selection = c('immigration'), margin = 'feature')
# Error in temp[names, , drop = FALSE] : subscript out of bounds

```

koheiw:
For future reference. There is an ugly inconsistency in the **Matrix** package that caused this problem.
```r
mt <- matrix(1:9, nrow = 3)
colnames(mt) <- c('', 'b', 'c')
mt
#        b c
# [1,] 1 4 7
# [2,] 2 5 8
# [3,] 3 6 9

as(mt, 'dgCMatrix')[,'']
# [1] 1 2 3
as(mt, 'denseMatrix')[,'']
# Error in as(x, "matrix")[, j, drop = drop] : subscript out of bounds
```

PR #1170, opened 2017-12-25
Make dist and simil robust against seletion with padding

koheiw:
Address #1169

### DIFF
--- a/R/textstat_dist.R
+++ b/R/textstat_dist.R
@@ -118,11 +118,10 @@ textstat_dist.dfm <- function(x, selection = NULL,
         stop(method, " is not implemented; consider trying proxy::dist().")
     }
     
-    if (!is.null(selection)) {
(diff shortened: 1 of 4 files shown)


koheiw:
There is an ugly inconsistency in the **Matrix** package that caused #1169:
```r
mt <- matrix(1:9, nrow = 3)
colnames(mt) <- c('', 'b', 'c')
mt
#        b c
# [1,] 1 4 7
# [2,] 2 5 8
# [3,] 3 6 9

as(mt, 'dgCMatrix')[,'']
# [1] 1 2 3
as(mt, 'denseMatrix')[,'']
# Error in as(x, "matrix")[, j, drop = drop] : subscript out of bounds
```
This should be solved but where should be report this? Here?
https://r-forge.r-project.org/tracker/?group_id=61

Issue #529, opened 2017-01-29
Wishlist: Better dplyr function support


methodds:
Hi,

its so awesome how much user friendlier quanteda became in the last year and I could imagine this to improve even more by adding better dplyr support to quanteda. Of course, corpora are not data frames, but often have docvars and imagine a world where you could do something like this:

```
library(quanteda)
corpus <- data_corpus_inaugural

dfm_grouped <- corpus %>% 
 mutate(war = ifelse(Year<1945, "pre-war", "post-war")) %>%
 group_by(war) %>% dfm(remove=stopwords('english'))
```
This could work if a corpus (or other quanteda objects) would recognize a dplyr call and pass ``docvars()`` accordingly etc. What do you think about this?

kbenoit:
Following on the discussion at #1171, here's a possibility for making the popular **dplyr** operations on corpus object docvars available from **quanteda**: We create a companion package called **quanteda.dplyr** that imports **dplyr**, exports its generics, and provides methods for our corpus objects.

These would mainly (or only) perform manipulation on document variables, or if grouping, group document variables using a limited subset of functions while concatenating texts in the group.

Currently we have `corpus_sample()` and `corpus_subset()`, for instance. An alternative would be `sample` and `filter` from **dplyr**, for sampling and subsetting documents. Other extensions would do the same directly on dovcars, but handling the documents correctly. **dplyr** candidates include:

* `filter`
* `group_by`
* `*_join`, `join`
* `mutate`
* `recode`
* `rename`
* `select`
* `transmute`

@hadley if you have time to take a brief look at #1171 your input on whether this is the best way to extend `dplyr::filter()` would be most appreciated!

kbenoit:
This issue was moved to quanteda/quanteda.tidy#1

PR #1171, opened 2017-12-27
replace textstat_select with filter method

kbenoit:
Extends `dplyr::filter()` to data.frame-type returns from `textstat_*()` functions, with added arguments for `pattern`, `valuetype`, and `case_insensitive`.

Removes the `selection` argument, since it is inconsistent with `filter()` and since equivalent functionality can be returned using the negation operator or negative conditions on a regular expression.


kbenoit:
See also #529.

koheiw:
I think the best way is using `subset` as it was originally. It should have `select` and `subset`, but we just disallow using those and `patterns` at the same time.

R/textstat-methods.R
```diff
@@ -1,71 +1,53 @@
...
-    id <- unlist(regex2id(pattern, x[[1]], valuetype, case_insensitive))
-    if (selection == 'keep') {
-        x <- x[id,]
-    } else {
-        x <- x[id * -1,]
+    # call dplyr filter, if ... arguments are supplied
+    if (length(list(substitute(...)))) {
+        if (!requireNamespace("dplyr", quietly = TRUE)) 
+            stop("You must install the dplyr package to use filter.textstat")
+        .data <- as.data.frame(dplyr::filter(dplyr::tbl_df(.data), ...))
```
hadley: Why do you coerce back and forth to a tibble here?

R/textstat-methods.R
```diff
@@ -1,71 +1,53 @@
...
-    attrs <- attributes(x)
+    attrs <- attributes(.data)
     
-    id <- unlist(regex2id(pattern, x[[1]], valuetype, case_insensitive))
-    if (selection == 'keep') {
-        x <- x[id,]
-    } else {
-        x <- x[id * -1,]
+    # call dplyr filter, if ... arguments are supplied
+    if (length(list(substitute(...)))) {
```
hadley: I think this would be cleaner with a helper:

```R
ndots <- function(...) nargs()
```

R/textstat-methods.R
```diff
@@ -1,71 +1,53 @@
...
+        if (!requireNamespace("dplyr", quietly = TRUE)) 
+            stop("You must install the dplyr package to use filter.textstat")
+        .data <- as.data.frame(dplyr::filter(dplyr::tbl_df(.data), ...))
     }
-    class(x) <- attrs$class
-    return(x)
+
+    # select on features if specified
+    if (!missing(pattern)) {
```
hadley: Why are you not checking for `NULL` here?

R/textstat-methods.R
kbenoit: For some reason I had trouble accessing `dplyr::filter.data.frame`. But this solves it:
```r
dplyr:::filter.data.frame(.data, ...)
```

kbenoit:
Thanks @hadley !

koheiw:
Since `dplyr` is very popular it makes sense to add `filter()`, but conservatives like me prefer `subset()`. Why don't we have both?

kbenoit:
Moved to [**quanteda.tidy**](https://github.com/quanteda/quanteda.tidy).

The `subset.data.frame()` function is supposed to do NSE on variable names, but here, `pattern`, `valuetype`, and `case_insensitive` are not variable names. What if one of them were? 

Although one could say that the same problem is true for `filter.textstat()`.

PR #1180, opened 2018-01-04
Issue 1179

koheiw:
Adds check for available fonts. Works on Windows and Linux, but not sure about Mac. @amatsuo , can you check on you system?


koheiw:
It is ready to be merged if it works on Mac. I cannot not improve the test coverage because the new code is platform dependent.

Issue #1007, opened 2017-10-05
summary() should not print anything


koheiw:
Just like the older version of `head.dfm()`, `summary.textmodel_wordfish_fitted()` prints model parameters. It should only return values just like `summary.lm()`.
```r
ieWF <- dfm(data_corpus_irishbudget2010, removePunct = TRUE) %>%
    textmodel_wordfish(dir = c(6,5))
textplot_scale1d(ieWF)
out <- summary(ieWF)
# Call:
#   textmodel_wordfish.dfm(x = ., dir = c(6, 5))
# 
# Estimated document positions:
#   theta         SE       lower       upper
# 2010_BUDGET_01_Brian_Lenihan_FF        1.8209535 0.02032345  1.78111954  1.86078748
# 2010_BUDGET_02_Richard_Bruton_FG      -0.5932779 0.02818836 -0.64852708 -0.53802872
# 2010_BUDGET_03_Joan_Burton_LAB        -1.1136753 0.01540256 -1.14386429 -1.08348625
# 2010_BUDGET_04_Arthur_Morgan_SF       -0.1219325 0.02846319 -0.17772032 -0.06614462
# 2010_BUDGET_05_Brian_Cowen_FF          1.7724224 0.02364097  1.72608615  1.81875874
# 2010_BUDGET_06_Enda_Kenny_FG          -0.7145784 0.02650254 -0.76652337 -0.66263342
# 2010_BUDGET_07_Kieran_ODonnell_FG     -0.4844821 0.04171476 -0.56624299 -0.40272114
# 2010_BUDGET_08_Eamon_Gilmore_LAB      -0.5616713 0.02967351 -0.61983141 -0.50351124
# 2010_BUDGET_09_Michael_Higgins_LAB    -0.9703106 0.03850541 -1.04578121 -0.89484000
# 2010_BUDGET_10_Ruairi_Quinn_LAB       -0.9589229 0.03892373 -1.03521343 -0.88263242
# 2010_BUDGET_11_John_Gormley_Green      1.1807220 0.07221463  1.03918133  1.32226270
# 2010_BUDGET_12_Eamon_Ryan_Green        0.1866457 0.06294119  0.06328093  0.31001039
# 2010_BUDGET_13_Ciaran_Cuffe_Green      0.7421895 0.07245436  0.60017891  0.88420001
# 2010_BUDGET_14_Caoimhghin_OCaolain_SF -0.1840821 0.03666256 -0.25594076 -0.11222351
```
[...]

PR #1190, opened 2018-01-10
Issue 1007

kbenoit:
Redoes the `textmodel_*()` functions to make their object output and behaviour consistent with the `lm/glm` family of models, including in particular the `predict`, `summary`, `print`, and `coef` methods.

Still needs to be done for `textmodel_affinity()`.

Solves #1007.


kbenoit:
Still need to:
- [ ] fix naming and transition to S3 for `textmodel_ca()`, redo the code in `textplot_scale1d()` for these changes
- [ ] fix naming and transition to S3 for `textmodel_lsa()`
- [ ] fix naming and transition to S3 for `textmodel_affinity()`, as well as additional methods

koheiw:
The first two look strange to me. `predict(ws)` should return a vector of fitted values (without the `predict.textmodel_wordscores` class). It is not only to make it similar to `lm` but to allow users to skip calculation of se. People working on large dataset might want fitted values (I do usually).
```
> predict(ws)
             fit         lwr         upr
R1 -1.317931e+00 -1.33106234 -1.30480034
R2 -7.395598e-01 -0.76196925 -0.71715035
R3 -8.673617e-18 -0.02352986  0.02352986
R4  7.395598e-01  0.71715035  0.76196925
R5  1.317931e+00  1.30480034  1.33106234
V1 -4.480591e-01 -0.47137808 -0.42474008
attr(,"class")
[1] "predict.textmodel_wordscores" "matrix"                      

> predict(ws, se.fit = TRUE)
$fit
             fit         lwr         upr
R1 -1.317931e+00 -1.33106234 -1.30480034
R2 -7.395598e-01 -0.76196925 -0.71715035
R3 -8.673617e-18 -0.02352986  0.02352986
R4  7.395598e-01  0.71715035  0.76196925
R5  1.317931e+00  1.30480034  1.33106234
V1 -4.480591e-01 -0.47137808 -0.42474008

$se.fit
[1] 0.006699613 0.011433605 0.012005250 0.011433605 0.006699613 0.011897667

attr(,"class")
[1] "predict.textmodel_wordscores" "list"                        

> predict(ws, se.fit = TRUE, interval = "confidence")
$fit
             fit         lwr         upr
R1 -1.317931e+00 -1.33106234 -1.30480034
R2 -7.395598e-01 -0.76196925 -0.71715035
R3 -8.673617e-18 -0.02352986  0.02352986
R4  7.395598e-01  0.71715035  0.76196925
R5  1.317931e+00  1.30480034  1.33106234
V1 -4.480591e-01 -0.47137808 -0.42474008

$se.fit
[1] 0.006699613 0.011433605 0.012005250 0.011433605 0.006699613 0.011897667

attr(,"class")
[1] "predict.textmodel_wordscores" "list" 
```

kbenoit:
The predict method does return just a named vector of fitted values, when `se.fit = FALSE, interval = "none"`. I changed the default for interval but we could easily put the default back to `"none"`, which is how `predict.lm()` works.

Add the attribute is the dumb S3 way to let other methods know that it's a predicted object. Mainly, we need this for `textplot_scale1d()`. Might be able to figure out another way to trigger that function in an easy to use way, or we could define a `print.predict.textmodel` method that removes the attribute so that the user is not exposed to that ugliness.

kbenoit:
Here we have a classic legacy dilemma with the output of `predict()`: if we match the existing (e.g.) `predict.lm()` behaviour, then we have an indeterminate output object type (vector, list, matrix, etc). 

The only way to maintain consistency would be always to return a list, and always to have fit be a matrix, even if it had a single column, but then that does not provide the simple vector of predicted values that you said you (and many other users) prefer esp. for large datasets.

So my vote is to mimic the existing idiom.

koheiw:
As long as we use generic methods like `predict`, we cannot keep shapes of output objects the same. I am pleased to know that we can still get a simple vector, but not sure if we should change the default value in the generic method.

kbenoit:
Fine if you want to revert the parameter default to `interval = c("none", "confidence")` but you will need to fix what it breaks in the tests/examples then, incl for `textplot_scale1d()` (but this will be simple). Users who want SEs for that plot will need to set either `se.fit = TRUE` or `interval = "confidence"`.

koheiw:
textmodel_wordscore

- [ ] `summary`
- [ ] `coef`
- [ ] `coefficients`
- [ ] `predict`
- [ ] `print.predict`

textmodel_wordfish

- [ ] `summary`
- [ ] `coef`
- [ ] `coefficients`
- [ ] `predict`
- [ ] `print.predict`

textmodel_nb

- [ ] `summary`
- [ ] `coef`
- [ ] `coefficients`
- [ ] `predict`
- [ ] `print.predict`

koheiw:
All the textmodels are S3 with new class labels. `textmodel_affinity` is almost done, but there is no `summary.textmodel_affinity()`. `print.textmodel_affinity()` also needs extra info.

Issue #1183, opened 2018-01-09
Wishlist: Kosuke Imai's QSS (Chap. 5)


A-Tago:
Dear Ken and Quanteda team,
I am going to teach a method class by using Kosuke Imai's _Quantitative Social Science_ and I wonder if you could formulate the Quanteda code of the Chapter 5 (section 1: titled "Textual data"). I am sure I could do with lots of time, but I think you guys can generate it with lesser energy and time. Currently, tm package is used but I think it would be nice to teach by using Quanteda.
Atsushi

stefan-mueller:
Good idea! I can look into this over the weekend and draft a new vignette which could be added to the "Replication" section on http://docs.quanteda.io. As the texts used in the book are available in the [package](http://qss.princeton.press/student-resources-for-quantitative-social-science/r-package/) accompanying the book, data availability shouldn't be an issue.

A-Tago:
Hello Stefan,
Many thanks!  I will use this in Waseda univ. where I am going to move in April. Hope students will get interested in text analysis and use Quanteda.  I bet they will. 
Atsushi 

iPhoneから送信

2018/01/09 21:44、Stefan Müller <notifications@github.com>のメール:

stefan-mueller:
Below you find the code to run the chapter with **quanteda**. I could reproduce all commands – the only difference is that `tfidf` seems to be calculated differently in **tm** than in **quanteda**. I have used `tfidf(x, scheme_tf = "prop")`, but the results are still not the exact same. I am not sure which command matches the default calculation of `tm::weightTfIdf()`. 

Other than that, the text analysis part in the book is quite limited, so I am not sure whether a separate vignette is necessary (but I could create one if you think it would be useful).

@A-Tago, if you use this for teaching purposes, you could ask your students to use `textmodel_nb()` to predict the authorship of the disputed papers and compare the results with the linear model.

```r
## Replication of Kosuke Imai's Quantitative Social Science, Chapter 5 (Section 1)
# https://kosukeimai.github.io/qss-package/articles/discovery.html

# library("devtools")
# install_github("kosukeimai/qss-package", build_vignettes = TRUE)

library(quanteda)
library(readtext)
library(qss)

# use readtext package to import all documents as a dataframe
corpus_texts <- readtext(system.file("extdata/federalist/", package = "qss"))

# create docvar with number of paper
corpus_texts$paper_number <- paste("No.", 1:nrow(corpus_texts), sep = " ")

[...]

PR #1197, opened 2018-01-12
Add vignette that reproduces Chapter 5.1  "Quantitative Social Science"

stefan-mueller:
See #1183.


kbenoit:
For the assignment of authors in the prediction example, it would nice to add in a docvar for the Author, and use that instead.

stefan-mueller:
See my changes above. Is that what you preferred? Happy to make more adjustments if required.

stefan-mueller:
Can we merge this pull request or are additional changes required, @kbenoit?

kbenoit:
I finally remembered to bring home my copy of _QSS_ so now will able to compare the vignette with the book. Should be done soon. The question is whether to translate the literal **tm** code as in the book into **quanteda**, or also to improve the code a bit further. (Probably a mixture of the two.)

Issue #1200, opened 2018-01-18
Find a workaround for the UBSAN warnings


kbenoit:
Each time we (re)submit to CRAN, we get flagged on the UBSAN warnings ([clang-UBSAN](https://www.stats.ox.ac.uk/pub/bdr/memtests/clang-UBSAN/quanteda/) and [gcc-UBSAN](https://www.stats.ox.ac.uk/pub/bdr/memtests/gcc-UBSAN/quanteda/)) arising from the TBB calls through **RcppParallel**.  These are happening when the examples and tests are calling functions using multi-threading, when the threads options is > 1, e.g.
```r
> quanteda_options("threads")
[1] 7
```

This is a [known issue in **RcppParallel**](https://github.com/RcppCore/RcppParallel/issues/36#issuecomment-358445542), but no immediate fix seems to be in sight.

To test this setup, we need to create a branch with some custom Makevars flags. The general method is described [here](https://cran.r-project.org/web/checks/check_issue_kinds.html) and in a [helpful blog post here](https://francoismichonneau.net/2014/03/how-to-do-ubsan-tests-r-package/).

Possible solutions:
1. Solve the underlying problem.  (if only)
2. Switch to 1 thread by default, for the package (if this in fact makes the warnings disappear - we would need to test this first of course). Not preferred, however, since most users will not reset the threads and hence will unknowingly deny themselves the benefits of multi-threading.
3. Switch to single-threaded examples in the examples sections for all multi-threaded functions, and for the tests. We could put the former in the examples sections, and the latter in `testthat.R`.
4. Adjust for CRAN, meaning we would need to detect CRAN, and then: 
    a. Skip the multi-threaded tests on CRAN; or change to a single thread on CRAN;
    b. Figure out a similar solution for the examples, for CRAN only.

HervePerdry:
I admit solution 4 comes to mind -- but it sounds a little VW-ish, doesn't it? :)

amatsuo:
Regarding 4. The detection of CRAN seems to be very simple:
https://github.com/r-lib/testthat/blob/master/R/skip.R#L103-L109

kbenoit:
@amatsuo as far as I understand, that only applies to tests, not examples.

HervePerdry:
[...]

Issue #1003, opened 2017-10-02
Need to fix UBSAN issues


kbenoit:
Got this email today from the CRAN team:

> Also, please take a look at the clang-UBSAN issues reported at
> https://www.stats.ox.ac.uk/pub/bdr/memtests/clang-UBSAN/quanteda/
> and try to fix.
> 
> Best,
> Uwe Ligges
> (CRAN team)

This seems to be triggered by the vignette only. We need either to explain this clearly as being not our fault, but rather that of **Rcpp**, or diagnose and fix it.

For latter, putting it to the rOpenSci Slack team might be a good start. I seem to remember some chatter about related issues there.

kbenoit:
See more at https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Using-Undefined-Behaviour-Sanitizer

I got this message today:

Begin forwarded message:

> From: Prof Brian Ripley <ripley@stats.ox.ac.uk>
> Subject: Re: CRAN package quanteda
> Date: 5 October 2017 at 10:24:30 BST
> To: "Benoit,KR" <K.R.Benoit@lse.ac.uk>, "CRAN@r-project.org" <CRAN@r-project.org>
> 

> > On 05/10/2017 09:56, Benoit,KR wrote:
> > Dear Uwe and the CRAN team,
> > We fixed the R version dependency already, and are working on the UBSan issues. It might be of interest to the checking system however to know that it was not a change in our package that broke the old-release versions, but rather changes to packages that we import. I’m not sure what is the best system to ensure that we depend on an R version at least as recent as the latest version dependency of all of the packages we import.
> > On the UBSan issues, what’s the best way to detect those? It looks like our package check passes, but that the vignette is somehow triggering issues.
> 

> 
> You study the manual, as the README in that area says!
> 
> But the UBSAN errors are from C++ code in package RcppParallel whose headers your code imports and whose checks show the same issue.  They occur in your examples, not just the vignette.  This needs to be solved in RcppParallel.

kbenoit:
Keep our eyes out for a solution in **Rcpp**.

koheiw:
Here is a nice page on how to set up Linux for the UBSAN test
https://francoismichonneau.net/2014/03/how-to-do-ubsan-tests-r-package/

Test environment details:
https://www.stats.ox.ac.uk/pub/bdr/memtests/clang-UBSAN/README.txt
[...]

PR #1201, opened 2018-01-18
Check ubsan

koheiw:
I was investigating #1003 and #1200. `rhub::check_with_sanitizers()` indicated the bit shift in our hash function is causing an error. See line 6221 (or search for 'runtime error') in https://builder.r-hub.io/status/original/quanteda_1.0.0.tar.gz-9672f200479b4108a6308a4d54edcc09. The log looks very different but the location of the error is very close to the TBB error in the [CRAN UBSAN check](https://www.stats.ox.ac.uk/pub/bdr/memtests/clang-UBSAN/quanteda/quanteda-Ex.Rout). 

After modifying the hash function, the runtime error has disappeared from check. I hope the same thing to happen in CRAN check.

### DIFF
--- a/src/quanteda.h
+++ b/src/quanteda.h
@@ -130,29 +130,12 @@ namespace quanteda{
             unsigned int add = 0;
             unsigned int seed = 0;
             for (std::size_t i = 0; i < vec.size(); i++) {
-                add = vec[i] << (8 * i);
-                if (seed <= UINT_MAX - add) { // check if addition will overflow seed
-                    seed += add;
-                } else {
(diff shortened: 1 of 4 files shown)


koheiw:
I tried the check as in https://francoismichonneau.net/2014/03/how-to-do-ubsan-tests-r-package/ but it did not detect the tbb error, so the only way for us to check is submitting to CRAN.

kbenoit:
The instructions I linked to in #1200 describe how to configure a system for the same UBSAN tests as used by CRAN.

Issue #1211, opened 2018-02-05
Fix legends in textplot_keyness() when there are more than two documents


koheiw:
The label for reference documents becomes the name of the first documents in the reference group like `text12392`.

kbenoit:
Can you fix this separately from the broader issues in #1213 ?

kbenoit:
See #1233.

PR #1213, opened 2018-02-05
Reimplement textplot_keyness to make it more consistent with other plots

koheiw:
For #1211, but reimplemented the function almost entirely. There are two behavioral changes:
- It plots the same number of features regardless of the value of `show_reference`
- Its plot is no longer a barchart as it is not a good visualization of keyness 

As a result, the plot looks more consistent with other textplots.


kbenoit:
How about adding a new parameter `show_bars = TRUE` that plots the older bars, and produces the plot you have shown when `FALSE`?

To improve the new plot, I would suggest removing the horizontal bars, and trying to plot the colour of each word on a scale from dark blue to dark red (like the bars) but shaded a darkness proportional to the by the relative x-axis value.

koheiw:
I don't think bars are appropriate because chi2 and pmi are not volume. But the plot could be fancier by default. I should make a theme for all the textplots, so that all will look like outputs of the same package.

kbenoit:
I have heard a lot of praise for the bars... including from my own inner graph child!

Like the idea of a "quanteda theme" but it should be able to be replaced or modified easily.

Issue #1217, opened 2018-02-07
Fail gracefully in print.summary.corpus when ndoc_original is not available


chmue:
Subsetting or selecting a `summary.corpus` object removes the attribute `ndoc_original`. The print function `print.summary.corpus` fails with an error when the attribute is not available.

#### Minimal example

```r
irishbudget2010 <- summary(data_corpus_irishbudget2010)
irishbudget2010[, c("Text", "Types", "Tokens")]
```
```
Error in if (ndoc_original > 1) "s" else "" : argument is of length zero

Enter a frame number, or 0 to exit   

1: function (x, ...) 
UseMethod("print")(x)
2: print.summary.corpus(x)
3: cat("Corpus consisting of ", ndoc_original, " document", if (ndoc_original > 1) "s" else "

Selection: 0
```

#### Suggested fix

Use a check whether the returned attribute is NULL in [this line](https://github.com/quanteda/quanteda/blob/209430fbfddeb7cf1f4ae0638e088c452d291bbb/R/corpus-methods-base.R#L101) and replace with a safe default.

```r
print.summary.corpus <- function(x, ...) {
    ndoc_original <- attr(x, "ndoc_original")
    if (is.null(ndoc_original)) ndoc_original <- 0
    cat("Corpus consisting of ", ndoc_original, " document", 
        if (ndoc_original > 1) "s" else "", 
        attr(x, "showing"), ":\n\n", sep = "")
    print.data.frame(x, row.names = FALSE)
    cat(attr(x, "meta"), "\n")
}
```

**Edit:** The suggested fix might not be the best way to handle this situation. Maybe removing the class `summary.corpus` and using `print.data.frame` makes more sense.

kbenoit:
Thanks @chmue. @koheiw, I'm struggling now to remember, why we have `ndoc_original` in the first place...  Seems like `corpus_subset()` should have set that value, so something is probably missing.

chmue:
Thank you for your work on this!

PR #1220, opened 2018-02-08
Issue 1217

kbenoit:
Solves #1217.

- Removes rownames from the  `summary.corpus` object
- drops special attributes from the `summary.corpus` object when that object is subset using `[`
- add tests


koheiw:
I modernized the `summary.corpus` method. However, I find `n` argument is totally unnecessarily, because if users what to get summaries of first n documents, they should subset the corpus using `head()` beforehand. 

Regarding the column names, I am not sure if they should be capitalized like (Text, Types, Tokens Sentences, because we keep all the labels in textstats output lowercased. Nonetheless, it is OK to capitalize them in `print.summary.corpus`

I also think that we should have summary methods for `tokens` and `dfm`. These are very easy to make.

Issue #1212, opened 2018-02-05
Return tidy data for as.data.frame.dfm?


chmue:
Currently, `as.data.frame.dfm` returns a data.frame with rownames, which is not tidy.

```r
idfm <- dfm(data_corpus_inaugural)
idfm <- dfm_trim(idfm, min_count = 1200)  # make it fit on screen
as.data.frame(idfm)
```

```
                 of the and  to that   ,   .   a  in  it  be is our we
1789-Washington  71 116  48  48   18  70  23  14  31  11  23  7   1  1
1793-Washington  11  13   2   5    1   5   4   0   3   2   2  0   0  0
1797-Adams      140 163 130  72   22 201  33  51  47  34  31 14   6  3
 [ reached getOption("max.print") -- omitted 55 rows ]
```

Ideally for me, it would return something like:

```r
tibble::rownames_to_column(as.data.frame(idfm), "docid")
```
```
             docid  of the and  to that   ,   .   a  in  it  be is our we
1  1789-Washington  71 116  48  48   18  70  23  14  31  11  23  7   1  1
2  1793-Washington  11  13   2   5    1   5   4   0   3   2   2  0   0  0
3       1797-Adams 140 163 130  72   22 201  33  51  47  34  31 14   6  3
 [ reached getOption("max.print") -- omitted 55 rows ]
```

I can work on a fix if you want to implement that.

kbenoit:
Valid point - currently 
```r
identical(
    as.data.frame(data_dfm_lbgexample),
    as.data.frame(as.matrix(data_dfm_lbgexample))
)
```
But we could make the first column named `document` rather than put this into the row names, for the data.frame only. It depends on what reverse imports we would break - but this is testable with `devtools::revdep_check()`.

@koheiw what do you think?

I would not want to make the return a tibble object since we are [considering a **quanteda.dplyr**](https://github.com/quanteda/quanteda/issues/529#issuecomment-354131987) extension package that would allow **dplyr**/tidyverse functions and return tibbles etc when appropriate.

koheiw:
The coercion method is not really our creation:
```r
as.data.frame.dfm <- function(x, row.names = NULL, ...) {
    as.data.frame(as.matrix(x), row.names = row.names, ...)
}
```
So, if you do not want row names:
```r
df <- as.data.frame(idfm, row.names = FALSE)
df$docid <- docnames(idfm)
```
I also do not think that users should coerce a DFM to a data frame, because it will brow up their RAM.

[...]

PR #1236, opened 2018-02-15
Improve as.data.frame.dfm() method

kbenoit:
Makes default the creation of a column called "document" and drops the rownames.

Solves #1212.

From `devtools::revdep_check()`:
```r
Checked politeness    : 1 error  | 0 warnings | 0 notes
```

Needs to be investigated.

New behaviour:
```r
as.data.frame(data_dfm_lbgexample[, 1:15])
#   document A B  C  D  E  F   G   H   I   J   K   L   M   N   O
# 1       R1 2 3 10 22 45 78 115 146 158 146 115  78  45  22  10
# 2       R2 0 0  0  0  0  2   3  10  22  45  78 115 146 158 146
# 3       R3 0 0  0  0  0  0   0   0   0   0   2   3  10  22  45
# 4       R4 0 0  0  0  0  0   0   0   0   0   0   0   0   0   0
# 5       R5 0 0  0  0  0  0   0   0   0   0   0   0   0   0   0
# 6       V1 0 0  0  0  0  0   0   2   3  10  22  45  78 115 146
```

### DIFF
--- a/R/dfm-classes.R
+++ b/R/dfm-classes.R
@@ -121,8 +121,7 @@ setMethod("+", signature(e1 = "numeric", e2 = "dfm"),
 #' @method as.matrix dfm
 #' @examples
 #' # coercion to matrix
-#' mydfm <- dfm(data_corpus_inaugural)
-#' str(as.matrix(mydfm))
+#' as.matrix(data_dfm_lbgexample[, 1:10])
 #' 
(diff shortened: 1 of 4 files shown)


kbenoit:
**politeness** seems to work fine when I ran local tests on a clone. I think the revdep error came from some **spacyr** issues.

Issue #1237, opened 2018-02-16
allow sparsity reduction of tf-idf weighted dfm


hadgi:
Currently, it is possible to perform dfm_trim() on dfm object and then add tf-idf weights, however, when I want to add tf-idf weights first by using all terms for tf-idf calculations and only then reduce the sparsity of the matrix, I get `Error in docfreq.dfm(x) : this dfm has already been term weighted as:inverse10`.

kbenoit:
Please provide a reproducible example.

hadgi:
Sorry. Here is one: 

`dfm.object1 <- dfm(c("the quick brown fox jumps over the lazy dog", "the quick brown foxy ox jumps over the lazy god"))
dfm.object1
dfm.trimmed1 <- dfm_trim(dfm.object, min_docfreq = 0.9)
dfm.trimmed1
dfm.tfidf1 <- dfm_tfidf(dfm.trimmed)
dfm.tfidf1

dfm.object2 <- dfm(c("the quick brown fox jumps over the lazy dog", "the quick brown foxy ox jumps over the lazy god"))
dfm.object2
dfm.tfidf2 <- dfm_tfidf(dfm.object)
dfm.tfidf2
dfm.trimmed2 <- dfm_trim(dfm.tfidf, min_docfreq = 0.9)
dfm.trimmed2`

kbenoit:
The short answer is: It's partly a bug (which I will address soon) and partly a feature, designed to prevent users from applying an inappropriate definition of the trim thresholds in `dfm_trim()` to a weighted dfm.

The simple reproducible error is:
```r
data_dfm_lbgexample %>%
    dfm_tfidf() %>%
    dfm_trim()
## Error in docfreq.dfm(x) : 
##     this dfm has already been term weighted as:inverse10 
```
More soon.

hadgi:
Thanks in advance for the fixing the bug part... and even bigger thanks in retrospect for quanteda :)

kbenoit:
Should be working now as of 4a5d606. The `min_docfreq = 0.9` will eliminate features that occur in fewer than 90% of your documents. In your example, with ndoc = 2, this means it will eliminate any feature that does not occur in both documents.

Issue #1181, opened 2018-01-04
Chaging min_count to percentile when  < 1.0?


koheiw:
`dfm_trim()` takes a fraction for `min_count` but it is hardly useful, because normalized frequency of words are very small, and it is really difficult to set a threshold like `min_count = 0.0001`. 

```
> colMeans(dfm_weight(dfmt, 'prop'))
               moscow                 trump                  said                  sure               whether 
         4.681466e-04          1.809751e-04          4.610098e-03          3.180712e-05          4.869455e-04 
                joint            us-russian         cybersecurity                 group             discussed 
         1.580790e-04          0.000000e+00          7.027506e-06          8.929641e-04          1.607924e-05 
              russian             president              vladimir                 putin                 might 
         3.366126e-04          1.563445e-04          0.000000e+00          8.911385e-06          2.685655e-04 
               really                   set                facing               overall            opposition 
         4.259691e-04          4.438829e-04          6.528476e-05          1.330071e-04          1.436127e-04 
                 idea                 moved                  back               working                  huge 
         6.577164e-04          5.672176e-05          1.017178e-03          7.629350e-04          1.788385e-04
```
It becomes if a fractional `min_count` is interpreted as percentile: i.e. `dfm_trim(min_count = 0.1)` means discard bottom 10% of words.

kbenoit:
The reason for making it a proportion instead of a percentile is that this creates an unambiguous difference between counts (`min_count` > 1) and relative amounts (0 < `min_count` < 1). This would not be the case with percentages.

Our users will just have to divide by 100 if they want to input a percentile!

koheiw:
I know the difference, but dividing by 100 does not work. If we want to keep the top 10% of the features, for example, we have to set `min_count = 0.0002388872`.

```r
corp <- data_corpus_irishbudget2010
dfmt <- dfm(corp, group = rep('all', ndoc(corp)))

freq <- colSums(dfmt)
freq_prop <- colSums(dfm_weight(dfmt, 'prop'))

dfm_trim(dfmt, min_count = quantile(freq, 0.9)) # min_count = 7
dfm_trim(dfmt, min_count = quantile(freq_prop, 0.9)) # min_count = 0.0002388872
```
We can find a good threshold like 5 or 10 for `min_count`, but who can find 1 / 5000 as a threshold without using `quantile()`? So I am proposing to do that automatically in `dfm_trim()` in the following way:
```r
if (min_count < 1)
[...]

PR #1239, opened 2018-02-17
Improve stability of dfm_trim()

kbenoit:
- fixes docfreq() to work even for weighted dfm objects.
- adds warning for leaving min_count default at 1 when using fractional docfreq values
- improves stability.

Addresses #1237.

Still leaves some behaviours that may be unexpected by some users, when trimming a weighted dfm with fractional values, or supplying partially fractional values as arguments. Related to #1181.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: quanteda
-Version: 1.0.4
+Version: 1.0.5
 Title: Quantitative Analysis of Textual Data
 Description: A fast, flexible, and comprehensive framework for 
     quantitative text analysis in R.  Provides functionality for corpus management,
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,7 @@
 
 * Eliminated unnecessary dependency on the **digest** package.
 * Updated the vignette title to be less generic.
+* Improved the robustness of `dfm_trim()` and `dfm_weight()` for previously weighted dfm objects and when supplied thresholds are proportions instead of counts.  (#1237)
 
 ### Performance improvements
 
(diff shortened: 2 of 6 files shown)


koheiw:
`dfm_trim()` can be useful for trimming a weighted DFM, but this use case brings us back to #1181. Since fractional `min_count` and `max_count` are interpreted as percentiles, there is not way to remove feature with TFIDF scores lower than 1.0. 

I also think that it is strange to take marginals (column sums) of TFIDF scores. For a TFIDF-weighted DFM, adding a `dfm_weight(limit = t)` which fills cell less than `t` with zero, and make eventually remove features (when columns are all zero).

kbenoit:
What if we add an o `dfm_trim(x, min_count = .01, literal_count = TRUE)` which means use the literal comparison of the count value to the cells. (This does not apply to docfreq.)

Or, we could call it: `count_type = c("proportion", "actual")` and if the min/max_count` value were > 1.0, then we would issue a warning and default to "actual". This will not change the current behaviour but then prints a warning for every count minimum! If we default to "actual" then it changes the behaviour for fractional values.

Aren't legacy issues fun. Welcome to > v1.0.

koheiw:
To make consistent with `valuetype`, how about `counttype`? We need `c('raw', 'proportion', 'percentile')` or, in short, `c('raw', 'prop', 'perc')`.

kbenoit:
We could keep the existing behaviour, if when 
```r
dfm_trim(x, min_count = .05)  # e.g.
```
and
```r
counttype = "raw" && min_count < 1 && x@weightTf == "count" && x@weightTf == "unary"
```
then we print a warning and call the function again using 
```r
dfm_trim(x, min_count = .05, counttype = "proportion")
```

I don't think we should use percentile since that's only different from proportion by a constant.

kbenoit:
I'm going to accept this since it improves the way `dfm_trim()` currently operates, without changing hte behaviour. In the meantime I will work on improving how `dfm_trim()` operates with the proportion v. count issue as we discussed recently (and will open an issue for this).

PR #1244, opened 2018-02-21
Dev wordcloud

koheiw:
This branch implements our own wordcloud with extra arguments to adjust its look. This is a legacy **graphics** based design, but compatible with old one. I will attempt full transition to **ggplot2** in the dev-wordcloud2 branch.


R/textplot_network.R
```diff
@@ -44,9 +44,9 @@
 #' @import network ggplot2 ggrepel
 #' @keywords textplot
 textplot_network <- function(x, min_freq = 0.5, omit_isolated = TRUE, 
-                             edge_color = 'skyblue', edge_alpha = 0.5, 
+                             edge_color = '#1F78B4', edge_alpha = 0.5,
```
kbenoit: I like the use of semantic colours better than the hex colours, as defaults. We can use some hex colours in the examples for illustration.

tests/depedency/r-version.R
```diff
@@ -0,0 +1,26 @@
+require(stringi)
+
+print_depends <- function(package, level = 0) {
```
kbenoit: Are there tests for this?

tests/testthat/test-default-methods.R
```diff
@@ -433,7 +433,7 @@ test_that("test default textplot methods", {
     )   
     expect_error(
         textplot_wordcloud(TRUE),
-        "textplot_wordcloud\\(\\) only works on dfm, tokens objects"
+        "textplot_wordcloud\\(\\) only works on dfm objects"
```
kbenoit: I agree that it makes no real sense to support `tokens` inputs.

R/textplot_wordcloud.R
```diff
@@ -1,70 +1,472 @@
...
+                               max_size = 4,
+                               max_words = 500,
+                               color = "#1F78B4",
+                               font = NULL,
+                               adjust = 0,
+                               rotation = 0.1,
+                               random_order = FALSE,
+                               random_color = FALSE,
+                               ordered_color = FALSE,
+                               labelcolor = '#4D4D4D',
```
kbenoit: Once again I prefer either a colour palette, or a semantic label.

R/textplot_wordcloud.R
```diff
@@ -1,70 +1,472 @@
...
+    }
+    word <- word[ord]
+    freq <- freq[ord]
+
+    theta_step <- 0.1
+    r_step <- 0.05
+    
+    graphics::plot.new()
+    op <- graphics::par(no.readonly = TRUE)
+    graphics::par(mar = c(0, 0, 0, 0), usr = c(-1, 1, -1, 1), family = font)
```
kbenoit: Too small - this truncates the edges of the cloud.

R/textplot_wordcloud.R
```diff
@@ -1,70 +1,472 @@
...
+    freq <- freq[ord]
+    group <- group[ord]
+    theta_step <- 0.05
+    r_step <- 0.05
+    
+    graphics::plot.new()
+    op <- graphics::par(no.readonly = TRUE)
+    graphics::par(mar = c(0, 0, 0, 0), family = font)
+    if (labelsize > 0) {
+        graphics::plot.window(c(-0.1, 1.1), c(-0.1, 1.1), asp = 1)
```
koheiw: This is margin for comparison cloud. We can make this adjustable through `margin`.

kbenoit: Better to calculate it and just get it right so that users don't have to. `wordcloud::comparison.cloud()` does this but adds so much margin that it is overkill. Can't we do something in between?

R/textplot_network.R
koheiw: The color is actually from a palet (second from the top):
```
 RColorBrewer::brewer.pal(8, 'Paired') 
# [1] "#A6CEE3" "#1F78B4" "#B2DF8A" "#33A02C" "#FB9A99" "#E31A1C" "#FDBF6F" "#FF7F00"
```
 This is an effort to set consistent color scheme across the package's visual output. Unfortunately, the colors do not have names in R and named R colors are not very good (this is why **RColorBrewer** exists).

tests/depedency/r-version.R
koheiw: This is a utility for us to to through all the R version requirement. We can remove and save somewhere else.

koheiw:
I agree that `text()` among many other things can be vectorized. I am changing these in in **dev-wordclound2** branch, but not here (this is only to incorporate the original code). It is slow because of the C++ function take a lot of time to find empty spaces.

R/textplot_wordcloud.R
kbenoit: When I clear all plots, and run this the first time
```r
textplot_wordcloud(obama_dfm, max_words = 500, rotation = 0.25, 
                   color = rev(RColorBrewer::brewer.pal(10, "RdBu")))
```
I see:
screenshot 2018-02-23 15 38 03

Then when I repeat the command a second time, I see:
screenshot 2018-02-23 15 38 14

It's as if the `par(op)` command does not reset the graphics parameters.

kbenoit: Or, it has something to do with verticality of the large words in the second plot.

koheiw: I now understood the issue. Pushed the fix.

kbenoit: I just tested it and it still occurs, for both comparison = FALSE and for comparison = TRUE. How to verify:

1. Clear the plot history using the broom icon in the Plots window. (Same as initializing a new graphics output device.
2. Run the example plot command. This produces a plot which looks correct.
3. Repeat the same plot command. This produces a plot with the top and bottoms sliced off. It actually looked worse when I tried it now than before. Happens for both values of comparison.

koheiw: I could see edge of the cloud to be cut off when I repeat the command, but fix seems to be easy. Please try again.

kbenoit:
The par issue seems ok now. However there is a new issue: The word plotting sizes in the comparison cloud seems to be out of balance. While I have not examined the code to confirm this, it looks like the size proportionality is being computed globally rather than within group, so the group with lower frequencies has its words all much smaller than the group with the terms that are more frequent. This is not apparently how the older function is computing it.

```r
textplot_wordcloud(obama_trump_dfm, comparison = TRUE)
```
rplot

```r
textplot_wordcloudold(obama_trump_dfm, comparison = TRUE)
```
rplot01

koheiw:
It is easy to do the same in the new one, but I wasn't really sure if the normalization in the old function is scientifically correct. We should look into it.

kbenoit:
Let's do the same, and in the meantime can look into it. Looks too odd otherwise!

R/textplot_wordcloud.R
```diff
@@ -1,70 +1,468 @@
...
+#'   order.
+#' @param labelcolor color of group labels. Only used when \code{compariosn=TRUE}.
+#' @param labelsize size of group labels. Only used when \code{compariosn=TRUE}.
+#' @param labeloffset  position of group labels. Only used when
+#'   \code{comparison=TRUE}.
+#' @param fixed_aspect if \code{TRUE}, the aspect ratio is fixed. Variable
+#'   aspect ratio only supported if rotation = 0
+#' @param comparison if \code{TRUE}, plot a wordclound that compares documents
+#'   in the same way as \code{\link[wordcloud]{comparison.cloud}}
+#' @param ... additional parameters passed to \link{text} (and \link{strheight},
```
kbenoit: This is indeterminate, since it passes any arguments not in the list or the deprecated set through to `graphics::strwidth()` and `graphics::strwidth()` and `text()`. The parameters for the first two functions are different from those for the third.

R/textplot_wordcloud.R
```diff
@@ -1,70 +1,468 @@
...
+    
+    font <- check_font(font)
+    x <- dfm_weight(x, 'prop')
+    x <- t(x) - Matrix::colMeans(x)
+    x <- as.matrix(x)
+
+    ndoc <- ncol(x)
+    theta_bins <- seq(0, 2 * pi, length = ndoc + 1)
+
+    if (is.null(color) < ndoc)
```
kbenoit: This is an error since the condition will always be true. Should this be `length(color)`? 

In calling the function, I am not able to change the colour when `comparison = TRUE`, and this statement is why.

Also, let's put the default color as a scheme in the function signature, not buried in the function.

kbenoit:
Also can we make the group labels in the comparison cloud a bit more prominent? Not necessarily with the gray background as in `wordcloud::comparison.cloud()` but something larger and perhaps boldface.

There is some inconsistent spacing between the comparison cloud and the cloud text with the comparison cloud, usually when the number of words plotted is small. But I noticed the same inconsistencies in `wordcloud::comparison.cloud()` however.

R/textplot_wordcloud.R
koheiw: Stopped passing ... to graphic functions.

R/textplot_wordcloud.R
koheiw: Fixed the bug, but not sure what to do with default colors. If it is not 'Paired', 'Accent' is good because its first two colors are very different (gray and green). This nonetheless goes away from setting a consistent color scheme.

Issue #1233, opened 2018-02-13
Confusing colors on the textplot_keyness plot.


brousseauj:
when plotting keyness, the colors used are red and blue, blue being the target. The target bars are blue but the text is red. Why the difference? Also, can we pass ggplot arguments to textplot to alter the style? 

![image](https://user-images.githubusercontent.com/17204759/36166620-f0d8b584-10a7-11e8-9c93-f7bed482daa9.png)

kbenoit:
Thanks for the feedback. @koheiw and I are in the process of fixing this plot, and will add to #1211 the issue about the words. I think they should be black by default. Also, when there just two classes, we will have the legend label read the name of the reference document, rather than the generic "Reference".

Yes, you should be able to change the plot using **ggplot2** `+` additions.

brousseauj:
Awesome! Off topic, thank you for building this package. It's truly powerful and has been the basis of building out my companies text analytics platform! I will be posting all the code in my GitHub!

kbenoit:
Thanks! I look forward to seeing that. Feel free to leave a testimonial at #461.

PR #1248, opened 2018-02-23
Issue 1213

koheiw:
Rewriting `textplot_keyness()` to address #1233, #1211, #1213.


kbenoit:
In the example for `textplot_keyness()`, the following fails:
```r
textplot_keyness(dem_key, show_reference = FALSE)
## Error: Aesthetics must be either length 1 or the same as the data (20): x, y, xend, yend, colour
```

koheiw:
I created two color schemes for our package using **colorspace**. These are palced in `test/misc/color.R`. The first color should work with all the plot including `textplot_network` so I kept it similar to `skyblue`. 

`quanteda_palette()`:
palette

`quanteda_palette2()`:
palette2

kbenoit:
I like the palette idea very much, let's absorb this PR and then work with the `colour.R` idea as a separate branch.  

I think it's going too far to make the defaults a "quanteda blue" scheme, since it's not contrasty enough for some plots. Keyness plots should have two colours closer to opposite ends of the rgb wheel, not from the same slice. I'd be fine with using ggplot defaults! 

The `color` arguments throughout should take the same values that are legal in **ggplot2**, which can include the integer colours. Ours does not work with integer colors. Can we change this?
```r
textplot_keyness(dem_key, color = 2:3)
## Error: Continuous value supplied to discrete scale
```

I changed it to allow a vector of colours length > 3, in case someone wants to use an RColorBrewer palette for instance.

kbenoit:
Please:

- [ ] review the colour changes to see if you can live them, and I promise a "quanteda blue" palette!
- [ ] make sure that `color` takes all ggplot legal values including integers, then change the @param documentation for that parameter
- [ ] add tests for `check_font()` in `utils.R`
- [ ] add roxygen documentation for `check_font()` in `utils.R`

Issue #1266, opened 2018-03-16
Selection argument in textstat_simil/dist is not working with interger index


koheiw:
```r
d1 <- dfm(c("a b c", "a b c d"))
d2 <- dfm(letters[1:6])
dtest <- dfm_select(d1, d2)


textstat_simil(dtest, margin = 'features', selection = 'a')
textstat_simil(dtest, margin = 'features', selection = 1) # error

textstat_simil(dtest, margin = 'documents', selection = 'text1')
textstat_simil(dtest, margin = 'documents', selection = 1) # error

textstat_dist(dtest, margin = 'features', selection = 'a')
textstat_dist(dtest, margin = 'features', selection = 1) # error

textstat_dist(dtest, margin = 'documents', selection = 'text1')
textstat_dist(dtest, margin = 'documents', selection = 1) # error

```

koheiw:
Probably this is not a bug but design. The documentation actually says if `selection` is a numeric vector, it is compares with `x`. However, taking column or raw index for `selection` is a more natural behavior.

kbenoit:
By design it should take a numeric index too, so it's a bug. Adding the ability to select from a numeric  index should be implemented. I suggest we add logical selection too.

koheiw:
If it takes numeric index, there will be an ambiguity whether the input to `selection` is an index for `x` or an object for which similarity will be calculated. One solution would be to restrict such objects to be a dfm.

kbenoit:
I agree that we cannot both allow an integer index _and_ an object supplied as a direct numeric for comparison with x. I think the original idea was to supply an object like the `y` in `proxy::similar()`. But since it's called `selection`, it should be a valid index for the `margin` specified, and this means character, integer, or logical. So for consistency, we should drop the direct supply of an object whose similarity will be computed, and make this argument behave as its name suggests.

Issue #1267, opened 2018-03-18
Improve error messages for invalid dfm indexing


kbenoit:
We could make these messages friendlier. In solving #1266, I have dropped error checking for `selection` indexing and prefer to reply on errors returned by sending an invalid index argument to `[.dfm` methods.
```r
> d <- dfm(c("a a b c", "b b a c c"))
> d
Document-feature matrix of: 2 documents, 3 features (0% sparse).
2 x 3 sparse Matrix of class "dfm"
       features
docs    a b c
  text1 2 1 1
  text2 1 2 2
> d[3, ]
Error in intI(i, n = x@Dim[1], dn[[1]], give.dn = FALSE) : 
  index larger than maximal 2
> d["text3", ]
Error in intI(i, n = x@Dim[1], dn[[1]], give.dn = FALSE) : 
  invalid character indexing
> d[c(TRUE, TRUE, TRUE), ]
Error in intI(i, n = x@Dim[1], dn[[1]], give.dn = FALSE) : 
  logical subscript too long (3, should be 2)
> d[, c(TRUE, TRUE, TRUE, FALSE)]
Error in intI(j, n = x@Dim[2], dn[[2]], give.dn = FALSE) : 
  logical subscript too long (4, should be 3)
> d[, c("a", "d")]
Error in intI(j, n = x@Dim[2], dn[[2]], give.dn = FALSE) : 
  invalid character indexing
```

koheiw:
How does a friendly message look like?
```r
> dim(x)
[1]  3 16
> x[5, 10]
  Error in x[5, 10] : Row index is too large. 
```
I thought "Document index is too large" is better, but it is also used by `fcm`, so row/column seems better.

kbenoit:
Probably just like indexing for `matrix` which states `"subscript out of bounds", although we could add "docname subscript out of bounds" or "featname subscript out of bounds" as appropriate.

koheiw:
If we use "docname subscript out of bounds" for rows, the error message will look funny for fcm, so I think just 
```r
 Error in x[5, 10] : Subscript out of bounds
```
would be the best. Much less scaly than 
```r
Error in intI(j, n = x@Dim[2], dn[[2]], give.dn = FALSE) : 
  logical subscript too long (4, should be 3)
```

PR #1268, opened 2018-03-18
Improve consistency of selection argument for textstat_dist/simil()

kbenoit:
Solves #1266.

Improves the consistency of the `selection` argument by enforcing it to be a valid index on the dfm dimension defined in `margin`. 

Also fixes a bug in `as.list.dist_selection()` that caused all but the first column (of the non-symmetric matrix) to be dropped, resulting in a list of length 1.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: quanteda
-Version: 1.1.2
+Version: 1.1.3
 Title: Quantitative Analysis of Textual Data
 Description: A fast, flexible, and comprehensive framework for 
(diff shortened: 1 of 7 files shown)


kbenoit:
Yes and the more **quanteda** grows, the more important that the code is lean, clean, and centralised in terms of checking as much as possible. The revision to the `selection` argument now puts error handling in dfm indexing, although we need to improve that with better error messages. (per #1267)

kbenoit:
There is still code repeated in the two functions (`textstat_dist()` and `textstat_simil()`) at the beginning and end, but it's still better than before.

PR #1276, opened 2018-03-20
 Set "smooth" slot of dfm in dfm_smooth()

### LINKED ISSUE #1274: Dfm attribute "smooth" is not set by dfm_smooth
The class documentation for `dfm` describes a `smooth` slot which holds the value of smoothed feature counts. As far as I understand, this value should be changed by `dfm_smooth` and `dfm_weight` if any smoothing is done. At the moment, neither function actually changes it.

chmue:
See discussion in #1274.

Fixes #1274


tests/testthat/test-dfm_weight.R
```diff
@@ -282,3 +282,19 @@ test_that("docfreq works previously a weighted dfm (#1237)", {
         computed
     )
 })
+
+test_that("smooth slot is correctly set (#1274)", {
```
kbenoit: I added these tests, since they should pass for this fix to have worked. They don't pass, which tells me that the `smooth` slot is getting reset somewhere. I would check the `as.dfm()` and the constructor methods for the `dfm` class. 

Test, test, test ... there's a lot to be said for [writing tests with expectations first and then coding](https://en.wikipedia.org/wiki/Test-driven_development) to meet the expectations. 

Also raises a question - what should `@smooth` be if we smooth a dfm by 1 and then smooth that smoothed dfm by 2?

chmue: :+1: will look into the tests and writing them!

I think if you smooth again, `@smooth` should be `x@smooth + smoothing` so 3 in your example. Because the entries in the dfm of `dfm_smooth(dfm_smooth(x, 1), 2)` should be exactly the same as `dfm_smooth(x, 3)`.

I fixed the code to not reset but add to `@smooth`. (I also fixed the test for this last case to use the dfms1).

chmue:
The tests still fail because the addition operation for dfms resets all the attributes.

The offending part is `as(x, "dgCMatrix")` which removes all slots that are not used in "dgCMatrix":

```r
str(as(data_dfm_lbgexample, "dgCMatrix"))
```
```

in [this line](https://github.com/quanteda/quanteda/blob/master/R/dfm-classes.R#L105)

**Comment by kbenoit** [comment]
Clearly we need more tests! We had this issue a lot early on, where operations from a parent class dropped attributes and slots from an inherited class. Looks like we didn’t catch all of the cases.

**Comment by chmue** [comment]
I think that the issue with the addition is different enough to warrant its own issue so I created #1279 to discuss how to best proceed to make additions keep attributes.

**Comment by chmue** [comment]
I added test for keeping slot values adding a constant to a dfm and fixed to code to copy the attributes over from the input dfm.

Tests seem to be ok. Could you have another look at the proposed changes @kbenoit ?

File: R/dfm-classes.R [hunk 2]
```diff
@@ -102,12 +102,12 @@ setMethod("rowMeans",
 #' @rdname dfm-class
 setMethod("+", signature(e1 = "dfm", e2 = "numeric"),
           function(e1, e2) {
-              as.dfm(as(e1, "dgCMatrix") + e2)
+            reassign_slots(as.dfm(as(e1, "dgCMatrix") + e2), e1)
```
kbenoit: @koheiw you added all of the `as.dfm()` functions to ensure backwards compatibility with older dfm versions that lacked slots we may have added later. Can you check whether this is the most stable way to keep the slot? It seems more consistent to use the `reassign_slots()` in `as.dfm()` to save the existing slot values there, rather than in each dfm-modifying function.

R/dfm-classes.R
```diff
@@ -102,12 +102,12 @@ setMethod("rowMeans",
...
           })
 #' @rdname dfm-class
 setMethod("+", signature(e1 = "numeric", e2 = "dfm"),
           function(e1, e2) {
-              as.dfm(e1 + as(e2, "dgCMatrix"))
+            reassign_slots(as.dfm(e1 + as(e2, "dgCMatrix")), e2)
```
kbenoit: Ditto.

R/dfm_weight.R
```diff
@@ -230,6 +230,7 @@ dfm_smooth.default <- function(x, smoothing = 1) {
 dfm_smooth.dfm <- function(x, smoothing = 1) {
     x <- as.dfm(x)
     if (!nfeat(x) || !ndoc(x)) return(x)
+    x@smooth <- x@smooth + smoothing
```
kbenoit: This is a policy decision, but I think this is the natural way for this to occur: we cumulate subsequent smoothing operations, so that the `@smooth` slot records the total smoothing applied to the _original_ dfm. @koheiw do you agree?

tests/testthat/test-dfm.R
```diff
@@ -409,6 +409,110 @@ test_that("dfm works with relational operators", {
...
+    tmp@docvars <- data.frame(test = letters[1:ndoc(tmp)])
+    expect_equal(
+        (tmp + 1)@docvars,
+        data.frame(test = letters[1:ndoc(tmp)])
+    )
+    expect_equal(
+        (1 + tmp)@docvars,
+        data.frame(test = letters[1:ndoc(tmp)])
+    )
+})
```
kbenoit: :+1:

kbenoit:
@koheiw I'll let you check (and update) this since you implemented the `as.dfm()` methods. See my comment above.

koheiw:
I added the second argument `slots` to `as.dfm()` so that it also set the S4 slot values. It made the internal logic more consistent, but broke `bootstrap_dfm()` because it generates duplicate document and data.frame errors when rownames are not unique. 

https://github.com/quanteda/quanteda/blob/7d77dbe283bba5db06435dff57dacda7c8200366/R/dfm-methods.R#L154-L155

One (simple) solution is to use `make.unique()`, but we could also stop using rownames entirely. What shall we do?

koheiw:
Segmented corpora have `_document` that can replace rownames. Every data.frame for document-variables should have  `_document`, `_docid`, and `_segid` so we can start moving toward that direction gradually.
```r
require(quanteda)
corp <- corpus(c('This is the first sentence. This is the second sentence.', 
                 'This is the 1st sentence. This is the 2nd sentence.'))
corp$documents

#                                                         texts
# text1 This is the first sentence. This is the second sentence.
# text2      This is the 1st sentence. This is the 2nd sentence.

corp_sent <- corpus_reshape(corp, "sentences")
corp_sent$documents

#                                texts _document _docid _segid
# text1.1  This is the first sentence.     text1      1      1
# text1.2 This is the second sentence.     text1      1      2
# text2.1    This is the 1st sentence.     text2      2      1
# text2.2    This is the 2nd sentence.     text2      2      2

corp_seg <- corpus_segment(corp, ".", pattern_position = "after")
corp_seg$documents

#                               texts _document _docid _segid pattern
# text1.1  This is the first sentence     text1      1      1       .
# text1.2 This is the second sentence     text1      1      2       .
# text2.1    This is the 1st sentence     text2      2      1       .
# text2.2    This is the 2nd sentence     text2      2      2       .
```

kbenoit:
I agree this is a very good idea. Ideally we would have a data object like an SQL table with a `UNIQUE` constraint on the id variable, but there is no such R object that I know of.

How about this for an idea: We change the `corpus$documents` object to a data.table. A data.table has no rownames, and has very fast functions to check for duplications.

```r
library(data.table)
dt <- data.table(text = c("This is the first sentence.",
                          "This is the second sentence.", 
                          "This is the 1st sentence.", 
                          "This is the 2nd sentence."), 
                 "_docid" = c("text1", "text1", "text2", "text2"),
                 "_segid" = c(1, 2, 1, 2)) 

# sets the "key" for the entire data.table
setkeyv(dt, c("_docid", "_segid"))

# the test that no docid/segid combination are repeated
stopifnot(!any(duplicated(dt, by = c("_docid", "_segid"))))

# fails for docid alone
stopifnot(!any(duplicated(dt, by = c("_docid"))))
## Error: !any(duplicated(dt, by = c("_docid"))) is not TRUE
```

From `?data.table::duplicated`:
> Because data.tables are usually sorted by key, tests for duplication are especially quick when only the keyed columns are considered. Unlike `unique.data.frame`, paste is not used to ensure equality of floating point data. It is instead accomplished directly and is therefore quite fast.

kbenoit:
Note we would need to add an attribute for the corpus version, and filter all objects through an `as.corpus.corpus()` - as we do for dfm inputs - to update objects before using them.

koheiw:
OK. The problem is in this line: 
https://github.com/quanteda/quanteda/blob/29c972c156cddb83da9f1a81516c783b0075d7a2/R/bootstrap_dfm.R#L88
If `x` is just a `dgCMatrix`, it should not error.

kbenoit:
Yes but we are failing tests in `test-dfm.R` as well. We can probably solve it in `[.dfm`, by allowing duplicate indexes. `dfm_compress()` assumes that there can be duplicate indexes and some functions even reply on that.

kbenoit:
@koheiw can you fix this one since you made the slot changes?

koheiw:
We can see this error even in the master:
```r
as.dfm(matrix(c(2, 1, 3, 2), nrow = 2, 
       dimnames = list(c('text1', 'text1'), c('word1', 'word2'))))
#  Error in data.frame(row.names = rownames(x)) : duplicate row.names: text1
```

`rbind.dfm` works in the master but docvars only have serial number
```r
mt <- rbind(dfm('a b c'), dfm('b c d'))
docvars(mt, 'n') <- 100:101
docvars(mt)
#     n
# 1 100
# 2 101
row.names(docvars(mt))
# [1] "1" "2"
```

We have only few options to deal with the duplicated document names:

1. make document names unique using `make.unique()`
2. save document names in `_docmane` instead of as `row.names`
3. give serial number to row names
4. assign NULL to the `docvar` slot

We did not face this problem because many of the dfm methods is doing 4, but I don't think this is right in terms of consistency. We made sure that corpus object always have data.frame in docvar, but NULL is allowed by `setClassUnion("dframe", members = c("data.frame", "NULL")) ` for a dfm. I prefer 2 or 3, which is similar to the solution to #1236.

kbenoit:
I think we might break even more behaviours and tests by not allowing duplicated dimension names in a dfm. Both a matrix and a (sparse) Matrix allow duplicate row names, although it's arguable that they should not, since it makes the indexing indeterminate: named indexes seem to take the first occurrence, which is a big problem when the rows have no innate order.
```r
mt <- rbind(dfm("a b c"), dfm("b c d"))
(mtm <- as.matrix(mt))
mtm["text1", , drop = FALSE]
#        features
# docs    a b c d
#   text1 1 1 1 0
```

Should we try to match the matrix and Matrix behaviour for the dimension name rules?

Do really need the `_docid` and `_segid` etc in the dfm's @docvars slot?

koheiw:
'_docid' and '_segid' are necessary only for `corpus_reshpe()` and `bootstrap_dfm()` at the moment.

The errors are occurring not from Matrix, but from data.frame. When we synchronize row names between them, data.frame fails due to duplication. There is no change in how we subset a dfm.

koheiw:
This is how we are avoiding name collision:
https://github.com/quanteda/quanteda/blob/033a435644c93f5bf33ae741223c4e76db742c87/R/corpus.R#L180-L181
Then, this is consistent approach.
```
x <- new("dfm", x, docvars = data.frame(row.names = make.unique(rownames(x)))) 
```

kbenoit:
The more "matrix-like" way would be to stop using rownames for the dfm docvars, and use a `_docname` variable instead, which is your proposal 2. above.  That way we retain the ability to have repeated row labels in a dfm.

So: 
* a corpus _must_ have a unique document name for each document; but
* a dfm may consist of rows that repeat data for the same document. We will do our best to keep the docvars consistent with the newly dimensioned dfm, but this will be imperfect if the dfm has been transformed by the user. (This is exactly how `dfm_group()` performs, for instance.)

koheiw:
Disentangled the dfm/fcm spaghetti through a series of changes. It is only failing two tests due to the dovars inconsistency (it currently does not assign row names to docvar data.frame):
```
test-as.dfm.R:87: failure: as.dfm for tm matrix objects
as.dfm(dtm) not equal to dfm(txt).
Attributes: < Component “docvars”: Attributes: < Component “row.names”: Modes: numeric, character > >
Attributes: < Component “docvars”: Attributes: < Component “row.names”: Lengths: 0, 3 > >
Attributes: < Component “docvars”: Attributes: < Component “row.names”: target is numeric, current is character > >

test-as.dfm.R:94: failure: as.dfm for tm matrix objects
as.dfm(tdm) not equal to dfm(txt).
Attributes: < Component “Dimnames”: names for current but not for target >
Attributes: < Component “docvars”: Attributes: < Component “row.names”: Modes: numeric, character > >
Attributes: < Component “docvars”: Attributes: < Component “row.names”: Lengths: 0, 3 > >
Attributes: < Component “docvars”: Attributes: < Component “row.names”: target is numeric, current is character > >
```

koheiw:
One vote for `_docname`. This is actually necessary if we want to use `XPtr` as in the `dev-xpointer` branch to improve the performance, because `XPtr` does not have name attribute. We need to change test expectation, but switching from row names to `_docnames` should be easy. We should do this in a new branch, after merging this.

kbenoit:
Decided: When `x` is a dfm, `x@docvars` should always be a data.frame and have `ndoc(x)` by the number of docvars dimensions. When there are no docvars, the columns of this data.frame will be zero. This is consistent with behaviour for a corpus docvars in the master branch, but inconsistent with how some dfm operations set the docvars of a dfm to `NULL` in the `master`.

We will open a new issue for `cbind.dfm()` and `rbind.dfm()` after we fix this, to make their behaviour more consistent.

To improve docvars more globally, a next major revision will replace them with `data.table` objects, to solve the inconsistencies with the row.names issues. (We could also consider tibbles.)

koheiw:
`textmodel_lsa()` is OK, when `margin = both`, the lower rank matrix has the same number of row and columns as the original.

```r
dim(ie_dfm)
# [1]   14 5140

dim(textmodel_lsa(ie_dfm)$matrix_low_rank)
# [1]   14 5140
dim(as.dfm(textmodel_lsa(ie_dfm)))
# [1]   14 5140

dim(textmodel_lsa(ie_dfm, margin = "features")$matrix_low_rank)
# [1] 14 10
dim(as.dfm(textmodel_lsa(ie_dfm, margin = "features")))
# [1] 14 10

dim(textmodel_lsa(ie_dfm, margin = "documents")$matrix_low_rank)
# [1]   10 5140
dim(as.dfm(textmodel_lsa(ie_dfm, margin = "documents")))
# [1]   10 5140

```

Issue #1278, opened 2018-03-21
dfm drops dictionary keys that do not exist in tokens when tolower = TRUE


koheiw:
Related to #1277 

```r
toks <- tokens("a b c")
dict <- dictionary(list(A = "a", B = "b", Z = "z"))

toks_key <- tokens_lookup(toks, dict)
types(toks_key)
# [1] "A" "B" "Z"

dfm(toks_key)
# Document-feature matrix of: 1 document, 2 features (0% sparse).
# 1 x 2 sparse Matrix of class "dfm"
# features
# docs  a b
# text1 1 1
```
However, this is correct:
```r
dfm(toks, dictionary = dict)
# Document-feature matrix of: 1 document, 3 features (33.3% sparse).
# 1 x 3 sparse Matrix of class "dfm"
# features
# docs  A B Z
# text1 1 1 0
```
This is correct too. 
```r
dfm(toks_key, tolower = FALSE)
# Document-feature matrix of: 1 document, 3 features (33.3% sparse).
# 1 x 3 sparse Matrix of class "dfm"
# features
# docs  A B Z
# text1 1 1 0
```

PR #1280, opened 2018-03-21
Disable gap re-indexing to keep unused dictionary keys in case-changes

koheiw:
This fixes a bug in `tokens_tolower()` that removes unused types (dictionary keys #1278). The source of the problem was deep in the code, but my fix is straightforward.


kbenoit:
Also please increment the version minor number by in `DESCRIPTION` by 01 and add a note to NEWS about this fix.

Issue #1285, opened 2018-03-23
Quanteda overrides base summary behavior 


ay60:
Not sure if that is bug or a feature but that makes running summary against big data frames difficult.

Without quanteda loaded:

```
> summary(data.frame(a=c("aaa","bbb","ccc"),stringsAsFactors=FALSE))
      a            
 Length:3          
 Class :character  
 Mode  :character  
```

With quanteda loaded:

```
> summary(data.frame(a=c("aaa","bbb","ccc"),stringsAsFactors=FALSE))
    a      
 NULL:aaa  
 NULL:bbb  
 NULL:ccc  
```

Quanteda version is `quanteda_1.1.1`

With quanteda loaded running summary against a data frame with a character column and with many rows it takes a very long time to run and produces not very useful result.

kbenoit:
I investigated this and am fixing it now, but it turns out it's not **quanteda**, but rather **network**.  We imported its entire namespace, however, so passed on the problem of it providing a `summary.character()` method. Will be fixed soon.

```r
> char <- c("aaa", "bbb", "ccc")
> df <- data.frame(char, stringsAsFactors = FALSE)
> 
> summary(char)
   Length     Class      Mode 
        3 character character 
> 
> summary(df)
     char          
 Length:3          
 Class :character  
 Mode  :character  
> library("network")
network: Classes for Relational Data
Version 1.13.0 created on 2015-08-31.
copyright (c) 2005, Carter T. Butts, University of California-Irvine
                    Mark S. Handcock, University of California -- Los Angeles
                    David R. Hunter, Penn State University
                    Martina Morris, University of Washington
                    Skye Bender-deMoll, University of Washington
 For citation information, type citation("network").
 Type help("network-package") to get started.

> summary(char)
aaa bbb ccc 
  1   1   1 
> summary(df)
   char    
 NULL:aaa  
 NULL:bbb  
 NULL:ccc 
```

ay60:
Thank you! 
Wonderful package by the way, thank you. I completely switched over from using tm to qaunteda now.

kbenoit:
@ay60 please re-install the latest `master` and try. Should be fixed now in b0bf173.

ay60:
It seems to have been fixed on the mainline. Thank you!

PR #1288, opened 2018-03-24
Fix issue 1285 summary.character()

kbenoit:
- Avoids namespace conflicts with **network**, and no longer imports that package's namespace, which was rewriting `summary.character()` and causing #1285.

- Improves tests.

- `summary.character()` in **quanteda** is renamed to `summary_character()` just to prevent any conflict.

- **network** is now in Suggests, not Imports, to avoid a namespace conflict warning on loading.

- Improves nfunction return consistency and tightens the tests to reflect the exact return object type.


kbenoit:
The solution was way too tedious to this problem, which was caused by somehow the **network** package's namespace being loaded in a way that included that package's rewrite of `summary.character()`, even though we were not calling that function. (We were calling a function that used it, and apparently that loaded this part of **network**'s namespace.) 

It was also a very hard condition to test.

My solution was to reimplement the `as.character()` method in **quanteda** so that it simply calls the `summary.default()` when supplied a `character`, which is how `summary(somecharacter)` behaves without getting **network** involved. 

A flaw with R's namespace system, but also a flaw in the **network** package, which hasn't been updated in 4 years, is not on a public repo, and does not use devtools/roxygen/etc. We should try to wean ourselves of that package entirely.

koheiw:
Go back to igraph?

kbenoit:
It certainly seems to be better maintained!

Issue #1322, opened 2018-04-22
Make a tokens_tortl() and tokens_toltr()


koheiw:
We already have `tokens_tortl()`:
https://github.com/ropensci/textworkshop18/issues/16

`tokens_toltr()` is only to remove the direction marker "\\u200F". We can also add `rtl` or `direction` argument to `tokens()` or `corpus()` to make this functionality more visible.

koheiw:
I am actually not sure if `tokens_toltr()` should only remove the direction marker or replace it with "\u200E".

Direction makers pass through tokenization, so it might be easier to be done in `corpus(rtl = TRUE)`. If we want to keep texts intact in corpus objects, `tokens(rtl = TRUE)` is the answer.

koheiw:
To avoid modifying tokens, we should insert the direction mark after punctuations in `print.tokens()` and `print.dfm()`, but this means that we will add `direction` attribute/slot to these objects.

```r
> dfm(toks[1])
Document-feature matrix of: 1 document, 6 features (0% sparse).
1 x 6 sparse Matrix of class "dfm"
       features
docs    מדינת ישראל נוסדה בשנת 1948 .
  text1     1     1     1    1    1 1
> dfm(tokens_tortl(toks[1]))
Document-feature matrix of: 1 document, 6 features (0% sparse).
1 x 6 sparse Matrix of class "dfm"
       features
docs    מדינת ישראל נוסדה בשנת 1948 .‏
  text1     1     1     1    1    1 1
```

PR #1327, opened 2018-04-27
Issue 1322

koheiw:
After contemplation #1322, I decided to make `tokens_tortl()` and `char_tortl()` as experimental functions to let native right-to-left language speakers to use. We could change `tokens()` to add this functionality, but I did to keep other functions changed.


tests/testthat/test-directionchange.R
```diff
@@ -0,0 +1,20 @@
...
+
+# Note: the line below does not appear correctly in R Studio
+
+test_that("char_tortl works.", {
+    skip_on_os("windows")
+    txt <- "מדינת ישראל נוסדה בשנת 1948."
+    expect_equal(char_tortl(txt), "מדינת ישראל נוסדה בשנת 1948.\u200F")
+    
+    txt <- "!מדינת ישר computer וסדה בשנת 1948."
+    # expect_equal(char_tortl(txt), "!מדינת ישר computer וסדה בשנת 1948.")
```
kbenoit: What's the expectation here, when there is a ltr word sequence sandwiched in between two sequences of rtl words?

Note that I could not even get this to edit properly on my console! and I am not sure what the correct expectation is.

PR #1328, opened 2018-04-30
Minor documentation typo fixes

kbenoit:
These don't affect the code, but I wanted to check with you @koheiw re the new pkgdown updating workflow before I merged them into master. These _should_ update to the pkgdown site, but only these.


koheiw:
I will try to do that by branching out the CRAN release and merging this branch on to that.

kbenoit:
Or build reference, but only push the new html for the two revised functions?

Issue #1372, opened 2018-05-30
tokens converts NA as "NA"


kbenoit:
Discovered in #1361.

Here, NA should be rendered as "", not "NA".
```r
tokens(c("a a b", NA))
## tokens from 2 documents.
## text1 :
## [1] "a" "a" "b"
## 
## text2 :
## [1] "NA"
```

PR #1373, opened 2018-05-30
 Handle NA document properly 

koheiw:
- Substitute NA with empty string in corpus constructor to address #1372 
- Change test expectations

### DIFF
--- a/tests/testthat/test-kwic.R
+++ b/tests/testthat/test-kwic.R
@@ -2,39 +2,36 @@ context('test kwic.R')
 
 test_that("test attr(kwic, 'ntoken') with un-named texts", {
     testkwic <- kwic(c(
-        'The quick brown fox jumped over the lazy dog',
-        'The quick brown fox',
-        'The quick brown dog jumped over the lazy dog',
+        "The quick brown fox jumped over the lazy dog",
+        "The quick brown fox",
+        "The quick brown dog jumped over the lazy dog",
         NA
-    ), 'fox')
-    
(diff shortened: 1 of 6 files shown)


tests/testthat/test-kwic.R
```diff
@@ -2,39 +2,36 @@ context('test kwic.R')
 
 test_that("test attr(kwic, 'ntoken') with un-named texts", {
     testkwic <- kwic(c(
-        'The quick brown fox jumped over the lazy dog',
```
kbenoit: 👍

tests/testthat/test-tokens.R
```diff
@@ -420,23 +420,23 @@ test_that("tokens works as expected with NA, and blanks", {
     )   
     expect_equal(
         as.list(tokens(c("one", NA, ""))),
-        list(text1 = "one", text2 = "NA", text3 = character())
```
kbenoit: Wow - we actually tested the error condition as if we expected it! This was originally clearly not a good test.

tests/testthat/test-tokens.R
```diff
@@ -420,23 +420,23 @@ test_that("tokens works as expected with NA, and blanks", {
...
+        list(text1 = "one", text2 = character(), text3 = character())
     )   
     expect_equal(
         as.list(tokens(c(NA, "one", ""))),
-        list(text1 = "NA", text2 = "one", text3 = character())
```
kbenoit: Ditto

Issue #1260, opened 2018-03-09
feature coocurrence as a graph -> fcg or fcm_graph ?


aourednik:
Many thanks for developing quanteda! The fcm feature is fast and very useful. This is an enhancement proposal.
Trying to apply textplot_network() or as.network() to a large fcm triggers an error: "fcm is too large for a network plot". This makes sense, since a visualization would take too much resources. But it would be nice to be able to convert the whole sparse matrix to a graph for graph-oriented treatment. Since the fcm seems of class dgTMatrix, it should be possible to convert to a graph with T2Graph(), as documented here:  https://stat.ethz.ch/R-manual/R-devel/library/Matrix/html/graph2T.html 
This triggers an error, though. Igraph's function graph_from_adjacency_matrix() makes it possible.
But this solution makes us leave the context of quanteda. Maybe a fcg or fcm_graph object would be a useful new feature, where word concurrence would be a graph object instead of a sparse matrix. Features and their links could then be annotated with extra metadata. fcm_graph could also propose  export functions to gml or [json](https://bl.ocks.org/mbostock/4062045), for direct interoperability with Cytoscape, Gephi or D3. The following solution does this. Would a tighter integration of this functionality in quanteda be possible?
```R
library("Matrix")
library("igraph")
#txtfcm is an fcm object produced with fcm(tokens(corpus(readtext(set_of_txt_files))))
txtfcm.graph <- T2graph(txtfcm) # Triggers an error: no slot of name "j" for this object of class "fcm"
txtfcm.graph <- graph_from_adjacency_matrix(txtfcm,weighted=TRUE) # works
V(txtfcm.graph)$freq <- rowSums(txtfcm)
txtfcm.graph <- simplify(txtfcm.graph) # remove loops and duplicate edges
txtfcm.graph <- delete_edges(txtfcm.graph,which(E(txtfcm.graph)$weight<0.5)) # delete weak coocurrences
txtfcm.graph <- delete_vertices(txtfcm.graph,which(V(txtfcm.graph)$freq < 3)) # delete low frequecy words
# associate an attribute to each vertex (word.fr.positive and word.fr.negative are lists of words)
V(txtfcm.graph)$negative <- sapply(rownames(topfeat),function(x){
  return(V(txtfcm.graph)$name %chin% words.fr.negative)
})
V(txtfcm.graph)$positive <- sapply(rownames(topfeat),function(x){
  return(V(txtfcm.graph)$name %chin% words.fr.positive)
})
[...]

PR #1376, opened 2018-06-07
Add as.igraph.fcm()

koheiw:
Uer will have access to all the functions in **igraph** through **as.igraph()**. We should also consider adding `textplot_network.igraph()` to allow users to color vertices according to community membership. This is a partial response to #1260.


kbenoit:
I left the definition of the `as.igraph.fcm()` method structured as you had it @koheiw although
1.   this only works if the **igraph** package is installed, since it's only in `Suggests:` (not `Imports:`).  
2.  it will report a namespace conflict if **igraph** is loaded after quanteda.

PR #1385, opened 2018-06-23
Add attributes to tokens object after combining

koheiw:
- Disallow combining tokens with different concatenate or what values
- Join and sort ngrams or skip values


koheiw:
This is for #1385

Issue #1388, opened 2018-06-27
Add warning/error if at least one column name is missing


stefan-mueller:
## Describe the bug

If a column name of a data.frame is missing, constructing a `corpus()` still works. Yet, `summary()`, for example, results in an error due to undefined docvars. I was wondering whether we might add a test to the corpus construction stage to check that column names are not `NA`?

## Reproducible code

```r
library(quanteda)

df <- data.frame(text = letters, docvar1 = paste(letters, letters), num = 1:26, 
  stringsAsFactors = F)

colnames(df) <- c("text", "docvar1")

corp <- corpus(df)

corp
#> Corpus consisting of 26 documents and 1 docvar.

summary(corp)
#> Warning in nsentence.character(object, ...): nsentence() does not correctly
#> count sentences in all lower-cased text
#> Error in `[.data.frame`(x, !is_system & !is_text): undefined columns selected
```

## Expected behavior

Already return an error (or at least a warning) when a corpus is constructed from an object with missing column names.

```
R version 3.5.0 (2018-04-23)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS Sierra 10.12.6

Matrix products: default
BLAS: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.5/Resources/lib/libRlapack.dylib

locale:
[1] en_IE.UTF-8/en_IE.UTF-8/en_IE.UTF-8/C/en_IE.UTF-8/en_IE.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] reprex_0.1.2   quanteda_1.3.0

loaded via a namespace (and not attached):
 [1] Rcpp_0.12.17       compiler_3.5.0     pillar_1.2.1       formatR_1.5       
 [5] plyr_1.8.4         tools_3.5.0        stopwords_0.9.0    digest_0.6.15     
 [9] debugme_1.1.0      lubridate_1.7.4    evaluate_0.10.1    memoise_1.1.0     
[13] tibble_1.4.2       gtable_0.2.0       lattice_0.20-35    rlang_0.2.1       
[17] Matrix_1.2-14      fastmatch_1.1-0    rstudioapi_0.7     yaml_2.1.19       
[...]

PR #1391, opened 2018-07-03
Revert "Disallow data.frame without colnames (#1388)"

kbenoit:
Reverts quanteda/quanteda#1390


kbenoit:
This replaces empty or NA document variables with `V1`, `V2`, etc. This is the same behaviour as `as.data.frame(matrix(1:6, nrow = 2))` for instance. 

It’s written “smartly” so that if the missing docvar is the third document variable, it will name the missing one `V3` instead of `V1`.

tests/testthat/test-corpus.R
```diff
@@ -466,15 +466,54 @@ test_that("corpus handles NA correctly (#1372)", {
...
+                     dvar1 = 1:5,
+                     dvar2 = letters[22:26],
+                     dvar3 = 6:10,
+                     stringsAsFactors = FALSE)
+    
+    # when one column name is NA
+    names(df)[3] <- NA
+    expect_equal(
+        corpus(df) %>% docvars() %>% names(),
+        c("dvar1", "V2", "dvar3")
```
koheiw: I am not sure about this expectation. I think this is more natural to make it `"dvar1", "V1", "dvar3"`, ignoring the named ones. If there column original names are `"V1", NA, "V2"`, then `"V1", "V3", "V2"`.

Or should the new column names be followed by column index?

kbenoit: Ah, you just found a potential run-time error - if the original names of the data.frame were `c("V2", NA)`. So we should probably issue one more call to `make.names(names(x, unique = TRUE)`. I can add that.

That's why we should always get a second pair of 👀  for a review!

koheiw: If `c("V2", NA)` becomes `c("V2", "V1")`, there will be no name collision.

This is the proposed code.
```
l <- !nzchar(names(x)) | is.na(names(x))
name <- setdiff(paste0("V", seq(length(x))), names(x))
names(x)[l] <- head(name, sum(l))
```
However, it seems that `as.data.frame.matrx()` dose nothing with colnames when at least one of them has a name.

```r
> m <- matrix(1:4, nrow = 2)
> m
     [,1] [,2]
[1,]    1    3
[2,]    2    4

> colnames(m)[2] <- "V1"
> as.data.frame(m)
  NA V1
1  1  3
2  2  4
```

kbenoit:
How about this?
```r
df <- data.frame(thetext = LETTERS[1:5],
                     docID = paste0("txt", 1:5),
                     x = 1:5,
                     V3 = letters[22:26],
                     x = 6:10,
                     stringsAsFactors = FALSE)
    names(df)[c(3, 5)] <- c(NA, "")
    crp <- corpus(df, text_field = "thetext", docid_field = "docID")
names(docvars(crp))
## [1] "V1"   "V3"   "V3.1"
```

I don't think we need to make this too user-friendly, because users should not really be feeding data.frames with no column names to `corpus()`.

koheiw:
I revered my change, because this is what we should do. 
```r
> df <- data.frame(Var.1 = 1:3, Var.3 = 1:3, 1:3, c("a", "b", "c"), fix.empty.names = FALSE)
> colnames(df)[4] <- NA
> df
  Var.1 Var.3   NA
1     1     1 1  a
2     2     2 2  b
3     3     3 3  c
> data.frame(df)
  Var.1 Var.3 Var.3.1 NA.
1     1     1       1   a
2     2     2       2   b
3     3     3       3   c
```
This is only to change `x <- as.data.frame(x)` to `x <- data.frame(x)`. data.frame is terribly mysterious.

PR #1402, opened 2018-07-18
Improve normialization of weights

koheiw:
Devision by max value sometimes make other edges very thin. Devision by a 90% percentile value seems to be much more robust.


kbenoit:
I didn't get a chance to review this PR but is this something we should document?

Issue #1413, opened 2018-08-09
Allow asymmetic window sizes in fcm()


enrico200165:
posted question [here on stackoverflow](https://stackoverflow.com/questions/51765064/quanteda-how-to-get-ngrams-and-their-frequences-given-n-1-predecessor-words-t
) without answer yet 
Is there a way to access a dfm using the n-1 precedessor types as indexes values?

[abstract pasted from link above] "For next word prediction using ngrams I would need to find all the ngrams (and their frequencies) given n-1 predecessor words.
In dfm I could not see any way to do that, so started implementing it manually on texstat_frequency (data.frame). ...
(Implicitly maybe wrongly excluding using regexes, that I normally love, becauses of prejudice that running them on hundred thousands strings might be too slow/heavy)

koheiw:
I thought this can be done by using asymmetric `window` values in `fcm()` but it raises error:

```r
> require(quanteda)
> txt <- c("a b c", "b d e")
> fcm(tokens(txt), "window", window = c(0, 1), tri = FALSE,)
Error in try(if (window < 2) stop("The window size is too small.")) : 
  The window size is too small.
In addition: Warning message:
In if (window < 2) stop("The window size is too small.") :
  the condition has length > 1 and only the first element will be used
Error in qatd_cpp_fcm(x, length(types), count, window, weights, ordered,  : 
  Expecting a single value: [extent=2].
```

koheiw:
Alternatively, the size of the window could be only an integer, but we could record words before/after in upper/lower triangle when a new argument `asymmetric = TRUE`.

kbenoit:
That’s what `ordered = TRUE` already does: shows in rows a term and then in columns, the counts of the words that occur following the row word, within the (one-sided/asymmetric) window size.

kbenoit:
Yes, that's what I found. If this did work, and we removed the`window >=2` condition, then it could easily serve as a generator for a bigram transition matrix. 

We need better substantive tests on this, clearly - should start a new issue to fix this.

enrico200165:
I installed today 20th August from github, I am still unable to see/find the  a simple way, updated question [on SO](url
![image](https://user-images.githubusercontent.com/250743/44348556-083e8000-a49b-11e8-9144-7d5e6ae67bce.png)
) to include my unsuccessful/wrong way and a simple text to be used if you reply with some sample code to illustrate how to do it

koheiw:
Don't you see something like this?
```r
[...]

PR #1415, opened 2018-08-15
Issue 1413

koheiw:
Fix counting when ordered = TRUE to solve #1413


kbenoit:
Nice fix @kohei, thanks. Not sure why our systems rebuilt all of the .Rd files. New version of **roxygen**?

PR #1418, opened 2018-08-27
Improve docvar assignment for dfm as a value

### LINKED ISSUE #1417: `docvars <-` calls deprecated as.data.frame.dfm when using a dfm
## Describe the bug

`data.frame()` in https://github.com/quanteda/quanteda/blob/master/R/docvars.R#L151 leads to following warning:

```
Warning message:
'as.data.frame.dfm' is deprecated.
Use 'convert(x, to "data.frame")' instead.
See help("Deprecated") 
```

## Reproducible code

```r
library(quanteda)
corpus.a <- corpus(data_char_ukimmig2010)
docvars(corpus.a) <- dfm_lookup(dfm(corpus.a), dictionary(list(keyword = 'the')))
```
## System information

```
quanteda_1.3.4
```

## Additional info

Oh and I just see that there is a missing `=` in the warning message. It should be `to = "data.frame"`.

kbenoit:
- removes `as.data.frame.dfm` deprecation warning
- improves `docvars(anycorpus) <- anydfm` 
- Fixes #1417


kbenoit:
We still have some differences in docvar assignment at the tokens and dfm stage - namely that it does not add row.names to the docvars - but this might be better than the corpus approach of using them. Eventually, however, we will need to make these consistent.

PR #1434, opened 2018-09-20
Misc documentation corrections

kbenoit:
Misc corrections brought to my attention by Arnab Dash (thanks!).


kbenoit:
@koheiw I added `corpus_trim()` and `char_trim()` to the documentation index (previously leyworded internal). Can you see any reason not to?

Issue #1426, opened 2018-09-06
Improve unused_dots() 


kbenoit:
Could make `unused_dots()` more informative. Right now:

```r
> textstat_readability(data_corpus_inaugural[1], measure = "Flesch.Kincaid", notarg = FALSE)
         document Flesch.Kincaid
1 1789-Washington       28.43216
Warning message:
In unused_dots(...) : ... is not used in textstat_readability.character()
```

Should:
* identify the unused arguments; and
* not print the name of the `unused_dots()` function itself.

PR #1448, opened 2018-10-09
Make unused_dots() more informative for #1426

koheiw:


koheiw:
```r
> corpus(data_char_ukimmig2010, something = TRUE, whatever = FALSE)
Corpus consisting of 9 documents and 0 docvars.
Warning message:
something, whatever arguments are not used in corpus.character()
```

Issue #1449, opened 2018-10-11
Add separator argument to kwic()


koheiw:
I found this ad-hoc solution to reconstruct original texts, but this is a wrong approach:
https://github.com/quanteda/quanteda/blob/edc0467173be541591068fb8aa3c09906237aebf/R/corpus.R#L371-L378

We instead add `separator` to `kwic()` and use `remove_separators` that is passed to `tokens()`.
```r
kw <- kwic(corp, separator = "", remove_separators = FALSE)
corpus(kw)
```
This is very easy to do.

PR #1450, opened 2018-10-12
Add separator argument for #1449

koheiw:
For #1449. I need this to be merged to upgrade `corpus.kwic()` in the dev-corpus2 branch.


koheiw:
We do, because phrases are comprised of individual tokens.

Issue #1210, opened 2018-02-01
Add min threshold to textstat_dist()?


koheiw:

kbenoit:
Not exactly a clear feature request, but if you mean an option to return a reduced size dist object based on the dist values, this is not really workable since the threshold would differ wildly across the matrix.

But maybe I've misunderstood. Are there examples of this for other dist operations? (from **proxy** for instance?)

koheiw:
I did not have time to explain until now. I noticed that people are struggling to calculate pairwise similarity between large number of sentences or n-grams to detect reuse of texts because it generates large dense matrix. 

We can make it more efficient if we floor values lower than a threshold to zero. Flooring does not have any impact on the analysis because people are usually only interested in very high similarity. I haven't seen the code really, but could be done either in `textstat_simil` or `textstat_dist`.

koheiw:
This is where flooring will should happen:

https://github.com/quanteda/quanteda/blob/76e2cbff386fb8ef4dcf079b4f6e9971137dda8a/src/dist_mt.cpp#L218

kbenoit:
OK, I see. So in the row/column loop in C++ we would just return a zero if the distance were below the minimum, and that would be one less cell in the sparse object returned by the C++ functions, so it is a smaller object returned to R.

The current return to R is a dist object, which is an efficient lower-diagonal vector of a distance matrix, basically a vector with a separate record of the object dimensions. But a zero in that object would still be a zero numeric, so we would have to change the return from `textstat_dist()` object to be something like that the `fcm` class is currently. Would also need an `as.dist()` coercion method for that.

The question is whether this is really needed. Is the number of zero cells caused by a removing distances below a minimum significant enough to warrant this solution? Current object format for an fcm is a sparse lower diagonal matrix. Current object format for dist is the same, except that it records small values too. What would the "sparsity" level of a typical dist return?

Finally, are there precedents for doing this with distance computations? Something whose practice we could emulate?

koheiw:
Nicolas Merz is facing this issue right now, so we can ask him to test if flooring helps. I also have a plant to perform huge pairwise distance calculation in near future for "fake news". Analysis of text reuse has been restricted due to the lack of an efficient tool, but the new argument might create a new trend in text analysis.

`wordspace::dist.matrix` returns a matrix, but can be coerced to a dist object.
https://www.rdocumentation.org/packages/wordspace/versions/0.2-0/topics/dist.matrix

[...]

PR #1451, opened 2018-10-12
Dev simil

koheiw:
Development of the core code for #1210 has been completed. Last thing is we should do is packaging.


koheiw:
I agree that this is the way, but we will make a full transition to the sparse matrix eventually. So we need warn users that the `dist`(or `simil`) is deprecated, and will be removed in near future.

koheiw:
It looks like this now
https://github.com/quanteda/quanteda/blob/6a02027c632e851fe64234e94a951b7abf651c6c/R/textstat_simil.R#L78-L93
I named the argument to switch the resulting object type `output`, because it is not the value that are affected. `textstat_simil()` sets a special class `simil`. I don't think there is no method for that, but kept only for backward compatibility.

koheiw:
I believe the new similarity/distance functions are ready to ship. Use of `dist` object for similarity is totally inappropriate (it does not record diagonals), so we have to urge users to switch to `sparse` quickly.

koheiw:
Thanks for reviewing @kbenoit. The C++ code always return `dgTMatrix` but we can use `dgCMatrix`. I didn't use the latter because compression takes sometime (only a few seconds, however). I can use `dsCMatrix` or `dgCMatrix`.

I will check revdev, leaving argument naming and documentation with you.

koheiw:
If we want to tell the users about important changes, we can print message as news when package is loaded.

kbenoit:
OK with `NULL` then, and let's restore the full set of dist attributes to the return value (which means only adding `method` I think).

kbenoit:
Also @koheiw can you pls check the `skip()`ed last test in the `test-textstat_proxy.R` since I was not clear under which conditions the sparse triangular versus symmetric matrix is returned. It might depend on conditions which is ol, but then I’d need to generalise the documentation which right now states dtCmatrix only.

koheiw:
If `dist` is an object class, `sparseMatrix` is more appropriate than `sparsematrix`. I think we have quite a few camel-cased argument values.

kbenoit:
I agree about `sparseMatrix`!  but I thought you would hate the camelCase. 😄

koheiw:
I hate mixing camels and snakes, but I tolerate camels as far as they are argument values, not names.

kbenoit:
OK all looking good now, except two more things. 

### 1. `diag = FALSE` when `value = `sparseMatrix`

For consistency, we should remove the diagonals in the second of these. Making it 0 sets it to `.` however and that's not really what we want, since this means zero in dist (correctly) but would mean 0 in simil (incorrectly). Better to make the diagonal NA if `diag = FALSE, value = "sparseMatrix")`.

```r
textstat_simil(test_mt, upper = TRUE, diag = FALSE, value = "dist")
#            2005-Bush 2009-Obama 2013-Obama 2017-Trump
# 2005-Bush             0.9142772  0.9366455  0.9432850
# 2009-Obama 0.9142772             0.9583689  0.9433585
# 2013-Obama 0.9366455  0.9583689             0.9431704
# 2017-Trump 0.9432850  0.9433585  0.9431704           

textstat_simil(test_mt, upper = TRUE, diag = FALSE, value = "sparse")
# 4 x 4 sparse Matrix of class "dsCMatrix"
#            2005-Bush 2009-Obama 2013-Obama 2017-Trump
# 2005-Bush  1.0000000  0.9142772  0.9366455  0.9432850
# 2009-Obama 0.9142772  1.0000000  0.9583689  0.9433585
# 2013-Obama 0.9366455  0.9583689  1.0000000  0.9431704
# 2017-Trump 0.9432850  0.9433585  0.9431704  1.0000000

(d <- textstat_dist(test_mt, upper = TRUE, diag = FALSE, value = "sparse"))
4 x 4 sparse Matrix of class "dsCMatrix"
           2005-Bush 2009-Obama 2013-Obama 2017-Trump
2005-Bush    .         73.55270   56.44466   56.75385
2009-Obama  73.55270    .         64.07808   73.23251
2013-Obama  56.44466   64.07808    .         46.87217
2017-Trump  56.75385   73.23251   46.87217    .      

diag(d) <- NA
d
# 4 x 4 sparse Matrix of class "dsCMatrix"
#            2005-Bush 2009-Obama 2013-Obama 2017-Trump
# 2005-Bush         NA   73.55270   56.44466   56.75385
# 2009-Obama  73.55270         NA   64.07808   73.23251
# 2013-Obama  56.44466   64.07808         NA   46.87217
# 2017-Trump  56.75385   73.23251   46.87217         NA
```

koheiw:
We definitely need `rank` because we not always have prior knowledge of the distribution of similarity, and optimal thresholds differ document to document. If we don't have `rank`, we need to rerun the function with different `min_smil` many times, taking several hours in each run.

People should explore sparse distance matrices by https://github.com/quanteda/quanteda/issues/1210#issuecomment-425683076, but not by coercing to `dist`. I though users can easily write functions to create a list from a matrix, but we can add `as.list()` if it is not the case.  `as.list.dgCMatrix` works not only with a distance matrix but also a dfm (good or bad).

kbenoit:
What if we integrate all the changes _except_ the `rank` for now, and discuss how to integrate it separately? We can leave underlying code in place to be reactivated. We should think more systematically about departing from a symmetric distance matrix since the possible return objects are too broad if we have a third now (or fourth). We can fix this later.

koheiw:
Hide `rank` from the users, but make it available via `...`?

kbenoit:
A solution to my objections would be to make it a global rank from the pairwise set, so that it could remain symmetric. If users want the top set within a document, they can use min or max with a threshold and then coerce to a list using the n. Would that work?

koheiw:
No. Global ranking is not only useless but also requires saving all the values. We should give up symmetry as there are asymmetric measures like Kulback. This is the main reason why we should abandon `dist` object as soon as possible, and the **proxy** package does not return that object type. `dist` is too constrained.

kbenoit:
That's what I thought (and can see from the C++ code).  We still have problems 2 and 3 above.

**proxy** does return a dist, with one more tag to indicate it's the identically formatted simil class.
```r
> proxy::dist(as.matrix(test_mt)) %>% class()
[1] "dist"
> proxy::simil(as.matrix(test_mt)) %>% class()
[1] "simil" "dist" 
> philentropy::distance(as.matrix(test_mt))
Metric: 'euclidean' using unit: 'log'.
         v1       v2       v3       v4
v1  0.00000 73.55270 56.44466 56.75385
v2 73.55270  0.00000 64.07808 73.23251
v3 56.44466 64.07808  0.00000 46.87217
v4 56.75385 73.23251 46.87217  0.00000
```
The last package has a ton of distance metrics BTW, and returns a symmetric (regular) matrix (which is easily coerced into a dist). dist remains a useful format because of the existing functions that use it (nearly all clustering methods) and the coercion methods to and from that exist. Its only disadvantage is that it cannot handle sparsity. For non-sparse distance measures, it's already efficient.

Coercion to dist from the dsCMatrix is also a bit problematic since it currently treats the blank cells as zero. And zeros mean very different things depending on the metric. I was thinking how to fix the sparse non-symmetric matrix returned with the current rank and realized that the problems run deeper. 

**HOWEVER** I realized my concerns are a bit overstated since we already return one non-symmetric object, the `dist_selection` we created for this. 
```r
> textstat_simil(test_mt, selection = c("2009-Obama", "2013-Obama"))
           2009-Obama 2013-Obama
2005-Bush   0.9142772  0.9366455
2009-Obama  1.0000000  0.9583689
2013-Obama  0.9583689  1.0000000
2017-Trump  0.9433585  0.9431704
> textstat_simil(test_mt, selection = c("2009-Obama", "2013-Obama")) %>% str()
 'dist_selection' num [1:4, 1:2] 0.914 1 0.958 0.943 0.937 ...
 - attr(*, "dimnames")=List of 2
  ..$ : chr [1:4] "2005-Bush" "2009-Obama" "2013-Obama" "2017-Trump"
  ..$ : chr [1:2] "2009-Obama" "2013-Obama"
 - attr(*, "Labels")= chr [1:2] "2009-Obama" "2013-Obama"
 - attr(*, "Size")= int 2
 - attr(*, "method")= chr "correlation"
 - attr(*, "call")= language textstat_simil.dfm(x = test_mt, selection = c("2009-Obama", "2013-Obama"))
> textstat_simil(test_mt, selection = c("2009-Obama", "2013-Obama"), rank = 3)
4 x 2 sparse Matrix of class "dgCMatrix"
           2009-Obama 2013-Obama
2005-Bush   .          .        
2009-Obama  1.0000000  0.9583689
2013-Obama  0.9583689  1.0000000
2017-Trump  0.9433585  0.9431704
> textstat_simil(test_mt, selection = c("2009-Obama", "2013-Obama"), rank = 3) %>% as.list()
[[1]]
[1] 0

[[2]]
[1] 1
...
```
So to make the rank function work, we would want to class it as `distsparse_selection` and make an `as.list()` method that excluded the blanks. We should also class the `dsCMatrix` version and make a special `as.list()` method for it as well. 

And think about what values the missing cells should take. We should make a list of the outstanding issues and split them up.

koheiw:
Making a new class `dist_sparse` is an option (we don't need `_selection` because `ncol(x) != nrow(x)` tells us about it) with `as.dist.dist_sparse()` and `as.list.dist_sparse()`.

koheiw:
If we don't find a good way to integrate the new function, I will make a separate package called **sparseproxy** or something and upload to CRAN. Only advanced users will use `textstat_proxy()` anyway. **quanteda** will call it internally.

koheiw:
Updated as per #1455 discussion.

kbenoit:
I'm seeing compiler warnings:
```
clang++ -std=gnu++11 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG -DARMA_DONT_PRINT_OPENMP_WARNING  -I"/Library/Frameworks/R.framework/Versions/3.5/Resources/library/Rcpp/include" -I"/Library/Frameworks/R.framework/Versions/3.5/Resources/library/RcppParallel/include" -I"/Library/Frameworks/R.framework/Versions/3.5/Resources/library/RcppArmadillo/include" -I/usr/local/include  -DARMA_64BIT_WORD=1 -fPIC  -Wall -g -O2 -c simil_mt.cpp -o simil_mt.o
simil_mt.cpp:105:11: warning: unused variable 'nrow' [-Wunused-variable]
    uword nrow = mt1.n_rows;
          ^
simil_mt.cpp:310:11: warning: unused variable 'nrow' [-Wunused-variable]
    uword nrow = mt.n_rows;
          ^
2 warnings generated.
```

kbenoit:
@koheiw This PR is ready to merge, but check that you are okay with 0e33de0 where I commented out the declarations for the unused `nrow` variables. (This could generate compiler warnings.)

:+1:

Issue #1447, opened 2018-10-07
tokens() options are inconsistent with what = "fasterword"


kbenoit:
### Different

`remove_punctuation`:
```r
tokens("This, a test.", what = "fasterword", remove_punct = TRUE)
## tokens from 1 document.
## text1 :
## [1] "This," "a"     "test."
```
and Twitter characters:
```r
tokens("@kenbenoit #quanteda", what = "fasterword", 
       remove_twitter = TRUE, remove_punct = TRUE)
## tokens from 1 document.
## text1 :
## [1] "@kenbenoit" "#quanteda" 

tokens("@kenbenoit #quanteda", what = "fasterword", 
       remove_twitter = FALSE, remove_punct = TRUE)
## tokens from 1 document.
## text1 :
## [1] "@kenbenoit" "#quanteda" 
```

`remove_separators`  behaves differently:
```r
txt <- "one two\nthree\tfour"

# unexpected
tokens(txt, what = "fasterword", remove_separators = FALSE)
## tokens from 1 document.
## text1 :
## [1] "one"              "two\nthree\tfour"

# correct
tokens(txt, what = "fasterword", remove_separators = TRUE)
## tokens from 1 document.
## text1 :
## [1] "one"   "two"   "three" "four" 

# correct
tokens(txt, what = "word", remove_separators = FALSE)
## tokens from 1 document.
## text1 :
## [1] "one"   " "     "two"   "\n"    "three" "\t"    "four" 

# correct
tokens(txt, what = "word", remove_separators = TRUE)
## tokens from 1 document.
## text1 :
## [1] "one"   "two"   "three" "four" 
```

### Same

`remove_url = TRUE` is ok:
```r
tokens("https://quanteda.io is our website", what = "fasterword", remove_url = TRUE)
## tokens from 1 document.
## text1 :
## [1] "is"      "our"     "website"

tokens("https://quanteda.io is our website", what = "fasterword", remove_url = FALSE)
## tokens from 1 document.
## text1 :
## [1] "https://quanteda.io" "is"      "our"     "website"
```

`remove_numbers` is ok:
```r
tokens("99 red balloons 4ever", what = "fasterword", remove_numbers = TRUE) %>%
    as.character()
## [1] "red"      "balloons" "4ever"
tokens("99 red balloons 4ever", what = "fasterword", remove_numbers = FALSE) %>%
    as.character()
## [1] "99"       "red"      "balloons" "4ever"  
```

`remove_symbols` is ok:
[...]

PR #1459, opened 2018-10-22
Address #1447

koheiw:
- Revert change in ad359e6fb12c164b1339bbb02b373d431a899a63.
- Fix handling of new line markers.
- Update documentation as well.
- Correct tests


kbenoit:
Thanks - I can see that this makes more sense now than the pre-fix behaviour (for `tokens(txt, what = "fasterword", remove_separators = FALSE)` output), but it basically means that with `what = "fasterword"`, the `separators` argument has no effect. Should we print a warning stating this if someone calls

```r
tokens(txt, what = "fasterword", remove_separators = FALSE)
```

Issue #1476, opened 2018-10-28
textmodel_wordscores performance across package versions


sjankin:
## Describe the bug

**textmodel_wordscores** produces different results across v1.2.0 and v1.3.(4). This possibly relates to handling of the smoothing parameter. Without smoothing parameter (defaulting to 0) prediction results are the same across versions. Specifying a smoothing parameter (Laplace or Jeffreys) produces different prediction results.

### v1.2.0

```r
packageVersion("quanteda")

(ws <- textmodel_wordscores(data_dfm_lbgexample, c(seq(-1.5, 1.5, .75), NA), scale = "linear"))
predict(ws)

[1] ‘1.2.0’

Call:
textmodel_wordscores.dfm(x = data_dfm_lbgexample, y = c(seq(-1.5, 
    1.5, 0.75), NA), scale = "linear")

Scale: linear; 5 reference scores; 37 scored features.
           R1            R2            R3            R4            R5            V1 
-1.317931e+00 -7.395598e-01 -8.673617e-18  7.395598e-01  1.317931e+00 -4.480591e-01 
```

### v1.3.4

```r

packageVersion("quanteda")

(ws <- textmodel_wordscores(data_dfm_lbgexample, c(seq(-1.5, 1.5, .75), NA), scale = "linear"))

predict(ws)

[1] ‘1.3.4’

Call:
textmodel_wordscores.dfm(x = data_dfm_lbgexample, y = c(seq(-1.5, 
    1.5, 0.75), NA), scale = "linear")

Scale: linear; 5 reference scores; 37 scored features.
           R1            R2            R3            R4            R5            V1 
-1.317931e+00 -7.395598e-01 -8.673617e-18  7.395598e-01  1.317931e+00 -4.480591e-01 
```

### v1.2.0

```r
packageVersion("quanteda")

(ws <- textmodel_wordscores(data_dfm_lbgexample, c(seq(-1.5, 1.5, .75), NA), scale = "linear", smooth = 1))

predict(ws)


[1] ‘1.2.0’

Call:
textmodel_wordscores.dfm(x = data_dfm_lbgexample, y = c(seq(-1.5, 
    1.5, 0.75), NA), scale = "linear", smooth = 1)

Scale: linear; 5 reference scores; 37 scored features.
           R1            R2            R3            R4            R5            V1 
-1.212047e+00 -6.952988e-01  2.222614e-18  6.952988e-01  1.212047e+00 -4.214674e-01 
```

### v1.3.4

[...]

PR #1480, opened 2018-10-30
Add tests and update Rd for #1476

kbenoit:
@koheiw does this behaviour seem correct to you? 

@patperry should the embedded x be smoothed or original? For methods that purport to recover the original estimates it might be unnatural to pass the unsmoothed object that differs from that on which the model was fitted.


koheiw:
I think `x` should be (unsmoothed) original, but it is inconvenient for the users if `predict` produces different results. There are two options:

1. Let users do smoothing outside textmodel_wordscores()`
2. Record smoothing value in the model and use the value in `predict()`

I prefer 1, but OK with 2.

kbenoit:
Not passing the smoothed `x` is consistent with how we do this in `textmodel_affinity()`, e.g.
```r
> (af <- textmodel_affinity(data_dfm_lbgexample, y = c("L", NA, NA, NA, "R", NA), smooth = 1))
Call:
textmodel_affinity.dfm(x = data_dfm_lbgexample, y = c("L", NA, 
    NA, NA, "R", NA), smooth = 1)

Training documents per class:L: 1, R: 1; total training features: 37
> head(af$x, nf = 20)
Document-feature matrix of: 6 documents, 20 features (50% sparse).
6 x 20 sparse Matrix of class "dfm"
    features
docs A B  C  D  E  F   G   H   I   J   K   L   M   N   O   P   Q   R   S   T
  R1 2 3 10 22 45 78 115 146 158 146 115  78  45  22  10   3   2   0   0   0
  R2 0 0  0  0  0  2   3  10  22  45  78 115 146 158 146 115  78  45  22  10
  R3 0 0  0  0  0  0   0   0   0   0   2   3  10  22  45  78 115 146 158 146
  R4 0 0  0  0  0  0   0   0   0   0   0   0   0   0   0   2   3  10  22  45
  R5 0 0  0  0  0  0   0   0   0   0   0   0   0   0   0   0   0   0   0   0
  V1 0 0  0  0  0  0   0   2   3  10  22  45  78 115 146 158 146 115  78  45
```

koheiw:
`textmodels_*()` should have smoothing argument only if it is integral part of the model (i.e. smoothing is not only pre-processing of the a dfm and you cannot apply `dfm_smonth()` outside of the function). In this view, `textmodel_wordscores()` and `textmodel_affinity()` are consistent in a bad way.

I always find it is difficult to maintain `textmodel_wordscores()` because of the strange variable names. If you do not want to change the function, you should remove it from **quanteda** and keep it as an "frozen" package in your account. The code of the function is not up to our standard.

patperry:
Three things you could potentially save:

1. The passed-in corpus or dfm (raw data)
2. the token counts for each of the reference classes (sufficient statistics)
3. the smoothed probability estimates for each reference class (parameter estimates)

I would argue that you should definitely save (3), and you should possibly save (2). Why? I think of the fitting methods as computing parameter estimates. Downstream analyses like predict take these estimates and apply them to new data.

kbenoit:
@koheiw's minority and inaccurate view of the `textmodel_wordscores()` code notwithstanding... 😲😲

@patperry The object already contains this information, although it does not save 2. separately from the original dfm that is passed through. The original question has to do with whether calling `predict()` on the fitted object should predict on the smoothed object (used to fit the model from the reference class token counts) or on the original, unsmoothed object. Right now both `textmodel_wordscores()` and `textmodel_affinity()` use the the original, unsmoothed object if `newdata = NULL` (predict default).

Question is whether this is the most appropriate behaviour in terms of the `x` that is passed.

If we recorded the smoothed reference counts, that could be a separate quantity, but what would we need that for, post-estimation?

PR #1489, opened 2018-11-12
Add missing and new lexical diversity measures

### LINKED ISSUE #1488: textstat_lexdiv does not support Yule's K despite being in documentation
## Describe the bug

Although the [documentation](https://quanteda.io/reference/textstat_lexdiv.html) of `textstat_lexdiv` suggests that Yule's K is supported as a lexdiv measure, it is actually not accepted as a possible value for the `measure` argument.

## Reproducible code
```r
> library(quanteda)
> library(data.table)
> test_dfm <-  corpus(c(d1 = "a a b b b c c d e f",
+                     d2 = "apple orange grape fruit french french") ) %>% dfm()
> 
> textstat_lexdiv(test_dfm, measure = 'K')
Error in textstat_lexdiv.dfm(test_dfm, measure = "K") : 
  Invalid measure(s): K
```

## Proposed Fix
The formula for Yule's K references this [paper](https://www.jstor.org/stable/30204534?seq=4#metadata_info_tab_contents).
The variable `M` is a vector of length `ndoc(x)` and ordered by `docnames(x)`.
Its computation relies on `quanteda::textstat_frequency` and `dplyr`. 

```r
 if ('K' %in% measure){
    
    # Yule's K is defined as 10^4 * (M-N)/N^2 where M = V(i) * i^2, with
    # V(i) being the number of types with exactly i counts, and 
    # N being the number of tokens
    
    M <- textstat_frequency(x, groups = docnames(x)) %>%  
      group_by(group, frequency) %>% 
      count() %>% ungroup() %>% 
      group_by(group) %>%
      summarize(temp_M = sum(frequency^2 * n)) %>% 
      select(temp_M)

    temp[, K := 10^4 * (M - n_tokens) / (n_tokens^2)]
  }
```
After adding this chunk to the current version of `textstat_frequency` and adding `'K'` to the list of accepted values for `measure`, the same test case was run. 

```r
> textstat_lexdiv_withYuleK(test_dfm, measure = 'K')
  document         K
1       d1 1000.0000
2       d2  555.5556
```

kbenoit:
Add's Yule's _K_, previously mentioned in the documentation but not implemented.  Also add's Simpson's _D_ and Herdan's _Vm_, both from Tweedie & Baayen (1998).

Fixes #1488.

@jiongweilua I implemented this without using **dplyr** since we have tried to avoid the overhead of importing the tidyverse (our package is big enough already!). I also tried to add tests from the literature to match (approximately - see the test notes) results from the sources. This needs a bit more work.

We also could use some checking for the measures re-implemented from Michalke's **koRpus** package since many (e.g. _S_) have no tests and no references.


R/textstat_lexdiv.R
```diff
@@ -91,26 +106,26 @@
...
 #' @export
-textstat_lexdiv.default <- function(x, measure = c("all", "TTR", "C", "R", "CTTR", "U", "S", "Maas"), 
+textstat_lexdiv.default <- function(x, measure = c("all", "TTR", "C", "R", "CTTR", "U", "S", "K", "Vm", "Maas"), 
                                 log.base = 10) {
     stop(friendly_class_undefined_message(class(x), "textstat_lexdiv"))
 }
 
 #' @export
-textstat_lexdiv.dfm <- function(x, measure = c("all", "TTR", "C", "R", "CTTR", "U", "S", "Maas"), 
+textstat_lexdiv.dfm <- function(x, measure = c("all", "TTR", "C", "R", "CTTR", "U", "S", "K", "Vm", "Maas"),
```
jiongweilua: This line omitted 'D' as an accepted argument for `measure`

R/textstat_lexdiv.R
```diff
@@ -91,26 +106,26 @@
...
-textstat_lexdiv <- function(x, measure = c("all", "TTR", "C", "R", "CTTR", "U", "S", "Maas"), 
+textstat_lexdiv <- function(x, measure = c("all", "TTR", "C", "R", "CTTR", "U", "S", "K", "Vm", "Maas"), 
                             log.base = 10) {
     UseMethod("textstat_lexdiv")
 }
 
 
 #' @export
-textstat_lexdiv.default <- function(x, measure = c("all", "TTR", "C", "R", "CTTR", "U", "S", "Maas"), 
+textstat_lexdiv.default <- function(x, measure = c("all", "TTR", "C", "R", "CTTR", "U", "S", "K", "Vm", "Maas"),
```
jiongweilua: This line omitted 'D' as an accepted argument for `measure`

R/textstat_lexdiv.R
```diff
@@ -91,26 +106,26 @@
 #' (result <- textstat_lexdiv(mydfm, c("CTTR", "TTR", "U")))
 #' cor(textstat_lexdiv(mydfm, "all")[,-1])
 #' 
-textstat_lexdiv <- function(x, measure = c("all", "TTR", "C", "R", "CTTR", "U", "S", "Maas"), 
+textstat_lexdiv <- function(x, measure = c("all", "TTR", "C", "R", "CTTR", "U", "S", "K", "Vm", "Maas"),
```
jiongweilua: This line omitted 'D' as an accepted argument for `measure`

kbenoit: Good 👁 @jiongweilua. Travis fails in the CI only because you did not also rebuild the documentation (.Rd) files. Best to have these options checked so you don't have to try to remember to do it manually.
screenshot 2018-11-13 09 08 28

koheiw:
I hope you to find a way to implement without **data.table** but look fine to me, otherwise. I have no idea about the new measures.

Issue #1491, opened 2018-11-16
CRAN failures on Linux


kbenoit:
The `textstat_proxy()` tests are failing and need to be fixed urgently, and the package resumitted to CRAN.  See https://cran.r-project.org/web/checks/check_results_quanteda.html.

```
checking tests ... [141s/146s] ERROR
  Running ‘testthat.R’ [141s/146s]
Running the tests in ‘tests/testthat.R’ failed.
Complete output:
  > Sys.setenv("R_TESTS" = "")
  > 
  > library(testthat)
  > library(quanteda)
  Package version: 1.3.13
  Parallel computing: 2 of 32 threads used.
  See https://quanteda.io for tutorials and examples.
  
  Attaching package: 'quanteda'
  
  The following object is masked from 'package:utils':
  
      View
  
  > 
  > Sys.setenv("_R_CHECK_LENGTH_1_CONDITION_" = TRUE)
  > 
  > ops <- quanteda_options()
  > quanteda_options(reset = TRUE)
  > 
  > test_check("quanteda")
  ── 1. Failure: record zeros even in the sparse matrix (@test-textstat_proxy.R#35
  textstat_proxy(mt)@x not equal to c(1, 0, 1).
  2/3 mismatches (average diff: 1)
  [1] 0 - 1 == -1
  [2] 1 - 0 == 1
  
  ── 2. Failure: record zeros even in the sparse matrix (@test-textstat_proxy.R#36
  textstat_proxy(mt, method = "cosine")@x not equal to c(1, 0, 1).
  2/3 mismatches (average diff: 1)
  [1] 0 - 1 == -1
  [2] 1 - 0 == 1
  
  ── 3. Failure: record zeros even in the sparse matrix (@test-textstat_proxy.R#36
  textstat_proxy(mt, method = "cosine", min_proxy = -0.5)@x not equal to c(1, 0, 1).
  2/3 mismatches (average diff: 1)
  [1] 0 - 1 == -1
  [2] 1 - 0 == 1
  
  ── 4. Failure: record zeros even in the sparse matrix (@test-textstat_proxy.R#36
  textstat_proxy(mt, method = "cosine", rank = 2)@x not equal to c(1, 0, 1).
  2/3 mismatches (average diff: 1)
  [1] 0 - 1 == -1
  [2] 1 - 0 == 1
  
  ── 5. Failure: record zeros even in the sparse matrix (@test-textstat_proxy.R#36
[...]

Issue #1455, opened 2018-10-22
Sparse distance matrix assignment of zeroes is misleading or wrong


kbenoit:
This is a substantive issue to consider before we accept PR #1451.

The goal of using a sparse matrix format to represent distances or similarities is a good one: to avoid huge n^2 dense objects when the number of distance nodes n is large. 

The implementation via sparse matrices, however, does not make sense in many cases, since these are treated as zeroes, rather than just being ignored.

### 1. It makes no sense for metrics that are signed, such as the Pearson's correlation default for `textstat_simil()`.  Consider this example:

```r
library("quanteda")
library("Matrix")

testdfm <- as.dfm(
    matrix(c(1, 2, 0, 5, 6,
             1, 3, 2, 6, 5,
             1, 2, 0, 5, 5, 
             2, 2, 0, 5, 6,
             6, 5, 0, 2, 1),
           nrow = 5)
) %>% 
    dfm_weight("prop")

textstat_simil(testdfm, method = "correlation") %>% 
    as.matrix()
#            text1       text2       text3      text4      text5
# text1  0.0000000  0.90212515 -0.30942637 -0.9580710 -0.9231344
# text2  0.9021252  0.00000000  0.08574929 -0.8091548 -0.9616447
# text3 -0.3094264  0.08574929  0.00000000  0.5160468  0.1078328
# text4 -0.9580710 -0.80915478  0.51604685  0.0000000  0.8903482
# text5 -0.9231344 -0.96164474  0.10783277  0.8903482  0.0000000

textstat_simil(testdfm, method = "correlation", min_simil = -0.5)
# 5 x 5 sparse Matrix of class "dsCMatrix"
#            text1      text2       text3     text4     text5
# text1  1.0000000 0.90212515 -0.30942637 .         .        
# text2  0.9021252 1.00000000  0.08574929 .         .        
# text3 -0.3094264 0.08574929  1.00000000 0.5160468 0.1078328
# text4  .         .           0.51604685 1.0000000 0.8903482
# text5  .         .           0.10783277 0.8903482 1.0000000
```

Here, there are some strong negative correlations. Setting `min_simil = -0.50` means that:
- the strongest negative correlations become zero
- the strongest positive correlations remain as is
- the weakest negative and positive correlations remain as is

Setting `min_simil = 0.50` means that all except the strongest positive correlations become zero.

`method = "cosine"` can also range from [0, 1]. Although this will not happen with word counts, it could happen with other (negatively valued) data.

[...]

PR #1492, opened 2018-11-17
Fix test to make it independent of the order of x values

koheiw:
#1491 was puzzling because it does not happen on any of my Linux machines, but the answer was simple: TBB works a bit differently on the CRAN machine with Xeon CPU.


koheiw:
I should have explain to you better, but this is what your stability test should look like. The problem was that the `x` values looks random, when you ignore `i` and `j` (they change correspondingly). However it is not, when it is converted to a dense matrix. 
```r
result <- data.frame()
colnames(result) <- paste0("v", 1:4)
for (i in seq_len(n)) {
    out <- quanteda:::qatd_cpp_similarity_linear(t(mt), 1, seq_len(ndoc(mt)), ndoc(mt), -1)
    result <- rbind(result, as.numeric(as.matrix(out)))
}
unique(result)
#   v1 v2 v3 v4
# 1  1  0  0  1
```

kbenoit:
Ok... But we can fix that in the output. The i and j values always sorted in address order in a sparse T matrix, so we should so the same.

There is still the problem 1 above though.

koheiw:
On problem 1, it is `dist` object's problem: do not record diagonals...
```r
toks <- tokens(c(doc1 = 'a b c', doc2 = 'd e f'))
mt <- dfm(toks)

textstat_simil(mt, method = "cosine") %>% as.matrix()
#      doc1 doc2
# doc1    0    0
# doc2    0    0
textstat_proxy(mt, method = "cosine") %>% as.matrix()
#      doc1 doc2
# doc1    1    0
# doc2    0    1
```
`dgTMatrix` is perfectly fine without sorting. I could sort but there might be performance penalty.

```r
> tri <- sparseMatrix(i = c(1,3:8), 
+                     j = c(9,2,6:10), 
+                     x = 7 * (1:7), giveCsparse = FALSE)
> 
> tri
8 x 10 sparse Matrix of class "dgTMatrix"
                              
[1,] .  . . . .  .  .  .  7  .
[2,] .  . . . .  .  .  .  .  .
[3,] . 14 . . .  .  .  .  .  .
[4,] .  . . . . 21  .  .  .  .
[5,] .  . . . .  . 28  .  .  .
[6,] .  . . . .  .  . 35  .  .
[7,] .  . . . .  .  .  . 42  .
[8,] .  . . . .  .  .  .  . 49
> tri@x
[1]  7 14 21 28 35 42 49
> 
> tri_rev <- sparseMatrix(i = rev(c(1,3:8)), 
+                         j = rev(c(9,2,6:10)), 
+                         x = rev(7 * (1:7)), giveCsparse = FALSE)
> 
> 
> tri_rev
8 x 10 sparse Matrix of class "dgTMatrix"
                              
[1,] .  . . . .  .  .  .  7  .
[2,] .  . . . .  .  .  .  .  .
[3,] . 14 . . .  .  .  .  .  .
[4,] .  . . . . 21  .  .  .  .
[5,] .  . . . .  . 28  .  .  .
[6,] .  . . . .  .  . 35  .  .
[7,] .  . . . .  .  .  . 42  .
[8,] .  . . . .  .  .  .  . 49
> tri_rev@x
[1] 49 42 35 28 21 14  7
> 
```

kbenoit:
OK, I get it now, but will raise a separate issue about as.matrix.dist() and what we should return when `diag = TRUE`.

koheiw:
the `as.matrix.dist()` issue would be difficult one :grimacing:

kbenoit:
`as.matrix.dist()` is fine, it's the `as.matrix.simil()` that needed different handling.

`dsTMatrix` as a format is superior to `dist` and I agree we should make the switch, *but*:

1.  we need a way to handle the diagonal for the similarity types, same as #1494, since this is wrong:
```r
textstat_proxy(mt, method = "cosine") %>% as.dist(diag = TRUE)
#      doc1 doc2
# doc1    0     
# doc2    0    0
```

2. zeroes should be real zeros, not items we dropped because of thresholds. I still see no solution to #1455 using sparse matrix objects for this.

Anyway @koheiw please take a look at my fix and merge if you approve.

Issue #1494, opened 2018-11-17
coercion to matrix of textstat_simil return zeros the diagonals


kbenoit:
Our method for `as.matrix.dist()` is from **base**, but we have our own `quanteda::as.matrix.simil()`. I suggest we add **proxy** to the Imports and use its `as.matrix.dist()` and `as.matrix.simil()` functions. The main difference is how it treats the diagonals. From `?proxy::as.matrix.simil`:

> `diag`: In the context of `as.matrix` the value to use on the diagonal representing self-proximities. In case of similarities, this defaults to `NA` since a priori there are no upper bounds, so the maximum similarity needs to be specified by the user.

```r
set.seed(10)
x <- matrix(rnorm(16), ncol = 4)

## should be the same: OK
proxy::simil(x)
#             1           2           3
# 2 -0.13028526                        
# 3  0.03843595  0.37908734            
# 4  0.90753240  0.19031358  0.41521589
textstat_simil(as.dfm(x))
#             text1       text2       text3
# text2 -0.13028526                        
# text3  0.03843595  0.37908734            
# text4  0.90753240  0.19031358  0.41521589

## should be the same: FAIL
proxy::simil(x) %>% quanteda:::as.matrix.simil()
#             1          2          3         4
# 1  1.00000000 -0.1302853 0.03843595 0.9075324
# 2 -0.13028526  1.0000000 0.37908734 0.1903136
# 3  0.03843595  0.3790873 1.00000000 0.4152159
# 4  0.90753240  0.1903136 0.41521589 1.0000000
proxy::simil(x) %>% proxy:::as.matrix.simil()
#             1          2          3         4
# 1          NA -0.1302853 0.03843595 0.9075324
# 2 -0.13028526         NA 0.37908734 0.1903136
# 3  0.03843595  0.3790873         NA 0.4152159
# 4  0.90753240  0.1903136 0.41521589        NA
## need to add diag argument to as.matrix.simil
proxy::simil(x) %>% proxy:::as.matrix.simil(diag = 1.0)
#             1          2          3         4
# 1  1.00000000 -0.1302853 0.03843595 0.9075324
# 2 -0.13028526  1.0000000 0.37908734 0.1903136
# 3  0.03843595  0.3790873 1.00000000 0.4152159
# 4  0.90753240  0.1903136 0.41521589 1.0000000

## should be the same:  FAIL
textstat_simil(as.dfm(x)) %>% quanteda:::as.matrix.simil()
[...]

PR #1495, opened 2018-11-17
Issue 1494

kbenoit:
Distinguishes `textstat_simil()` similarity objects from `dist` objects so that the coercion to matrix defaults to 1.0 for self-similarities. The previous approach of classing them as `dist` meant that the diagonal became 0.0 which is exactly the opposite of what it should be.

Solves #1494


kbenoit:
That's a perfectly valid point. I investigated why, and it's stated in `proxy::simil.Rd:

> Distance measures can be used with simil, and similarity measures with dist. In these cases, the result is transformed accordingly using the specified coercion functions (default: pr\_simil2dist(x) = 1 - abs(x) and pr\_dist2simil(x) = 1 / (1 + x)). Objects of class simil and dist can be converted one in another using as.dist and as.simil, respectively.

But this could all be achieved through coercion methods, e.g. `as.dist()`, `as.simil()`, if we wanted.

I think the most natural output for each is a specially classed `dsTMatrix`, but with `NA` rather than `.`/0 for the values not computed. For people who don't need the matrix format at all and want to avoid the space required to store `NA`, we should either:

a.   give an `output` option with multiple choices such as `output = c("dsTMatrix", "dist", "list", "data.frame")`. This is a bit confusing however to have the return class as an argument. C/Java programmers would look away in horror 😨  ; or

b.  output `dsTMatrix`, and provide `as.dist()`, `as.list()`, `as.data.frame()` methods to coerce it. And with a., we would need these anyway. However this does not avoid the dense state before output to a sparse equivalent. So we might want to go straight from the C++ return to a list or data.frame of pairs, as a more limited option from a.

**However** this PR simply corrects something wrong in our existing approach, and for us to update CRAN I think we should accept it and then plan the revision of the dist formats and integration of a more direct link to the `textstat_proxy()` outputs for the next release.

Issue #1500, opened 2018-11-19
Need (internal?) tokens_split()


kbenoit:
This would resolve the inefficiency noted [here](https://github.com/quanteda/quanteda/issues/1445#issuecomment-440066936) but with a general solution. I actually started writing this based on the (internal) `tokens_segment()` but then went for a quicker fix since I only needed an internal hyphen solution for 
https://github.com/quanteda/quanteda/blob/0dc5b345b4370ca8bc90c8cd0ecd6e4eb687e995/R/tokens.R#L221-L225

Proposal:
```r
tokens_split(x, pattern, valuetype = c("glob", "regex", "fixed"),
  case_insensitive = TRUE, keep_pattern = FALSE)
```
which would be the opposite direction, otherwise similar behaviour, to `tokens_compound()`.

kbenoit:
BTW here I implemented a hyphen-specific version for dfm features [here](https://github.com/quanteda/quanteda/blob/master/R/textstat_lexdiv.R#L283), probably about as efficiently it's possible to do in straight R.

But it's simpler for a dfm easier than for tokens because a) can use column indexing from matrices and b) feature order does not matter for dfm.

kbenoit:
To Do: Reimplement the inefficient solution to #1498 once `tokens_split()` is working.

kbenoit:
Solved in #1520 via `tokens_chunk()`.

PR #1502, opened 2018-11-24
Dev tokens split

koheiw:
This is not passing the test because of the errors potentially relaited to #1285, but the `tokens_split()` and `tokens_replace()` are working. Can @kbenoit check the issue?

If `network::print.summary.character()` is the root cause we should ask the developer to remove the function.

This branch contains a lot of changes because I needed to upgrade `pattern2id` to add multi-multi substitution capability to `tokens_replace()`. `tokens_split()` is relatively simple wrapper to   `tokens_replace()`.


kbenoit:
Sure, I will investigate shortly.

tests/testthat/test-dictionaries.R
```diff
@@ -383,3 +383,42 @@ test_that("dictionary merge values in duplicate keys", {
...
+  
+  ids4 <- quanteda:::pattern2list(dict4, type, "glob", FALSE)
+  expect_identical(unique(names(ids4)), names(dict4))
+})
+
+
+test_that("split_values() handle concatenators correctly", {
+  skip("To be fixed shortly")
+  test_that(
+    quanteda:::split_values(list(A = "a_a", "b_b"), "_", " "),
```
kbenoit: Does this return value make sense? Not really to me, returning two lists after splitting a single-element (top-level) list by a single character.

R/dictionaries.R
```diff
@@ -574,37 +592,48 @@ list2dictionary <- function(dict) {
...
 
-# Import a Lexicoder dictionary
-# dict <- read_dict_lexicoder('/home/kohei/Documents/Dictionary/Lexicoder/LSDaug2015/LSD2015.lc3')
+#' Import a Lexicoder dictionary
+#' 
+#' @param path the path from which to read in the Lexicoder dictionary
+#' @keywords dictionary internal
+#' @examples
+#' \dontrun{
+#' # dict <- read_dict_lexicoder("/home/kohei/Documents/Dictionary/Lexicoder/LSDaug2015/LSD2015.lc3")
```
kbenoit: How about we add these files to `extdata` or somewhere that they can be accessed if a user wants to try them? (And when running `devtools::run_examples()`? I will do the same for my local file references!

koheiw:
On the API design, your suggestion in #1500, looks like 
```r
tokens_segment <- function(x, pattern,
                           valuetype = c("glob", "regex", "fixed"),
                           case_insensitive = TRUE, 
                           extract_pattern = FALSE,
                           pattern_position = c("before", "after"),
                           use_docvars = TRUE) {
}
```
but I think `tokens_split()` will be used for post-processing like `tokens.tokens()`, so `separator` sounds better to me. 

On `glob`, I could not find reason to support that value type here. That is too loose as a patter to split strings. I am even thinking to remove this value type from `corpus_segment()`.

kbenoit:
OK, so let's think of it this way: We have complementary, opposite functions in these pairs:

* `*_segment()`, `*_group()`: These split or merge objects by redefining the _document_ units.
* `*_split()`, `*_compound()`: These split or merge tokens (or features) by redefining the token (or feature) values. These make sense for tokens, but also the approach I used here (for hyphens only) https://github.com/quanteda/quanteda/blob/dev-tokens_split/R/textstat_lexdiv.R#L283 could be generalized for dfm features too, although I not suggesting (right now, anyway) that we generalize expose that function, and since `dfm_compound()` could not be implemented.

So if split and compound are the natural opposites, then it suggests we match the arguments of `tokens_split()` as closely as we can to `tokens_compound()`.

Use cases?

1. Remove internal patterns. This is where I think the glob is (still!) useful, but we would want to disallow single-character patterns that contain a wildcard - so no `"*"` or `"?"`. Otherwise, what's wrong with them?
```r
toks <- tokens("London-Manchester-Dublin Amsterdam-Frankfurt-London")
tokens_split(toks, pattern = "-*-", valuetype = "glob", keep_pattern = FALSE)
# tokens from 1 document.
# text1 :
#  [1] "London"     "Dublin"     "Amsterdam"  "Frankfurt"  "London"  
```
    This is admittedly a very specific and likely, rare use case.

2. To perform, ex post, the `remove_hyphens = TRUE` function that started this thread.

```r
tokens_split(toks, pattern = "-", valuetype = "fixed", keep_pattern = TRUE)
# tokens from 1 document.
# text1 :
#  [1] "London"     "-"          "Manchester" "-"          "Dublin"     "Amsterdam"  "-"         
#  [8] "Frankfurt"  "-"          "London" 
```

3. To reverse `tokens_compound()`.

```r
tokens_compound(
    tokens("multi word expression"), 
    pattern = phrase("multi word expression"), 
    concatenator = "_"
) %>%
    tokens_split("_", valuetype = "fixed", keep_pattern = FALSE)
# tokens from 1 document.
# text1 :
# [1] "multi"      "word"       "expression"
```

Right now, `tokens_split()` can do all but the first. I've used my proposed argument names rather than the ones currently in `tokens_split()`, obviously. My argument for the pattern, valuetype is to continue the very nice consistency we have adopted in using these across the other functions, and to make it complementary to `tokens_compound()`.

kbenoit:
Also we have an edge case failure if the pattern completely matches a token:
```r
toks <- tokens("five-star rating", remove_hyphen = TRUE)
toks
# tokens from 1 document.
# text1 :
# [1] "five"   "-"      "star"   "rating"
tokens_split(toks, "-", "fixed", TRUE)
# Error in if (pattern == "") { : argument is of length zero
# In addition: Warning message:
# In max(stri_length(unlist(pattern, use.names = FALSE))) :
#  
#  Error in if (pattern == "") { : argument is of length zero  
```

Pleased to see that this works as expected:
```r
tokens_split(tokens("One- and two-star ratings -hyphen", what = "fasterword"), "-", "fixed", FALSE)
# tokens from 1 document.
# text1 :
# [1] "One"     "-"       "and"     "two"     "-"       "star"    "ratings" "-"       "hyphen" 
```

koheiw:
I think `*_split()` should have similar API to `*_segment()` than to `*_compound()` because the task is similar. As I mentioned earlier, how glob pattern is treated in `corpus_segment()` makes me feel terribly bad :disappointed:. This is why I think we need some thing different from `pattern`.

If we try to make `tokens_split()` similar to `tokens_compound()`, `concatenator` (or `separator`) is what it needs, not `pattern`.

kbenoit:
Ok, I can live with that for token_split(). Let’s keep it as you’ve done. 

You’re right that `separator` is a good antonym for `concatenator`. Should the default be space or `_`?

But what’s wrong with glob? (Here or in corpus_segment?) if you have an issue with glob in corpus_segment we should open a new issue to discuss. 

I’ll focus on the test failure shortly.

R/dictionaries.R
kbenoit: I'll make this a separate PR.

tests/testthat/test-dictionaries.R
kbenoit: See 118142f and my comments on that commit.

R/dictionaries.R
kbenoit: Moved to #1507.

tests/testthat/test-dictionaries.R
kbenoit: It makes sense to me now - although I changed the test expectation to not expect a name for the second pair of return elements, since none is specified for the `"b_b"`.

kbenoit:
I've approve the PR but:

- I see the problem with glob patterns clearly now, but also think this applied to regex matches, which are indeterminate unless we restrict them to a single character. Should we allow multi-character `separator` patterns? If not, then we really are saying we allow fixed or charclass matches. Since we don't have those implemented (charclass I mean) and we are not checking regex expansion length, I propose we just restrict valuetype to "fixed". We can keep the argument for now, in case we expand it later, but fixed should be enough for the immediate needs. A future feature request to expand the valuetype for this function could address the trickier issues of multi-character separator length.

- 118142f was passing but then I added a breaking test in 2e017b7 that I would have fixed myself, except that it involves the `pattern2id()` / `search_fixed()` functions that you know better than I, so I will let you trap and fix this. (Probably trapping the condition before calling these is better but you might want to fix it more fundamentally.)

Then it's ready to merge IMHO.

kbenoit:
That makes a lot of sense, and then we could enforce the single-character resolution. It would apply not as an additional argument to `valuetype` but as an option for single-character pattern matches such as `separator` or `concatenator`.

koheiw:
Do you want to extend the `valuetype`'s possible values to be `c("glob", "fixed", "regex", "charclass")`?

kbenoit:
That’s one option, but then we would not include the `charclass` option on most usages of `pattern` since very few are expecting the pattern to resolve to a single character, which is how I think of charclass versus a regex that contains a character class. 

We could document this option however in the `valuetype.Rd` man page but explains there that not all arguments are available with every function.

So for this function it would be
```r
tokens_split(x, separator, valuetype = c(“fixed”, “charclass”), ...)
```
that makes sense. And we would exclude multi-character separator patterns, so you can remove the tests that I added but that are currently skipped.

However - can we think of any other functions for which we would allow a `valuetype = “charclass”`?

koheiw:
Sounds good. `tokens_segment()` and `corpus_segment()` should use `separator`.

kbenoit:
Let’s leave the segment commands to another discussion, since I think they need multi-character patterns, and find the existing `corpus_segment()` functionality very useful as is (including with globs!). But I think we’ve come to a good conclusion on `tokens_split()`.

koheiw:
Agreed on `tokens_split()`. I will resurface the idea of `corpus_extract()` and `tokens_extract()` in other place...

PR #1496, opened 2018-11-17
textstat_lexdiv: Support for tokens objects, and options for removal of punctuation, symbols, numbers & hyphens.

jiongweilua:
References #1487 

- Added support for `x` to be a tokens object
- Added functionality to remove punctuation, symbols and numbers based on regex methods consistent with `tokens_select` and `dfm_remove`
- Created internal function `compute_lexdiv_stats` for computation of lexical diversity statistics

Will work on creating support for `remove_hyphens` separately.

Issue #679, opened 2017-04-21
corpus_segment by fixed token lengths


kbenoit:
We should add an option to segment documents by fixed word length segments.

lmullen:
I'm not sure if this is suitable for your purposes, but there will be a function to do this in the next release of tokenizers. It's not quite finished, but it's functional. 

https://github.com/ropensci/tokenizers/blob/master/R/chunk-text.R#L17

koheiw:
I believe this should be `tokens_segment()` instead of `corpus_segment()`, because we never know the accurate number of tokens until tokening texts. For `corpus_segment()`, it should be the number of characters (this can be done with regex actually, `.{1,100}` for 100 characters).

kbenoit:
But `tokens_anything()` would need to input tokens and output tokens. I don't think `tokens_segment()` would do that.

koheiw:
It will, taking tokens and return reshaped tokens.

kbenoit:
Yes that could work. It causes us to need to think carefully about what we allow a "token" to be, since currently we have a `what = "sentence"` option that allows a "token" to be a sentence. How would `tokens_segment()` work with sentence "tokens"?

koheiw:
tokens_segment will return groups of sentences when a fixed length is given, but I am not sure who will use it in this way...

koheiw:
I think `coups_reshape()` can be implemented as `corpus_segment()` + `corpus_group()` more consistently with other functions, because reshaping from sentences to documents is grouping corpus by original document IDs.

kbenoit:
Agreed - I was thinking that `corpus_segment` would become an renamed internal function, with the exported `corpus_segment` a deprecated function. `corpus_reshape` would work as it does not, but reply on the internal `corpus_segment` that would also use `corpus_group`. 

Is this what you meant?

koheiw:
My idea is to keep `corpus_segment()` (largely) unchanged and create `corpus_group()`. We will remove `corpus_reshape()`, because there is no reshape functions for tokens and dfm. We can export `tokens_group()` to complete the set.

kbenoit:
But group only aggregates. Reshape allows going up or down.

koheiw:
From documents to sentences use `corpus_segment()`
From sentences to documents use `corpus_group()`

kbenoit:
[...]

PR #1508, opened 2018-11-27
Introduce support for for MATTR, MSTTR, MTLD, vocd-D, hd-D in textstat_lexdiv.tokens

jiongweilua:
The purpose of this PR is to add support for three of the five "moving measures" discussed in #1496  to `textstat_lexdiv.tokens()`. They are:

- MSTTR:  Mean Segmental Type-Token Ratio 
- MATTR: Moving-Average Type-Token Ratio (Covington & McFall, 2010)
- MTLD: Moving-Average Measure of Textual Lexical Diversity  (McCarthy & Jarvis, 2010) 

The high level change in the code architecture is as follows:
- `compute_mattr`, `compute_msttr` and `compute_mtld` were generated as helper functions that compute the corresponding statistics at the token level
- these helper functions are called within `textstat_lexdiv.tokens()` 
- as these new measures are only supported for `tokens` objects, and the existing `textstat_lexdiv.tokens()` relies on `textstat_lexdiv.dfm`, I also added a handler to split the `measures` into static and dynamic measures. 

In terms of the testing files:
- To separate tests for the previously existing code logic, and the code for the new measures, I also restricted measures for existing textstat_lexdiv tests to only static measures. 

@kbenoit I think 1) the new helper functions I introduced are fairly modular and robust, 2) I have covered the most obvious bases for exception handlers and error catchers, and 3) the tests for the three measures are also fairly representative. But there are quite a lot of moving parts and changes so I thought it would be good to have your input right now on this build (it passed the various CI checks) before it gets messier!

Thanks! :)


koheiw:
I though about what the best way to construct a ngram-feature matrix earlier (documents are ngrams). We could make a new C++ function similar to one for `fcm()`, but I have a feeling that we can achieve that simply by duplicating and adjusting `p` and `i` vectors when we construct a DFM.

R/tokens_segment.R
```diff
@@ -98,3 +98,66 @@ tokens_segment.tokens <- function(x, pattern,
...
+#'                   endeavor to express the high sense I entertain of this
+#'                   distinguished honor.")
+#' toks <- tokens(txts)
+#' tokens_chunk(toks, size = 5)
+tokens_chunk <- function(x, size, discard_remainder = TRUE) {
+    UseMethod("tokens_chunk")
+}
+
+#' @export
+tokens_chunk.tokens <- function(x, size, discard_remainder = TRUE) {
```
koheiw: I think I can do this more efficiently by `tokens_segment()`.

kbenoit: No doubt! `tokens_chunk()` is already built on `tokens_segment()` after inserting a value on which to segment the text into regular-sized "chunks". It's exactly like `tokens_segment()` but with the segments defined by length, rather than by delimiter. The function to do the insertion is called (unimaginatively, but descriptively) `insert_values()` in `tokens_segment.R`. It's pretty efficient for R. So it may not be worth modifying `tokens_segment()` further at this point for what is probably a very specific use-case.

But then maybe not. I can see where we could certainly want the ability to reshape a corpus or a tokens object into fixed-length texts or tokens, rather than segmenting them on a pattern.

koheiw: We had discussion on this here https://github.com/quanteda/quanteda/issues/679. If it allows overlapping of tokens, there must be use cases (e.g. automated audio transcripts that do not have punctuation at all!).

For example, `list(c("a", "b", "c", "d")` will be
```
list(c("a", "b"), c("c", "d"))
```
when `length = 2` and `overlap = FALSE`, 

```
list(c("a", "b"), c("b", "c"), c("c", "d"))
```
when `length = 2` and `overlap = TRUE`. This also solves the moving window problem.

kbenoit: True! I'd forgotten. We also had a lot of discussion about the segment versus reshape options, and went with `corpus_segment()` and `tokens_segment()` operating on a pattern match, and `corpus_reshape()` being limited to redefining documents based on syntactical divisions. `corpus_segment()` does not need to be tokenized first because it looks for a pattern in the text, rather than a pattern match to tokens.

I think it's clear to us that:
- we need a function similar to what in this PR is called `tokens_chunk()`
- the addition of the `overlap` option is an interesting idea, although I am not sure what would be the use case
- `tokens_chunk()` makes sense but `corpus_chunk()` does not, unless it operates only on characters (since it has no "tokens" until it has been tokenized)

I dislike the idea of building a "chunk_size" argument into `tokens_segment()` because it would be incompatible with the pattern argument, which is currently a required argument, and because then it would behave differently from `corpus_segment()` that also splits on patterns.

I'm not exactly wild about `tokens_chunk()` as a name although that's pretty much an accurate description, and I cannot think of anything better right now.

koheiw: Then, we can add `tokens_chunk()` as an experimental function. `overlap` can be used, for example, to detect text reuse:
```r
tokens_chunk(toks, size = 10, overlap = TRUE) %>%
dfm() %>% 
textstat_proxy(min_prox = 0.9)
```

koheiw:
Please use C++ `tokens_chunk()` in  #1520

kbenoit:
The moving average measures are going to take more work and a lot of optimization and testing, so rather than hold up the changes already implemented in this PR, I moved the MTLD etc development to a new branch and stripped them from the `MATTR` branch so we can merge this one.

PR #1515, opened 2018-12-07
Fix 1514

kbenoit:
Corrects the problem where the keyword pattern was not displaying correctly at the top of the textplot_xray panel.  This was caused by the kwic attribute `keywords` not getting properly set.  Because this was different for characters versus dictionaries, I used a bit of hack to choose unique values (dictionary) or unique patterns (character vector).

The rest of the changes are just linting.


R/kwic.R
```diff
@@ -88,40 +88,43 @@ kwic.corpus <- function(x, pattern, window = 5,
...
-                        class = c("kwic", "data.frame"), 
-                        ntoken = ntoken(x), 
-                        valuetype = valuetype, 
-                        keywords = attr(keywords_id, 'pattern'),
-                        tokens =  attr(temp, "tokens"))
+    result <- structure(temp,
+                        class = c("kwic", "data.frame"),
+                        ntoken = ntoken(x),
+                        valuetype = valuetype,
+                        keywords = if (is.dictionary(pattern)) {
```
kbenoit: @koheiw Here is the key change, you might know a better way to do this.

koheiw: I am not sure if this is the cause 
```r
quanteda:::pattern2list(dictionary(list(A = "a*", B = "b*")), 
                        c("apple", "apricot", "banana"), "glob", TRUE)
```
retunrs
```r
$A
[1] 1

$A
[1] 2

$B
[1] 3

attr(,"key")
[1] "A" "B"
```
so `unique(names(keywords_id))` should be enough.

koheiw: If you want dictionary values instead of keys, why don't you unlist before `pattern2list()`?

kbenoit: I'm pretty sure I get what you're saying, but since you're the grand master of the `pattern2list` mojo, how about you modify that part of this PR (if you think it needs it)?

koheiw:
Firstly, I think labels should be keys when `pattern` is a dictionary, so special handling of dictionary is not necessary.
```r
textplot_xray(kwic(head(toks), dictionary(list(A = "citizen*", B = "government*"))))
```
rplot01

However, I was not sure how `textplot_xlay()` understands which match is for which pattern. As it turned out, it does not know. This function needs re-engineering.
```r
textplot_xray(kwic(head(toks), c("citizen*", "government*")))
```
rplot02
```r
textplot_xray(kwic(head(toks), c("government*", "citizen*")))
```
rplot03

kbenoit:
Totally agree on the first point. In fixing the bug, I changed the behaviour to something that it should not be. I'll fix this.

Will look into the second and fix that too. Good eye! And that's why we get a second pair of them on PRs.

kbenoit:
See #1521, which I think we should discuss there, then fix in this PR.

R/kwic.R
```diff
@@ -88,40 +88,47 @@ kwic.corpus <- function(x, pattern, window = 5,
...
     
     # add document names if none
     if (is.null(names(x))) {
         names(x) <- paste0(quanteda_options("base_docname"), seq_len(x))
     }
-    
-    keywords_id <- pattern2list(pattern, types, 
-                              valuetype, case_insensitive, attr(x, 'concatenator'))
+
+    keywords_id <- pattern2list(pattern, types,
```
kbenoit: @koheiw can you assist with this? Here I need to output the `keywords` attribute below as the key from the dictionary that produced the match. This is in the names of `keywords_id` but they are in a different order to the return `temp`. Setting a breakpoint and running this code, I see:
```r
toks <- tokens(c(alpha1 = paste(letters, collapse = " "),
                 alpha2 = paste(LETTERS, collapse = " ")))
quanteda:::kwic.tokens(toks, dictionary(list(ukey = "u", fkey = "f")), window = 3)
Browse[2]> n
debug at /Users/kbenoit/Dropbox (Personal)/GitHub/quanteda/quanteda/R/kwic.R#124: result <- structure(temp, class = c("kwic", "data.frame"), ntoken = ntoken(x), 
    valuetype = valuetype, keywords = names(keywords_id))
Browse[2]> keywords_id
$ukey
[1] 21

$ukey
[1] 47

$fkey
[1] 6

$fkey
[1] 32

attr(,"key")
[1] "ukey" "fkey"
Browse[2]> temp
  docname from to   pre keyword  post
1  alpha1    6  6 c d e       f g h i
2  alpha1   21 21 r s t       u v w x
3  alpha2    6  6 C D E       F G H I
4  alpha2   21 21 R S T       U V W X
```
What I would like is for `keywords` to match the `temp` vector in rows. Probably easiest to return from the C++ function? or to sort beforehand?

R/kwic.R
```diff
@@ -88,40 +88,47 @@ kwic.corpus <- function(x, pattern, window = 5,
...
+kwic.tokens <- function(x, pattern, window = 5,
+                        valuetype = c("glob", "regex", "fixed"),
+                        separator = " ",
                         case_insensitive = TRUE, ...) {
-    
+
     valuetype <- match.arg(valuetype)
     types <- types(x)
+
+    # coerce a character vector into a dictionary
```
kbenoit: This is the alternative to [my proposal](https://github.com/quanteda/quanteda/issues/1521#issuecomment-446475936) in #1521: We make each character element a dictionary key for itself, and then ensure that kwic handles dictionaries correctly. (It already does - but see below on the keywords.)

R/kwic.R
koheiw: It is not possible currently to trace the match back to pattern, because we are using `SetNgrams`. 
https://github.com/quanteda/quanteda/blob/612718b1bc2a3d287955850f701e966b002bb145/src/kwic_mt.cpp#L79-L80
But it becomes possible if we change it to `MultiMapNgrams`, which is used in `tokens_lookup()`. However, this means that we have to deal with tricky nested and overlapped matches.

kbenoit: Then we need an R function to map the names and numeric type index values in `keywords_id` to `temp$keyword`. But that will be tricky if dictionaries have overlapping values. Or will it - since they will be interchangeable?

R/kwic.R
kbenoit: I ended up coercing characters into a list, not a dictionary, because this changed the behaviour of sequences that were not wrapped in `phrase()`. For dictionary objects, we always consider value sequences as phrases, but not for other pattern types.

R/kwic.R
kbenoit: See `map_keywords()` in `kwic.R`. It works unless pattern is a dictionary and the matches of its values intersect, in which case the `match()` operation will take the first key as its match.

We could warn users about this, since if they want to work around it, then they should perform kwic separately on each (intersecting) key, and then it will work fine for `textplot_xray()`.

koheiw: This is a non-trivial task.

kbenoit: Yes I discovered that! We could warn them in the documentation for `textplot_xray()`.

kbenoit:
@koheiw let me know what you think via a review.

kbenoit:
True. Maybe we could match the first word according to its type index, to get the phrase as the `keywords` entry that matches the corresponding kwic row. We just need a workaround for this case, and right now it's only used for the xray plot facet label.

koheiw:
Another difficult case is 
```r
> stringi::stri_split_fixed(kwic(txt, phrase(c("is i*", "is it", "is in")))$keyword, " ")
[[1]]
[1] "is" "it"

[[2]]
[1] "is" "in"

[[3]]
[1] "Is" "it"

[[4]]
[1] "is" "it"

[[5]]
[1] "Is" "it"

[[6]]
[1] "is" "it"
```
It is impossible to 1-1 match the patterns and keywords here. I think these require too much work only for `textplot_xlay()`.

kbenoit:
OK, I tried but have given up for now trying to solve that issue.

We've fixed all but the sequences. Since PR fixes some clear errors and represents a general improvement, I suggest we merge it, and outline the problems you identify above as a new issue to be solved separately. Those problems existed before this PR, so I don't see why they should hold up us solving other problems. See the warning notes I added to `textplot_xray.Rd`.

Plus the only real consequence is for the facet labels in `textplot_xray()`. If a user really wanted to fix these, it's possible to do so by modifying the ggplot object.

koheiw:
Partial fixes can make problem even more complex. Why don't you remove the pattern label entirely from the plot as you proposed initially?

kbenoit:
Well, dictionaries need facets, and it works for them in the PR (except for phrases) and is broken and wrong in master. As I mentioned, what's broken in the PR was also broken in master, but at least the PR fixes the broken bits for non-phrases.

koheiw:
If the purpose is adding dictionary keys as labels, you could loop over keys and `rbind()` the output from 
`qatd_cpp_kwic()`. I was doing something similar earlier for `tokens_lookup()`. If pattern is a character vector or list from `phrase()`, they will be tread as in the same key (no overhead).
https://github.com/quanteda/quanteda/blob/4ee90dbba53621af427b99822552b849e8fbb144/R/tokens_lookup.R#L112-L117
This is a slow method but the most accurate. `kwic()` cannot be use for inspecting giant tokens object anyway. 
We could make `qatd_cpp_kwic()` more like `qatd_cpp_tokens_lookup()` in the future.

kbenoit:
That's a good solution. Essentially, it's an lapply of kwic over the elements of the list or dictionary keys, and reassembling the results including (attributes) to output the single kwic. I'll work on that when I emerge from the 3-day holiday tunnel that starts this evening.

kbenoit:
OK, I implemented the "iterate over keys" approach, but did this in a way that is consistent for all different pattern types for kwic. It preserves the existing behaviours but now correctly associates the "keyword" attribute as the pattern that was matches, and passes this to `textplot_xray()` for correct faceting. Maybe not the most efficient approach but it's more transparent to keep the manipulations required to do this on the R side, and makes the C++ workhorse parts of kwic more flexible if they can be kept general and then used in `kwic.tokens()` in high-level ways to get the results and objects we want.

Take a look and let me know if you agree that it's ready to merge. I'm eager to see the end of this one!

koheiw:
I changed the code to make it simpler and faster (along with upgrading of core functions). The main structural change is moving matched pattern in the `keyword` attribute to the `pattern` column. By doing this, users can more easily combine `kwic` objects. We should let them to `rbind()` to `kwic` objects instead of passing multiple objects via `...` in `textplot_xlay()`.

R/dictionaries.R
```diff
@@ -93,7 +93,7 @@ split_values <- function(dict, concatenator_dictionary, concatenator_tokens) {
                                                  concatenator_tokens))
         names(result) <- key
     }
-    result
+    return(result)
```
kbenoit: Makes no functional difference, however according to [our own style guide](https://github.com/quanteda/quanteda/wiki/Style-guide) we [should not use `return()`](https://style.tidyverse.org/functions.html#return) to send an object's evaluation back to the parent environment at the end of a function. I'm not bothered by it either way.

kbenoit:
All looks good to me. I also thought about adding `pattern` as a column to the kwic object. A good addition.

Ready to merge I'd say.

kbenoit:
🎉🎉 🎉

R/dictionaries.R
koheiw: I consider this one of the most strange R conventions. We should return values explicitly like in other (proper) programing languages.  For me, [Google R Style Guide](https://google.github.io/styleguide/Rguide.xml) is much more agreeable than **tidyverse**'s.

kbenoit: They're about 90% compatible but the **tidyverse** guide is a bit more complete. We can't agree with https://google.github.io/styleguide/Rguide.xml#object for **quanteda** at least. (Sometimes with R you have to embrace a slight bit of chaos!)

We already have local exceptions (to both), so let's add the suggestion of an explicit `return(finalresult)` to the **quanteda** Style Guide.

Issue #1472, opened 2018-10-25
Call for feedback on textstat_proxy()


koheiw:
We added ` textstat_proxy()` function for computing document similarity with minimum threshold, but kept it experimental until we decide the format of the resulting objects. There are technological discussion #1455 #1469 on the issue, but we want to learn users perspective.

~~Since the function is experimental, you can only access it using `quanteda:::textstat_proxy()`.~~

koheiw:
I started using the function myself in a real project, where I want to compute similarity between two sets of documents (say, corpus A and corpus B). In doing so, I noticed that the `textstat_proxy()` is not very useful yet, because it can only compute similarity within corpus A or B, not between A and B. 

It currently has these arguments `textstat_proxy(x, selection, margin, method , min_proxy, rank)` but I suggest to change it to be `textstat_proxy(x, y, margin, method , min_proxy, rank)`, where `y` can be either feature-selected `x` or a DFM of entirely different set of documents.

kbenoit:
Do you mean sth like this?
``` r
library("quanteda")
## Package version: 1.3.16
## Parallel computing: 2 of 12 threads used.
## See https://quanteda.io for tutorials and examples.

dfm1 <- dfm(data_corpus_inaugural)[1:2, ]
dfm2 <- dfm(data_corpus_dailnoconf1991)[1:2, ]

rbind(dfm1, dfm2) %>%
  textstat_simil()
##                       1789-Washington 1793-Washington Haughey_FF_Taois.txt
## 1793-Washington             0.8362466                                     
## Haughey_FF_Taois.txt        0.8946073       0.7951655                     
## Spring_Lab_Leader.txt       0.8815240       0.7905395            0.9575113
```

koheiw:
I know that combining two DFM has been the recommended approach, but it leads to computation of pair that I am not interested parties in.

We might need a mode to exclude selected features or documents from computation so that the pairs will be only selected and non-selected.

koheiw:
Additionally, the function holds x and transposed x for quick multiplication internally, so it makes more sense to pass two matrix instead of combining.

kbenoit:
OK so what do you mean exactly by "compute similarity...between corpus A and B"? Is it

a) combining all documents or features within a corpus to make one "document" per corpus and then comparing those? (that could use `dfm_group() %>% rbind()`.

b) for features, combining the two corpora to get feature similarity? (That could use `rbind()` as above.)

[...]

PR #1523, opened 2018-12-15
Dev textstat proxy

koheiw:
As discussed in #1472

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -49,12 +49,12 @@ qatd_cpp_kwic <- function(texts_, types_, words_, window, delim_) {
     .Call(`_quanteda_qatd_cpp_kwic`, texts_, types_, words_, window, delim_)
 }
 
-qatd_cpp_similarity_linear <- function(mt, method, targets_, rank, limit = -1.0) {
-    .Call(`_quanteda_qatd_cpp_similarity_linear`, mt, method, targets_, rank, limit)
+qatd_cpp_similarity_linear <- function(mt1, mt2, method, rank, limit = -1.0) {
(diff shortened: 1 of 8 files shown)


koheiw:
If we allow users to apply  `rank` or `min_proxy` only when documents/features are non-intersecting, it solves the issue of asymmetry:
``` 
> mt <- dfm(data_corpus_inaugural)
> textstat_proxy(mt[40:55,], mt[57:58,], rank = 3)
16 x 2 sparse Matrix of class "dgTMatrix"
                2013-Obama 2017-Trump
1945-Roosevelt   .          .        
1949-Truman      .          .        
1953-Eisenhower  .          .        
1957-Eisenhower  .          .        
1961-Kennedy     .          .        
1965-Johnson     .          .        
1969-Nixon       .          .        
1973-Nixon       .          .        
1977-Carter      0.9430332  .        
1981-Reagan      .          0.9370056
1985-Reagan      0.9445279  0.9358077
1989-Bush        .          .        
1993-Clinton     .          .        
1997-Clinton     0.9451457  .        
2001-Bush        .          0.9379407
2005-Bush        .          .        
```

koheiw:
I am adding value check if `y` is a dfm, and agree that we should not advertise experimental functions on SO ever.

Keeping `textstat_proxy` as a separate function is the way to avoid breaking existing code for `textstat_dist/simil()`. We need a way to make changes in APIs, otherwise our packages become unable to deliver invitations and die soon.

I agree rounding distance to zero is a problem, so I can disable `rank` and `min_proxy` for distance measures. However, rounding is fine with similarity in my view. **Matrix** is also so much nicer than pairwise structure. 

If you want me to separate this function to an independent package, I will do. It would be used for non-textual data too.

PR #1530, opened 2018-12-24
Remove unnecessary check

koheiw:
This branch was originally change `textstat_proxy()` to allow users to pass empty DFMs, but it turned out that there are deeper issues with Matrix's dimnames: `dimnames()`, `colnames()` and `rownames()` convert `character()` to `NULL` automatically, and and cause caches in several functions. So I added functions such as `set_dfm_dimnames()` that directly write to the S4 slot. I am open to different names of the internal functions, nonetheless.


tests/testthat/test-textstat_proxy.R
```diff
@@ -254,8 +254,8 @@ test_that("textstat_proxy works with non-intersecting documents or features", {
 
 test_that("raises error when dfm is empty (#1419)", {
     mt <- dfm_trim(data_dfm_lbgexample, 1000)
-    expect_error(textstat_proxy(mt),
-                 quanteda:::message_error("dfm_empty"))
+    expect_silent(textstat_proxy(mt))
```
kbenoit: I like the error message better, otherwise you are saying that all empty features have a distance or similarity of zero, when in fact you don't know a thing about them.

koheiw: If you like it, we should change it to warn instead of error, because error breaks programs that depends on the function in extreme cases.

kbenoit:
That outputs zeros but the distances and similarities are not zero. I think we should match the behaviour of the **dist** and **stats** functions here, e.g.
```r
mt <- dfm_trim(data_dfm_lbgexample, 1000)

textstat_dist(mt)
##    R1 R2 R3 R4 R5
## R2  0            
## R3  0  0         
## R4  0  0  0      
## R5  0  0  0  0   
## V1  0  0  0  0  0
proxy::dist(as.matrix(mt))
##    R1 R2 R3 R4 R5
## R2 NA            
## R3 NA NA         
## R4 NA NA NA      
## R5 NA NA NA NA   
## V1 NA NA NA NA NA
stats::dist(as.matrix(mt))
##    R1 R2 R3 R4 R5
## R2 NA            
## R3 NA NA         
## R4 NA NA NA      
## R5 NA NA NA NA   
## V1 NA NA NA NA NA

textstat_simil(mt)
##    R1 R2 R3 R4 R5
## R2  0            
## R3  0  0         
## R4  0  0  0      
## R5  0  0  0  0   
## V1  0  0  0  0  0
proxy::simil(as.matrix(mt))
##     R1  R2  R3  R4  R5
## R2 NaN                
## R3 NaN NaN            
## R4 NaN NaN NaN        
## R5 NaN NaN NaN NaN    
## V1 NaN NaN NaN NaN NaN
stats::cor(t(as.matrix(mt)))
##    R1 R2 R3 R4 R5 V1
## R1 NA NA NA NA NA NA
## R2 NA NA NA NA NA NA
## R3 NA NA NA NA NA NA
## R4 NA NA NA NA NA NA
## R5 NA NA NA NA NA NA
## V1 NA NA NA NA NA NA
```

koheiw:
We can let `textstat_simil()` and `textstat_dist()` to return `NA` but not `textstat_proxy()` because it is inefficient and has no harm as far as users are aware of it. 

Generally speaking, we are coming to the point where we should provide interfaces to the same underlying functions only different in terms of safety. **armadillo**, for example, has different types of [element assessors](http://arma.sourceforge.net/docs.html#element_access ): one raises errors when indices are our-of-range while others just crash the program (but faster) just like regular C++ vectors. 

In the same vein, I had an idea to separate **quanteda-core** from **quanteda**. The former is a set of high performance functions for expert users and developers, while the latter is for beginners and end users.

tests/testthat/test-textstat_proxy.R
kbenoit: As long as we return NA for the distance or similarity matrix when the documents or features are empty, I'm fine with NA and no warning for `textstat_simil()` and `textstat_dist()`.

kbenoit:
> We can let `textstat_simil()` and `textstat_dist()` to return NA but not `textstat_proxy()` because it is inefficient and has no harm as far as users are aware of it.

:+1:

Can you make that change please?

Sounds like most of the code that is now in **proxyC** will replace `quanteda::textstat_proxy()` soon anyway.

kbenoit:
Suggest we remove `set_fcm_dimnames()` since this should not ever be allowed to happen:
```r
x <- fcm(c("a b c", "a b c"), "window", window = 1)
quanteda:::set_fcm_dimnames(x) <- list(paste0("feature", 1:3), paste0("ALTFEAT", 1:3))
# Feature co-occurrence matrix of: 3 by 3 features.
# 3 x 3 sparse Matrix of class "fcm"
#           features
# features   ALTFEAT1 ALTFEAT2 ALTFEAT3
#   feature1        0        2        0
#   feature2        0        0        2
#   feature3        0        0        0
```

koheiw:
I was not entirely sure if we need `set_fcm_dimnames()`, but here is a reason why I kept it.

FCM are a square matrix when created by `fcm()` but we can select row or columns afterwords, so rectangular FCM is valid. Then, we should be able to coerce a rectangular matrix to a FCM using `set_fcm_dimnames()`. If don't have it, `fcm_select()` will fail.
```r
quanteda_options(threads = 8)
require(quanteda)

dmt <- dfm(data_corpus_inaugural) %>% dfm_trim(100, docfreq_type = "rank")
fmt <- fcm(dmt)
dim(fmt[,"people"]) 
# [1] 146   1

fmt2 <- quanteda:::matrix2fcm(Matrix::rsparsematrix(146, 1, density = 1))
dim(fmt2)
# [1] 146   1

fcm_select(fmt[,1:10], "a*")
# Feature co-occurrence matrix of: 12 by 2 features.
# 12 x 2 sparse Matrix of class "fcm"
#           features
# features      and among
#   and      308636 13074
#   among         0   155
#   a             0     0
#   an            0     0
#   as            0     0
#   all           0     0
#   at            0     0
#   are           0     0
#   american      0     0
#   against       0     0
#   any           0     0
#   america       0     0
```
It is also possible that rows and columns of FCM to be different:
```r
fmt[1:5,6:10]
# Feature co-occurrence matrix of: 5 by 5 features.
# 5 x 5 sparse Matrix of class "fcm"
#        features
# features      to  life     no   have    me
#    of     811716 16251  82158 169059 31631
#    the   1156450 22263 115796 239131 44502
#    and    538505 12763  55336 115255 20103
#    :        8560   401   1054   2134   211
#    among   11809   240   1147   2432   430
```

kbenoit:
**on the fcm issue:**

While `[.fcm()` will break this, we designed the `fcm_select()` and `fcm_remove()` operations to preserve symmetry. I think at one point I even tried to enforce symmetry via the class definition, but that only applies to the constructor and most fcm modification operations did not (re)construct the fcm.

``` r
x <- fcm("a a b a c aa d bb b bb a", context = "window", window = 3)
x
## Feature co-occurrence matrix of: 6 by 6 features.
## 6 x 6 sparse Matrix of class "fcm"
##         features
## features a b c aa d bb
##       a  3 4 2  1 1  2
##       b  0 0 1  2 1  2
##       c  0 0 0  1 1  1
##       aa 0 0 0  0 1  1
##       d  0 0 0  0 0  2
##       bb 0 0 0  0 0  1
fcm_select(x, c("a*", "b*"))
##         features
## features a b aa bb
##       a  3 4  1  2
##       b  0 0  2  2
##       aa 0 0  0  1
##       bb 0 0  0  1
x[1:2, 3:4]
## Feature co-occurrence matrix of: 2 by 2 features.
## 2 x 2 sparse Matrix of class "fcm"
##         features
## features c aa
##        a 2  1
##        b 1  2
```

But I can see why we want the `set_fcm_dimnames()`, and it's only internal, so let's keep it.

kbenoit:
**on the `pad_dfm()` idea**

Doesn't `dfm_select()` already do this, even without the proposed `pad` argument? Let's just rewrite it internally to avoid the `cbind()`.

koheiw:
`dfm_select()` does not have padding function (and it should not do):
```r
mt <- dfm(("a b c"))
dfm_select(mt, c("a", "b", "c", "d"))
# Document-feature matrix of: 1 document, 3 features (0.0% sparse).
# 1 x 3 sparse Matrix of class "dfm"
#        features
# docs    a b c
#  text1 1 1 1
```
It would be better to to make a separate function because padding interacts with `valutype` and `case-insensitive`, making it difficult to use.

`cbind()` is not good because it is a function too add all the elements provided, while the new function should add only missing features and sort them all.

kbenoit:
Interesting, that’s right when `pattern` is a vector, but when it’s a dfm then it does pad by default. I agree it should not for any other pattern type.

koheiw:
I think we should move the special behaviour for a DFM to a new function.

koheiw:
Admittedly `dfm_pad()` is not great as `tokens_select(pad = TRUE)` is similar to the default behavior of `dfm_select()`, which do not drop zero-count features. `dfm_conform()` is certainly better, but you still needs to do use `cbind()` to make conformant matrices `a2` and `b2`:
```
a2 <- dfm_conform(a, cbind(a, b))
b2 <- dfm_conform(b, cbind(a, b))
```
A more efficient way is just combining features:
```
a2 <- dfm_conform(a, union(featname(a), featname(b)))
b2 <- dfm_conform(b, union(featname(a), featname(b)))
```
This means `dfm_conform()` is taking union of features, so it can be `dfm_unite()`:
```
a2 <- dfm_unite(a, b)
b2 <- dfm_unite(b, a)
```
We need a universal rule to sort the features here, however.

kbenoit:
OK, let's move the `dfm_unite()` (still need a better name...) to a new issue and new PR, since it's independent of this, and fix this last change request so we can merge this PR https://github.com/quanteda/quanteda/pull/1530#issuecomment-450179539.

Can you finish that one?

kbenoit:
@koheiw I just added some breaking tests, which I will let you address where you see most appropriate, but I would strongly suggest they be made in `textstat_proxy()` since there zeros currently returned are incorrect.

When those tests pass I think this PR is ready to merge.

koheiw:
I made changes in the R code of `textstat_proxy()` to return `NA` for zero vectors (when threshold is not used), but not sure about the output of `proxy::simil()`. For example, cosine similarity of R1 and R5 is 1 according to the function... I think it should be `NA` or 0 (as inner product is 0). Do you know why?
```r
> as.matrix(mt)
    features
docs A B C D E F G H I J K L M N O P Q R S T
  R1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
  R2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
  R3 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1
  R4 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1
  R5 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0

> textstat_simil(mt, method = "cosine")
   R1 R2 R3 R4
R2 NA         
R3 NA NA      
R4 NA NA  1   
R5 NA NA NA NA
> proxy::simil(as.matrix(mt), method = "cosine")
   R1 R2 R3 R4
R2  1         
R3  0  0      
R4  0  0  1   
R5  1  1  0  0
```

kbenoit:
Ah, interesting case. The product of the vector sums in the denominator for the cosine similarity formula mean that for any vector of zeros, the denominator is zero, and therefore the result is undefined (even 0/0 is undefined). So that should really be `NaN`. I am guessing that `proxy::simil(x, method = "cosine")` uses some workaround so that two identical vectors, regardless of their values, produce a value of 1. 

In the example above, cossimil(R3, R4) = 1 (correctly) but the similarity of any ntoken(x)>0 document with any(ntoken(x))==0 document should be NaN. 

`NA` is also good enough for me so it looks like the existing behaviour of `textstat_simil()` in this example is correct, so we just need to alter the test to reflect that. The code is fine.

koheiw:
It is a bug in **proxy**:

https://github.com/koheiw/proxy/blob/09fa7db01687657afb76ef5a7f15faaebc212770/src/distance.c#L449-L482
```
    if (!R_FINITE(xy)) return NA_REAL;
    xy /= sqrt(xx) * sqrt(yy);
```
The second line should come first.

This is a good reminder that mimicking existing tools does not maker our package better.

kbenoit:
I suspected something like that. 

Matching existing packages is not always the best way to define our behaviours, but it's usually a good start. There's also the goal of integrating well into the existing R ecosystem (as crazy as it often is). 

And I'd trust the **stats** package (which **proxy** of course is not!).

koheiw:
Then, we should stop using `dist` object (or its variant) for saving similarity.

koheiw:
I don't think this is correct either. If variance of R1, 2, 5 are zero, their autocorrelation should be `NA`. 
```r
stats::cor(t(as.matrix(mt)), method = "pearson")
   R1 R2 R3 R4 R5
R1  1 NA NA NA NA
R2 NA  1 NA NA NA
R3 NA NA  1 NA NA
R4 NA NA NA  1 NA
R5 NA NA NA NA  1
```
This is correct
```r
textstat_proxy(mt, method = "correlation", use_na = TRUE) 
5 x 5 sparse Matrix of class "dsTMatrix"
   R1 R2 R3 R4 R5
R1 NA NA NA NA NA
R2 NA NA NA NA NA
R3 NA NA NA NA NA
R4 NA NA NA NA NA
R5 NA NA NA NA NA
```

koheiw:
Diagonals become one because of `as.matrix.dist()`. 
```r
textstat_simil(mt, method = "correlation") %>% as.matrix()
   R1 R2 R3 R4 R5
R1  1 NA NA NA NA
R2 NA  1 NA NA NA
R3 NA NA  1 NA NA
R4 NA NA NA  1 NA
R5 NA NA NA NA  1
```
This means that we cannot even write a test without abandoning `dist` objects.

kbenoit:
I agree with your comment 
https://github.com/quanteda/quanteda/pull/1530#issuecomment-451082696 that the result should be `NA` for the diagonal. This is what `stats::cor(t(as.matrix(mt)), use = “pairwise.complete.obs”)` produces too. But we don’t need such subtleties so the correlation of any quantity with another that has zero variance should be zero.

And yes this is a flaw of using a modified dist object that does not record the diagonal.

So let’s take the plunge and move to the `dsTMatrix` as the default return object for `textstat_simil()` and `textstat_dist()`. This has the advantage of providing a matrix that is efficient when symmetric. We can provide coercion methods to dist for distance objects and maybe even proxy::simil class objects and to the list format as we do currently, but the default can be `dsTMatrix`. This will be a breaking change for some other packages but we can deprecate this. I think we should deprecate the diag argument too so that we always provide the diagonal, which would avoid the problems we are seeing for zero-variance or non-existent features.

So I’m convinced now that the flaws of the missing diagonal for the simil derivative class of dist outweigh the advantages of compatibility with the **proxy** library. However I remain firmly convinced that zero must mean zero in a `dsTMatrix` return value, not a missing value. 

**However let’s do this in a separate branch and PR after we accept this one.** That allows us to focus on the transitional issues and edge cases that motivated it. **In the meantime** let’s `skip()` the breaking tests and add some To Do comments to fix them in the transition to `dsTMatrix`. Then we can discuss alternatives such as the list or pairwise data.frame to avoid the zeros mean something other than zeros issue.

Sound like a plan?

kbenoit:
On the Jaccard, eJaccard, and Dice question, this needs more investigation and I’d be happy to get this PR merged and raise them as a separate issue with a separate solution. Our PRs lately have tended to become overly broad in the issues they encompass.  Although clearly some of them are like trying to tie up a loose end for a knitted sweater that then starts to unravel a whole sleeve!!

koheiw:
Please merge this when you think ready. This branch only breaks your tests against **stats** and **proxy**.

PR #1537, opened 2019-01-03
Correct spliting by i index

koheiw:
A hot fix


kbenoit:
Ah the index from zero feature of sparseCMatrix...

Issue #1545, opened 2019-01-08
Can't apply or change dfm weights with dfm_weight after grouping with dfm_group


juba:
## Description and reproducible code

Suppose I've got the following dfm, weighted with a `boolean` scheme :

```r
m <- matrix(
  c(3,1,5,
    0,0,1,
    4,0,0,
    1,0,2),
  ncol = 3, byrow = TRUE,
)
dfm <- as.dfm(m)
dfm <- dfm_weight(dfm, scheme = "boolean")
```

Later on I use `dfm_group` on it : 

```r
grouped_dfm <- dfm_group(dfm, groups = c(1, 1, 2, 2))
```

The resulting `dfm` is, coherently, not boolean anymore. But if I want to use `dfm_weight` to convert it back to boolean, I can't because the grouped dfm is still considered weighted : 

```
> dfm_weight(grouped_dfm, scheme = "boolean")
Error in dfm_weight.dfm(grouped_dfm, scheme = "boolean") : 
  this dfm has already been term weighted as: boolean
```

And I can't either "unweight" it with `dfm_weight(scheme = "count")` as it generates the same error.

The workaround I use for now is to directly modify `grouped_dfm@x`.

## Expected behavior

I would expect the result of `dfm_group` not to be weighted, at least for weights that are not preserved by the grouping operation.

## System information

```
R version 3.5.2 (2018-12-20)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: KDE neon User Edition 5.14

Matrix products: default
BLAS: /usr/lib/x86_64-linux-gnu/atlas/libblas.so.3.10.3
LAPACK: /usr/lib/x86_64-linux-gnu/atlas/liblapack.so.3.10.3

locale:
 [1] LC_CTYPE=fr_FR.UTF-8       LC_NUMERIC=C               LC_TIME=fr_FR.UTF-8       
 [4] LC_COLLATE=fr_FR.UTF-8     LC_MONETARY=fr_FR.UTF-8    LC_MESSAGES=fr_FR.UTF-8   
 [7] LC_PAPER=fr_FR.UTF-8       LC_NAME=C                  LC_ADDRESS=C              
[10] LC_TELEPHONE=C             LC_MEASUREMENT=fr_FR.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[...]

PR #1547, opened 2019-01-08
Reset weighting scheme when grouped

koheiw:
#1545


kbenoit:
I’m not sure this is really a good “feature”. For instance summing a tf-idf or document frequency weight as part of a `dfm_group()` operation has never been a good idea, but then resetting the flag that indicates where the values came from originally makes it worse.

Maybe we could list the weight types that we think are appropriate for summing in `dfm_group()`? But my basic advice would be: don’t group a weighted dfm. Group first, then weight.

koheiw:
At least, Boolean should me allowed, because grouping of Boolean DFM is counting number of documents with certain features. This is very useful and I do similar operation very often to get daily or monthly counts.

kbenoit:
OK, here is what we can safely aggregate:

- sum: count, prop. 
- Boolean (we would just re-run this through `dfm_weight(x, scheme = “boolean”)`)

However these are likely to cause problems if that is not what a user wanted. What if we just issue a warning for all but count and prop, and let the user do the re-weighting?

kbenoit:
OK here is my proposal to close this out.

1.  We keep perform the grouping and summing as normal for count and prop weighted dfms. 
2. We perform the summing for all other weight types, but issue a warning that these weights are no longer correct.
    - this includes "boolean", so that a user could get sums on the original Boolean weights, but could also convert again to Boolean using a `dfm_weight(x, "boolean")` where `x` has been `dfm_group()`ed.
3. We record a grouping slot to show whether (and how?) the dfm has been grouped. Since @koheiw has been working on grouping schemes esp in branch `dev-corpus2`, I won't get more specific than this at this point, except to raise this possibility in general terms.
3.  We add a `force = FALSE` to `dfm_weight()`, where if it is `TRUE`, we would weight an already weighted dfm, and overwrite the existing weight slot setting.

kbenoit:
OK I thought we should put this PR to bed, so I implemented the following changes to `dfm_weight()` and `dfm_group()`:

- `dfm_weight()` will not weight an already-weighted dfm unless `force = TRUE`
- `dfm_group()` will not group a dfm that has been weighted non-count or non-prop or document-weighted as non-unary, unless `force = TRUE`
- adds tests
- updates `NEWS.md` about these changes

koheiw:
I am fine with ti. Generally speaking, `force` is a good way to leave difficult decisions with users while avoiding errors. `textmodel_*` has `force` too but they are more or less the same, I think.

Issue #1549, opened 2019-01-09
Selection is not working in textstat_simil()


koheiw:
Current master:
```r
> as.matrix(textstat_simil(data_dfm_lbgexample, selection = "R1"))
   R1       R2         R3         R4        R5        V1
R1  1 0.176795 -0.2927247 -0.3199658 -0.320108 -0.124303
```
This is induced in latest update of the function `textstat_simil()` and `textstat_dist()`.

PR #1550, opened 2019-01-09
Fix orientation of selected simi matrix

koheiw:
#1549 I was surprised that such a big bug passed through the unit tests.


koheiw:
Appveyor says **statnet.common**, which only supports R 3.5 and being depended by **network** is missing.

kbenoit:
I suppose it’s inevitable that we have to increase the R dependency version but that’s a big jump.

Issue #1533, opened 2018-12-30
Add a more efficient dfm_conform() type function


kbenoit:
This would be to make two dfm objects conform feature for feature, much like `dfm_select(x, selection = somedfm)` currently does. 

See the discussion here https://github.com/quanteda/quanteda/pull/1530#issuecomment-450525243 on #1530 for a discussion of the starting issues.

The idea would be to deprecate this in `dfm_select()` and encourage the use of this more efficient approach.

PR #1555, opened 2019-01-10
Add dfm_conform() and tests for #1533

koheiw:


kbenoit:
Nicer code than `dfm_select(x, pattern = somedfm)` but I am thinking we can produce the desired behaviour in a simpler way by adding a `fill` argument to `dfm_select()`. We already match the behaviour when pattern is a dfm. 1) This avoids unnecessary function growth and 2) means we can maintain the current behaviour of `dfm_select()` but just augment it with `fill = FALSE` as the default. When `TRUE`, we get the behaviour of `dfm_conform()`.

This could easily be done by keeping this code but marking it as internal and then calling it from `dfm_select()`.

```r
``` r
library(quanteda)
## Package version: 1.3.18
## 
## 
##     View
txt <- c(
  doc1 = "aa bb BB cc DD ee",
  doc2 = "aa bb cc DD ee"
)
mt <- dfm(txt, tolower = FALSE)

# identical
dfm_conform(mt, dfm("aa xx cc"))
## Document-feature matrix of: 2 documents, 3 features (33.3% sparse).
## 2 x 3 sparse Matrix of class "dfm"
##       features
## docs   aa xx cc
##   doc1  1  0  1
##   doc2  1  0  1
dfm_select(mt, dfm("aa xx cc"))
## Document-feature matrix of: 2 documents, 3 features (33.3% sparse).
## 2 x 3 sparse Matrix of class "dfm"
##       features
## docs   aa xx cc
##   doc1  1  0  1
##   doc2  1  0  1

# different but could be made identical if we added a fill = TRUE argument
dfm_conform(mt, c("aa", "xx", "cc"))
## Document-feature matrix of: 2 documents, 3 features (33.3% sparse).
## 2 x 3 sparse Matrix of class "dfm"
##       features
## docs   aa xx cc
##   doc1  1  0  1
##   doc2  1  0  1
dfm_select(mt, c("aa", "xx", "cc"))
## Document-feature matrix of: 2 documents, 2 features (0.0% sparse).
## 2 x 2 sparse Matrix of class "dfm"
##       features
## docs   aa cc
##   doc1  1  1
##   doc2  1  1
```

koheiw:
I thought so too, but I am convinced that it is better no to use `pattern` for filling, because it interacts with `valuetype` and `case-insensitive`.

koheiw:
The biggest difference between them is probably the dfm_conform changes the order of features, while dfm_select does not.

kbenoit:
What if we specify in the description of the `fill` argument that when `fill = TRUE`, that automatically, `valuetype = “fixed”` and `case_insensitive = FALSE`? In other words, only fixed fills.

koheiw:
The behavior is actually `sort_and_fill`, not just `fill`: `dfm_select(valuetype = "fixed", case_insensitive  = FALSE)` + `dfm_sort(margin = "features")` ~ `dfm_conform()`.

koheiw:
I generally don't think it is a good idea to put too many functions to single command, because it makes discovering the function more difficult. I don't say "one command for one function" but `dfm_conform()` should be a separate function.

kbenoit:
OK, after reviewing our usage of the pattern/valuetype/case_insensitive trio, it turns out that dfm is only ever a a valid pattern type in `dfm_select()`. To make this consistent then, we would want to deprecate that and suggest that the new function be used for this purpose. I have a few SO answers out there that use the conforming feature of `dfm_select()` so I'd suggest we deprecate it initially just in the documentation (and suggest the new function instead) but keep the functionality as is, without a warning. Later we can add a deprecation warning.

As for a name, I think we should consider something better than "conform", especially since conforming matrices have a specific meaning that is different from a column match. (I know I probably suggested this name to begin with!) How about `dfm_featurematch()`? Can we not think of something better than `dfm_conform()`?

And should it be 
```r
dfm_featurematch(x, y = NULL, featnames = NULL)
```
where `x` and `y` are dfm objects and either `y` or `featnames` must be specified, but not both?

or should it just be 
```r
dfm_featurematch(x, featnames)
```
where if a user wants to match one dfm to another, then they use `dfm_featurematch(x, featnames(y))`?

koheiw:
I did not like passing a DFM to `features`, so making it `y` sounds better. Since `y` is abstract enough, it is OK to pass a character vector to it, I think.
```r
dfm_featurematch(x, y)
# or
dfm_featurematch(x, y = NULL, features = NULL)
```
That said, "featurematch" is weird. 

`dfm_match()` is good apart from possible confusion with `match()`. We have the `force` argument in `predict()`, which does the same when `TRUE`:
```r
predict(object, newdata = NULL,  se.fit = FALSE, interval = c("none", "confidence"),
        level = 0.95, rescaling = c("none", "lbg", "mv"), force = TRUE, ...)
```
So it could be `dfm_force()`.

Other possible verbs are add, insert, complement, complete, supply...

kbenoit:
I was reluctant too about match but we can just make it clear in the documentation. How about:

```r
dfm_match(x, y)
```
where `y` can be a character vector or a dfm from which the featnames will be taken.

Or, we make it `dfm_match(x, featnames)` where `featnames` is always a vector, and people would need to send it `featnames(somedfm)`. That's more consistent, but I like the flexibility and simplicity of just `dfm_match(x, y)`.

`featurematch` is more specific and descriptive of the function, and does not cause confusion with other methods called `match()` (that return a logical) but as long as cannot think of anything else we would match for two dfm objects, let's just go with `match`. I cannot think of anything else.

koheiw:
I am fine with `dfm_match(x, features)`. People should wrap a DFM by `featnames()`, which is a trivial thing to do.

PR #1558, opened 2019-01-13
Fix encoding

koheiw:
On some Windows systems, dimnames are not encoded.


koheiw:
It is difficult to test encoding issues in CI, but AppVayor seems to be working in the same way as my Windows. Windows 10 itself is moving towards to full-UTF8 rapidly. This is great, but needs attention.

PR #1559, opened 2019-01-14
Add textstat_readability documentation

jiongweilua:
The intent of this pull request is to improve the documentation of `textstat_readability`.

**Notes**
- For measures that appeared multiple times such as Average Sentence Length, Average Word Length, I defined them upfront and used short forms such as ASL, AWL.
- For as many measures as possible, I tried to include the author that proposed the original measure in the `Details` section, and the corresponding primary source in the `References`.

**For Discussion**
- I would like to suggest that we rename the arguments `Bormuth` into `Bormuth.MC` (Bormuth's Mean Cloze) and `Coleman.Liau` into `Coleman.Liau.ECP` (Coleman-Lia's Estimated Cloze Percent). When searching for these measures of readability online, there were often multiple search results - some of them are converted to some form of grade level score, while others are these intermediary scores. It would be good to be more precise.
- Many of the original readability measures apply to 100 / 150 sample words from the text. In this function, we have scaled the readability measures  accordingly. Is it necessary to state this scaling explicitly in the documentation? I did it for `FORCAST` and `Strain`, but then stopped because I'm not too sure whether it is necessary.


kbenoit:
Great stuff @jiongweilua thanks for this!

### Changes I made

* some formatting improvements. More need to be made - in fact we should adopt a **quanteda** bibliographic style, add this to the Wiki style guide, and make sure we use it. We should use some standard social science citation method, e.g. APA. Worth opening a new issue for and discussing.
* added Björnsson to the documentation.
* removed the list of valid measures from the function signature. We were checking this internally anyway, not though `match.arg()`. The new list provides the valid names. This looks *much* cleaner.

### Still need to make re formatting

The formulas look rather messy and could be cleaned up, I can do a few to show you what I mean.

### re your questions

>  would like to suggest that we rename the arguments Bormuth into Bormuth.MC (Bormuth's Mean Cloze) and Coleman.Liau into Coleman.Liau.ECP (Coleman-Lia's Estimated Cloze Percent). 

Yes happy for you to do this. Let's keep the old names as valid aliases but only document the changed ones. Please make this change.

> Many of the original readability measures apply to 100 / 150 sample words from the text. In this function, we have scaled the readability measures accordingly. 

Good to spot this and note it, but this is what I meant by wanting to go back to the sources and verify that we are happy with this implementation. Ideally we would do this for all measures, and keep a checklist or just put each into tests.

jiongweilua:
Hi @kbenoit, 

1. Messy Formulas: Yes indeed I think phrases such as "No. of Words with >=3 Syllables" are long, clunky, could do with some abbreviations, but using too many might be counterproductive if users have to keep cross-checking with a legend / set of abbreviations. It would be great if you could maybe identify a core of terms we want to abbreviate, and I can apply that to all the documented measures.

2. "Bormuth" and "Coleman.Liau" Renaming: Just to be clear, do you mean that I should rename the default arguments be `"Bormuth.MC"` and `"Coleman.Liau"`, but that  `"Bormuth"` and `"Coleman.Liau.EMC"` should be accepted if provided as arguments? 

3. Compiling Sampling-Dependent Measures: Yes I will compile this. And we can figure out where to document it / test it later.

kbenoit:
I meant that we can change the names to 
- Bormuth.MC (but “Bormuth” will still work and will be equivalent to “Bormuth.MC” but not documented)
- Coleman.Liau.ECP (but “Coleman.Liau” will still work and will be equivalent to “Coleman.Liau.ECP” but not documented)

On the 150 word measures, they should be left as is (but with your new note) but would be worth investigating from the original sources and any other implementations we can find whether our current computation is the best way to implement these in the modern era. The 150-word sample was not chosen originally because it was substantively the best way to compute the text’s readability, but instead only as a convenience since all of this stuff had to be done by hand back then!

tests/testthat/test-textstat_readability.R
```diff
@@ -145,4 +145,10 @@ test_that("textstat_readability works for renamed Bormuth.MC and Coleman.Liau.EC
                       textstat_readability(data_char_sampletext, measure = 'Bormuth.MC'))
     expect_identical(textstat_readability(data_char_sampletext, measure = 'Coleman.Liau'),
                       textstat_readability(data_char_sampletext, measure = 'Coleman.Liau.ECP'))
-})
\ No newline at end of file
+})
+
+test_that("textstat_readability raises error for non-included measures",{
```
jiongweilua: @kbenoit I have no idea why this test is not even though the expected and actual messages are identical. I've skipped it in this commit to ensure that the package builds sucessfully, but I am running into this error on my console:

```r
Error: Test failed: 'textstat_readability raises error for non-included measures'
* `quanteda::textstat_readability(data_char_sampletext, measure = "Gibberish")` threw an error
 with unexpected message.
Expected match: "Invalid measure(s): Gibberish"
Actual message: "Invalid measure(s): Gibberish"
```

kbenoit: 😅 It's because the "(" is a regular expression character.

jiongweilua: 😅 thanks!

PR #1575, opened 2019-01-24
Coerce all assigned docvars to character

### LINKED ISSUE #1574: docnames<-() can assign illegal (non-character) document names
Our dfm functions require all dimnames to be of mode `character`, but these can be set to other values. The error should come at assignment via `dimnames<-` but does not.

``` r
library("quanteda")
## Package version: 1.3.19
## 
## 
##     View
dfmat <- dfm(c(d1 = "a b c c c d", d2 = "c c a d e b b"))
docnames(dfmat) <- 1:2
str(dfmat@Dimnames)
## List of 2
##  $ docs    : int [1:2] 1 2
##  $ features: chr [1:5] "a" "b" "c" "d" ...
dfmat
## Document-feature matrix of: 2 documents, 5 features (10.0% sparse).
## Error in validObject(.Object): invalid class "dfm" object: Dimnames[1] is not a character vector

docnames(dfmat) <- as.character(1:2)
str(dfmat@Dimnames)
## List of 2
##  $ docs    : chr [1:2] "1" "2"
##  $ features: chr [1:5] "a" "b" "c" "d" ...
dfmat
## Document-feature matrix of: 2 documents, 5 features (10.0% sparse).
## 2 x 5 sparse Matrix of class "dfm"
##     features
## docs a b c d e
##    1 1 1 3 1 0
##    2 1 2 2 1 1

docnames(dfmat) <- c(TRUE, FALSE)
dfmat
## Document-feature matrix of: 2 documents, 5 features (10.0% sparse).
## Error in validObject(.Object): invalid class "dfm" object: Dimnames[1] is not a character vector
```
This https://github.com/JBGruber/LexisNexisTools/blob/master/R/LexisNexisTools.R#L793 is what is breaking the revdep check #1572 with [LexisNexisTools](https://github.com/JBGruber/LexisNexisTools), although I will also issue a PR and ask @JBGruber to use our replacement and accessor functions rather than grabbing internal quantities directly. ☝️☝️☝️

kbenoit:
Fixes #1574 by coercing all assigned docvars to `character`.

### DIFF
--- a/R/docnames.R
+++ b/R/docnames.R
@@ -34,14 +34,16 @@ docnames.corpus <- function(x) {
     # didn't use accessor documents() because didn't want to pass
     # that large object
     if (is.null(rownames(x$documents))) {
-        paste0('text', seq_len(ndoc(x)))
+        paste0("text", seq_len(ndoc(x)))
     } else {
         rownames(x$documents)
     }
 }
 
 #' @param value a character vector of the same length as \code{x}
-#' @return \code{docnames <-} assigns new values to the document names of an object.
+#' @return \code{docnames <-} assigns new values to the document names of an object.  
+#' docnames can only be character, so any non-character value assigned to be a
+#' docname will be coerced to mode `character`.
 #' @export
 #' @examples 
 #' # reassign the document names of the inaugural speech corpus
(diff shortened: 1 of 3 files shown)


kbenoit:
I'll add a few tests before merging. Can we think of any other functions that operate directly on our docnamed objects' Dimnames?

PR #1577, opened 2019-01-25
More fixes to 1574

kbenoit:
Found a few more places where we should coerce the elements of `dfmat@Dimnames` to `character`. 

I even wrote an S4 validator function but that doesn't work since in dfm.R you are now creating the dfm without dimnames, and then setting them separately. And the class validators are not invoked with low-level manipulations anyway, so it's not a robust solution.

Anyway take a look at these latest additions.

### DIFF
--- a/R/dfm-methods.R
+++ b/R/dfm-methods.R
@@ -87,8 +87,8 @@ as.dfm.default <- function(x) {
 #' @method as.dfm dfm
 #' @export
 as.dfm.dfm <- function(x) {
-    # make sure those dimensions are character
-    x@Dimnames <- lapply(x@Dimnames, as.character)
+    # make sure the dimension names are character
+    set_dfm_dimnames(x) <- lapply(x@Dimnames, as.character)
     # for compatibility with older dfm objects
     if (identical(dim(x@docvars), c(0L, 0L)))
         x@docvars <- data.frame(matrix(ncol = 0, nrow = nrow(x)))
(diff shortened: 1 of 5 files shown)


R/dfm-methods.R
```diff
@@ -87,6 +87,8 @@ as.dfm.default <- function(x) {
 #' @method as.dfm dfm
 #' @export
 as.dfm.dfm <- function(x) {
+    # make sure those dimensions are character
+    x@Dimnames <- lapply(x@Dimnames, as.character)
```
koheiw: Why do you use `lapply` instead of `set_dfm_dimnames()`?

koheiw:
I notice there is a function `provideDimnames` whose man says

> provideDimnames(x) returns x, with “NULL - free” dimnames, i.e. each component a character vector of correct length.

but not yet sure how to use it or if it is useful for us.

kbenoit:
Works like this - we use it to set default doc- and featnames if we implement a "start with 1" workaround.

``` r
library("quanteda", warn.conflicts = FALSE)
## Package version: 1.3.19
library("Matrix")
spmat <- as(Matrix(matrix(1:6, nrow = 2)), "dgCMatrix")

provideDimnames(spmat, base = list(
  paste(quanteda_options("base_docname")),
  quanteda_options("base_featname")
))
## 2 x 3 sparse Matrix of class "dgCMatrix"
##       feat feat1 feat2
## text     1     3     5
## text1    2     4     6
```

R/dfm-methods.R
kbenoit: OK I changed it in 789b9ea but the RHS still needs to access the slot directly since `as.dfm()` is a very low-level function.

PR #1581, opened 2019-01-28
Fix 1570

kbenoit:
Attempts to fix #1570.
- weans `bootstrap_dfm()` and `corpus_sample()` from data.table sampling.

However there appears to be a deeper problem, that stems from data.table. See the discussion here:
https://github.com/Rdatatable/data.table/issues/3300

@koheiw I thought we implemented this setting but I am still seeing the error with the very first data.table call in my R session (macOS). I tried this by creating a dummy internal function 
```r

try_dt <- function() {
    dt <- data.table(iris)
    dt[, samp := sample(Petal.Length), by = Species]
    setDF(dt)
}
```
in `utils.R` (not in there now) and that also triggered the warning. So it's not anything we are doing wrong, it's the use of data.table itself.


kbenoit:
@koheiw this does not solve the OMP warning, but improves the sampling methods. Take a look and if you are ok, let's merge this and try to submit to CRAN despite the OMP issues. (I think the **data.table** team will have to fix these.)

kbenoit:
@koheiw pls let me know if you have any issues with this PR, and if not I will merge before CRAN v1.4.

Issue #1599, opened 2019-02-07
compute information gain in 


randomgambit:
Hello,

A useful feature would be to add the information gain as defined p.17 of https://web.stanford.edu/~jurafsky/slp3/4.pdf. That would be a pretty useful metric. 

What do you think?
Thanks!

jiongweilua:
Hi @randomgambit , currently we already have `textstat_entropy` as a function, which supports the computation of entropy along both documents and features:

```r
> textstat_entropy(dfm(data_corpus_inaugural))
1789-Washington 1793-Washington      1797-Adams  1801-Jefferson  1805-Jefferson 
       7.770221        6.094922        7.754489        7.768814        7.953447 
   1809-Madison    1813-Madison     1817-Monroe     1821-Monroe      1825-Adams 
       7.608605        7.625116        7.982008        8.080853        7.838801 
...
> textstat_entropy(dfm(data_corpus_inaugural), margin = 'features')
fellow-citizens              of             the          senate             and 
      3.6849336       5.5333166       5.5327733       2.8159221       5.6547444 
          house representatives               :           among    vicissitudes 
      2.9139771       3.6163486       4.7401661       5.0918876       2.3219281 
       incident              to            life              no           event 
      2.5000000       5.6068178       5.3148646       5.2873975       3.0306391 
...
```
The measure computed by `textstat_entropy` is the empirical analogue of 
![image](https://user-images.githubusercontent.com/28591528/52401836-9a265d80-2aba-11e9-9789-c47bbae6a7ab.png)
where Pi is the probability of document / feature appearing estimated with their proportion in the dfm.

May I clarify if this sufficient what you need, or do you refer to information gain from a word feature to evaluate feature importances in for a text classification model produced by `textmodel_nb`?  (e.g. if fellow-citizens is included, what's the information gain relative to when it's excluded? )

If it is the latter, I am happy to  help write the implementation. 

But I think before I get started, it would be good to have @kbenoit 's input about whether this feature evaluation functionality we want to build into quanteda in the first place.

randomgambit:
Hi @jiongweilua , yes I think it would be nice to have that feature importance metric for the naive bayes model. This is a very efficient classification algorithm and it would be interesting to be able to find out the most relevant words. One can do that by looking at the word likelihoods, but I am not sure its the optimal way to do (the likelihood are biased estimates)

kbenoit:
[...]

PR #1601, opened 2019-02-09
Add gain.text_modelnb for information gain (issue-1599)

jiongweilua:
Addresses issue #1599. 

**Notes:**
- Information gain formula used : 
image
where  ci is the ith class and wbar means that a document does not contain the word w. 
- Implementation uses internal quantities from a fitted `textmodel_nb` object and relies on Bayes Rule to compute the terms involving wbar. 
- Can either return class-specific information gain, or overall information gain for each feature.

**Remaining Work**
- Documentation to be improved and unit tests to be added.


R/textmodel_nb.R
```diff
@@ -331,3 +331,80 @@ colNorm <- function(x) {
 print.predict.textmodel_nb <- function(x, ...) {
     print(unclass(x))
 }
+
+
+#' Compute the information gain of features from a fitted \link{textmodel_nb}
+#' object
+#' 
+#' Computes the information gain of features from a fitted \link{textmodel_nb}
+#' object.  See
```
kbenoit: add citation here

R/textmodel_nb.R
```diff
@@ -331,3 +331,80 @@ colNorm <- function(x) {
...
+#'          d2 = "Chinese Chinese Shanghai",
+#'          d3 = "Chinese Macao",
+#'          d4 = "Tokyo Japan Chinese",
+#'          d5 = "Chinese Chinese Chinese Tokyo Japan")
+#' trainingset <- dfm(txt, tolower = FALSE)
+#' trainingclass <- factor(c("Y", "Y", "Y", "N", NA), ordered = TRUE)
+#' tmod <- textmodel_nb(trainingset, y = trainingclass, prior = "docfreq")
+#' gain.textmodel_nb(tmod)
+#' gain.textmodel_nb(tmod, class_specific = FALSE)
+#' @export
```
kbenoit: we can put this on the man page with `predict.textmodel_nb()`, using `@rdname`.  

We should add this function to `@seealso` for the `textmodel_nb()` function.

R/textmodel_nb.R
```diff
@@ -331,3 +331,80 @@ colNorm <- function(x) {
...
+gain.textmodel_nb <- function(model, base = 2, class_specific = TRUE) {
+    if (!("textmodel_nb" %in% class(model))) stop("model must be a textmodel_nb object")
```
kbenoit: better to define a generic and then have this as the method. Not only does it make type checking for the first argument unnecessary, but also means you can just call `gain()`. also implement a  `.default` method with the error check as per other functions.

R/textmodel_nb.R
```diff
@@ -331,3 +331,80 @@ colNorm <- function(x) {
...
+    if (class(class_specific) != "logical") stop("class_specific must be a logical (TRUE or FALSE)")
```
kbenoit: `is.logical()` 😄

R/textmodel_nb.R
```diff
@@ -331,3 +331,80 @@ colNorm <- function(x) {
...
+
+    # The formula for entropy of word w is
+    # - summation over all classes ( P(class) * log P(class))  (Term 1)
+    # + P(w) * summation over all classes (P(class |w ) * log(P(class|w)) ) (Term 2)
+    # + P(w') * summation over all classes (P(class | w') * log(P(class|w')) ) (Term 3)
+
+    # Term 1: Initial Entropy
+    # Get summation over all classes of  P(class) * log P(class)
+    initial_entropy <- apply(data.frame(model$Pc), MARGIN = 2, function(x) -1 * x * log(x, base = base))
```
kbenoit: Not sure why you are coercing these to data.frames when keeping them as matrix quantities will work just fine.

R/textmodel_nb.R
```diff
@@ -331,3 +331,80 @@ colNorm <- function(x) {
...
+    if (class_specific == F) initial_entropy <- sum(initial_entropy)
+
+    # Term 2: Entropy Given w
+    # Get the conditional entropy := summation over all classes P(class|w) * log(P(class|w))
+    conditional_entropy <- apply(model$PcGw, MARGIN = 2, function(x) x * log(x, base = base))
+    if (class_specific == F) conditional_entropy <- colSums(conditional_entropy)
```
kbenoit: `TRUE` or `FALSE` but never `T` or `F` ...

R/textmodel_nb.R
```diff
@@ -331,3 +331,80 @@ colNorm <- function(x) {
...
+#'          d3 = "Chinese Macao",
+#'          d4 = "Tokyo Japan Chinese",
+#'          d5 = "Chinese Chinese Chinese Tokyo Japan")
+#' trainingset <- dfm(txt, tolower = FALSE)
+#' trainingclass <- factor(c("Y", "Y", "Y", "N", NA), ordered = TRUE)
+#' tmod <- textmodel_nb(trainingset, y = trainingclass, prior = "docfreq")
+#' gain.textmodel_nb(tmod)
+#' gain.textmodel_nb(tmod, class_specific = FALSE)
+#' @export
+gain.textmodel_nb <- function(model, base = 2, class_specific = TRUE) {
```
kbenoit: To make it fit with the man page for predict, let's call the first argument `object`.

R/textmodel_nb.R
```diff
@@ -331,3 +331,80 @@ colNorm <- function(x) {
...
+    # Now get the unconditional entropy  := P(w) * conditional entropy
+    unconditional_entropy <-  t(data.frame(t(conditional_entropy)) * model$Pw)
+
+    # Term 3: Entropy Given w'
+    # Get the second conditional entropy := summation over all classes P(class | w') * log(P(class|w'))
+    # where P(w') denotes the probability that a word is not in a document
+    # To derive P(class|w'), we note:
+    # P(class | w') = P(class ∩ w') / P(w') = (P(class) - p(class ∩ w)) / P(w')
+    # = ( P(class) - [P(w|c) * P(c)] ) / ( 1 - P(w')) # Bayes Rule
+    Pcintersectw <- data.frame(model$PwGc) * model$Pc
```
kbenoit: Be careful of the difference between matrix multiplication - `%*%` - and regular multiplication which will recycle the shorter object along the columns of the longer.

R/textmodel_nb.R
```diff
@@ -331,3 +331,80 @@ colNorm <- function(x) {
...
+    # PcGwcomp is a dataframe that captures conditional on a word not being in the document,
+    # what is the probability that it belongs to a class?
+
+    conditional_entropy_comp <- apply(PcGwcomp, MARGIN = 2, function(x) x * log(x, base = base))
+    if (class_specific == F) conditional_entropy_comp <- colSums(conditional_entropy_comp)
+
+    unconditional_entropy_comp <- t((1 - model$Pw) * data.frame(t(conditional_entropy_comp)))
+
+    # Add the three terms and return either as matrix of named-list
+    if (class_specific == T) {
```
kbenoit: if this is logical then you can omit the `== TRUE` and just use `else` for the `FALSE` case

R/textmodel_nb.R
randomgambit: @kbenoit I agree it would be useulf to cross-check the computation with a well-known example. Jurafsky et al. do not provide one. maybe looking on google scholar can help?

kbenoit: Sure, anything we/you can find would be useful (and citeable).

R/textmodel_nb.R
randomgambit: this is great! were you able to check that the computation is correct?

jiongweilua: Added `gain` to `@seealso` to the man page for see also `textmodel_nb()`, not too sure what you meant by the `@rdname` part though @kbenoit

R/textmodel_nb.R
jiongweilua: Added a `gain.default` method to be exported, but I am not too sure how to implement the error check. Should I be looking at `friendly_class_undefined_message`?

R/textmodel_nb.R
jiongweilua: i initially coerced to data.frame because I ran into the following error:
```r
 Error in apply(conditional_entropy, 1, `*`, object$Pw) : 
  dim(X) must have a positive length 
```
but I have since fixed this by using `sapply`

R/textmodel_nb.R
jiongweilua: I think the latest commit I pushed [29d3101](https://github.com/quanteda/quanteda/pull/1601/commits/29d31018d43407ce2a0f0ac4273b8e1a52d92931) has implemented the matrix multiplication correctly, but will be best to have another reviewer to check.

R/textmodel_nb.R
kbenoit: Yes - I added this in 757a721. No worries.

kbenoit:
Closing until we can return to this, after some (promised) input from Dan Jurasky.

Issue #1603, opened 2019-02-12
Can't easily rename docvars


kbenoit:
The indexing is clearly off:

``` r
library("quanteda", warn.conflicts = FALSE)
## Package version: 1.4.0
## Parallel computing: 2 of 12 threads used.
## See https://quanteda.io for tutorials and examples.

corp <- corpus(c("A b c d.", "A a b. B c."),
  docvars = data.frame(testdv = 10:11)
)
summary(corp)
## Corpus consisting of 2 documents:
## 
##   Text Types Tokens Sentences testdv
##  text1     5      5         1     10
##  text2     6      7         2     11
## 
## Source: /private/var/folders/1v/ps2x_tvd0yg0lypdlshg_vwc0000gp/T/RtmpwTZK9J/reprexae23aa52f45/* on x86_64 by kbenoit
## Created: Tue Feb 12 12:19:06 2019
## Notes:

names(docvars(corp))[1] <- "renameddv"
docvars(corp)
##       testdv renameddv
## text1     10        10
## text2     11        11
summary(corp)
## Corpus consisting of 2 documents:
## 
##   Text Types Tokens Sentences testdv renameddv
##  text1     5      5         1     10        10
##  text2     6      7         2     11        11
## 
## Source: /private/var/folders/1v/ps2x_tvd0yg0lypdlshg_vwc0000gp/T/RtmpwTZK9J/reprexae23aa52f45/* on x86_64 by kbenoit
## Created: Tue Feb 12 12:19:06 2019
## Notes:
```

kbenoit:
I experimented with this for a bit and it's a deep problem, due to the deep and mysterious nature of `names<-`. I am not sure there is actually a way for us to do this on the data.frame itself without also reassigning it to the corpus.

PR #1605, opened 2019-02-13
Fix docvar addressing for corpus

kbenoit:
- fixes names(docvars())<- to solve #1603
- linting changes for docvars.R
- ensures V1, V2, ... names for docvars without names (vectors)

@koheiw I know this will be solved in `dev-corpus2` (and provides yet one more good reason to move to that structure) but this provides a fix in the meantime. Ain't pretty but it works.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: quanteda
-Version: 1.4.0
+Version: 1.4.1
 Title: Quantitative Analysis of Textual Data
 Description: A fast, flexible, and comprehensive framework for 
     quantitative text analysis in R.  Provides functionality for corpus management,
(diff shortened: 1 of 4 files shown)


koheiw:
Do we have to fix this now?

kbenoit:
Why not? It's a minor internal change but makes https://github.com/quanteda/blog.quanteda.org/pull/14 work.

koheiw:
I will take time to review and certainly conflict with `dev-corpus2`.

kbenoit:
Sounds good. I think the conflict is minimal since it's all internal to indexing within the `docvars<-.corpus()`. The rest of the code changes are just tidying up following `lintr::lint()`, nothing substantive.

kbenoit:
🤣 It's not that bad! 

Next week we can go to http://www.mamastrattoria.com.au/ my treat, all you can eat spaghetti.

Issue #1614, opened 2019-02-19
changes to metadoc()


kbenoit:
Deprecate, with an aim to removing in 2.0. Everything simply becomes a docvar.

To get system docvars, we can:

1) `docvars(x, field = NULL, system = TRUE)` returns entire `attr(x, "docvars")`

2) These would also work directly
    - `docvars(corp, "docid_")`
    - `docvars$docid_`
(but `docvars<-()` or `docvars$<-` will _never_ work on system docvars)

So: allow `docvars(x, "docid_")` to work, by changing docvars() to:
```r
docvars.corpus <- function(x, field = NULL) {
    x <- as.corpus(x)
    select_docvars(attr(x, 'docvars'), field, user = TRUE, system = !is.null(field), drop = TRUE)
}
```

**Disregard all of what is written below!**

* deprecate this, in favour of:
```r
docvars(x, field, type = c("user", "meta", "system", "all")) <- value
```
* to set a meta docvar, these would be equivalent:
    - `docvars(corp, type = "meta", field = "sourceurl") <- "htttp://example.com"`
    - `docvars(corp, "sourceurl_") <- "htttp://example.com"`
    - `corp$"sourceurl_" <- "htttp://example.com"`

* Automatically treat docvars based on name and record this way using this scheme:
    - `name` is a user docvar
    - `name_` is a meta docvar, as in `metadoc()<-`
    - `name__` is a system docvar, which cannot be assigned by a user, but can be read

* `docvars("name")` will retrieve the name if fully specified, regardless of type

kbenoit:
Un-deprecate for 1.5 to stop warnings in tests?

PR #1617, opened 2019-02-19
Rework meta into metacorpus for #1614

kbenoit:
To re-implement metacorpus(), as per #1614.

- splits the "meta" into meta$system and meta$user


R/corpus-methods-quanteda.R
```diff
@@ -120,10 +119,9 @@ as.corpus.default <- function(x) {
 #' @export
 #' @method as.corpus corpus
 as.corpus.corpus <- function(x) {
-    
     if (is.character(x)) {
         attr(x, "docvars") <- upgrade_docvars(attr(x, "docvars"))
-    } else {
+    } else if (is.null(meta(x, type = "system")[["package-version"]])) {
         x <- upgrade_corpus(x)
```
koheiw: You should change the `upgrade_corpus()` a bit, because there are lines like:
`attr(result, "meta")$created <- as.POSIXlt(Sys.time())`

R/corpus.R
```diff
@@ -160,8 +165,8 @@ corpus.character <- function(x, docnames = NULL, docvars = NULL, ...) {
     names(x) <- docvar[["docname_"]]
     class(x) <- "corpus"
     attr(x, "unit") <- "documents"
-    attr(x, "meta") <- meta("character", ...)
     attr(x, "docvars") <- docvar
+    x <- meta_init(x, source = "character", user = meta)
```
koheiw: I think it is more flexible if  `meta_init()` only returns a list. Then 
```r
attr(x, "meta") <-  meta_init(source = "character", user = meta)
```
If the function is remaked to `make_meta()` then it becomes easier to understand what happening in functions:
```r
attr(x, "meta") <-  make_meta(...)
attr(x, "docvars") <-  make_docvars(...)
```

R/meta.R
```diff
@@ -0,0 +1,132 @@
...
+    # type checking
+    stopifnot(is.character(source))
+    stopifnot(is.list(user))
+    # make sure every element of user is named
+    if (length(user)) stopifnot(length(names(user)) == length(user))
+    UseMethod("meta_init")
+}
+    
+#' @rdname meta_init
+meta_init.corpus <- function(x, source, user = list()) {
```
koheiw: I prefer the function be called "make_meta" and return a list just line `make_docvars()`.

R/corpus.R
kbenoit: Agreed, I will change this to a `meta()` that is exported and a `meta_system()/meta_system<-()` that work just like `meta()` but are not exported and that also has an initialize option. That way we can encapsulate all of the access to the attributes (or for a dfm, slots) inside `meta_system()`. Our goal should be only to access the attrs/slots there, not in any other function.

R/meta.R
kbenoit: See above

kbenoit:
And of course, we need to add tests once we have finalized the function names.

Issue #1615, opened 2019-02-19
dev-corpus2 docvars<- breaks on NULL


kbenoit:
Can't easily NULL a variable:
``` r
library("quanteda", warn.conflicts = FALSE)
## Package version: 1.4.1
## Parallel computing: 2 of 12 threads used.
## See https://quanteda.io for tutorials and examples.

corp <- corpus(c(d1 = "one", d2 = "two two"),
  docvars = data.frame(
    test1 = 1:2,
    test2 = c("a", "b")
  )
)
corp
## Corpus consisting of 2 documents and 2 docvars.
## 
## 
## [1] "one"     "two two"
## 
## docvars:
##   docname_ docid_ docnum_ segnum_ test1 test2
## 1       d1     d1       1       1     1     a
## 2       d2     d2       2       1     2     b
## 
## meta:
## $source
## [1] "character"
## 
## $`package-version`
## [1] '1.4.1'
## 
## $`r-version`
## [1] '3.5.2'
## 
## $system
##   sysname   machine      user 
##  "Darwin"  "x86_64" "kbenoit" 
## 
## $directory
## [1] "/private/var/folders/1v/ps2x_tvd0yg0lypdlshg_vwc0000gp/T/RtmpSl1y9o/reprex450012dc182e"
## 
## $created
## [1] "2019-02-19"
docvars(corp)
##   test1 test2
## 1     1     a
## 2     2     b
docvars(corp, "test2") <- NULL
corp
## Corpus consisting of 2 documents.
## 
## 
## [1] "one"     "two two"
## 
## docvars:
##   docname_ docid_ docnum_ segnum_
## 1       d1     d1       1       1
## 2       d2     d2       2       1
## 
## meta:
## $source
## [1] "character"
## 
## $`package-version`
## [1] '1.4.1'
## 
## $`r-version`
## [1] '3.5.2'
## 
## $system
##   sysname   machine      user 
##  "Darwin"  "x86_64" "kbenoit" 
## 
## $directory
## [1] "/private/var/folders/1v/ps2x_tvd0yg0lypdlshg_vwc0000gp/T/RtmpSl1y9o/reprex450012dc182e"
## 
## $created
## [1] "2019-02-19"
docvars(corp)
## data frame with 0 columns and 2 rows
```

Issue #1612, opened 2019-02-19
Make sure units attribute is being set correctly


kbenoit:
tokens_chunk(), corpus_reshape(), corpus_segment()

koheiw:
There are multiple functions to copy attributes.

- `set_dfm_slots()`
- `set_fcm_slots()`
- `slots()`

We have to redeign/reorganize them.

kbenoit:
Will the solution here solve the breaking tests in PR https://github.com/quanteda/quanteda/pull/1793? Those failures are caused by attributes not being robustly copied.

koheiw:
It might, becasue it makes `as.dfm()` more robust.
https://github.com/quanteda/quanteda/blob/981ec34294773c751ea353a920d81b29eeec6198/R/dfm-methods.R#L74-L79

Issue #1611, opened 2019-02-19
docname uniqueness behaviour


kbenoit:
* element name is always the "docname": character element name for corpus, list element name for tokens, and rownames for dfm.

* docnames will always be unique

* if a corpus is constructed where the inputs for docnames are not unique, then they will be made unique

* `docid_` is an internal factor recording the document identifer
    - this will become non-unique after `tokens_chunk()`, `corpus_segment()`, or `corpus_reshape()`
    - we will add `corpus(x, ..., unique_docnames = TRUE)` by default to enforce that all docnames will be unique, in which case the docname and `_docid` will be the same.
    - if `corpus(x, ..., unique_docnames = FALSE)` and supplied docnames are not unique, then the docnames will be made unique, but the `docid_` will be non-unique, and `attr(x, "unit")` will be `segments`. This makes a corpus that has been just constructed look like one that has just been segmented. We also would then set `segnum_` to a sequence within `docid_`.

* We remove `docname_` and `docnum_` from docvars since these are not needed.

* Redefine `docnames()<-` and `names()<-` for corpus, tokens, and dfm to enforce the uniqueness behaviour.

koheiw:
I was thinking what is the most robust way to sync `docname_` and element names. One way is to override the functions that change element names such as `row.names()<-` and `rownames()<-` but we cannot rally prevent them from tempering with object attributes, because there is no private/public distinction in variables in R.

So how about letting users to change the element names, but overwrite them by `docname_` every time they intact with the objects? I think this can be done just by adding a few likes in `as.corpus()`, `as.tokens()`, `as.dfm()`.

kbenoit:
Don't we already do this? Maybe I misunderstood the nuance of your suggestion, but I thought we captured that here:

https://github.com/quanteda/quanteda/blob/dev-v2/R/docnames.R#L83-L128

The only way a user can override this is for a dfm, by accessing the `@Dimnames` slot directly. If that is what they want to do, I saw either a) let them, they must be experts and have a very good reason for going this close to the metal, or b) let them and to hell with them for being so foolish if they don't know what they are doing.

koheiw:
My suggestion is to remove 
https://github.com/quanteda/quanteda/blob/dev-v2/R/docnames.R#L83-L128
and change
https://github.com/quanteda/quanteda/blob/dev-v2/R/corpus-methods-quanteda.R#L121-L125
to
```r
as.corpus.corpus <- function(x) {
[...]

PR #1620, opened 2019-02-19
Issue 1611

koheiw:
#1611 #1615 #1612


R/docvars.R
```diff
@@ -75,28 +80,26 @@ make_docvars <- function(n, docname = NULL, unique = TRUE) {
...
+                   "segid_" = integer(), 
                    stringsAsFactors = FALSE)
     } else {
         if (unique) {
             docnum <- match(docname, unique(docname))
             if (any(duplicated(docname))) {
-                segnum <- stats::ave(docname == docname, docname, FUN = cumsum)
-                docid <- paste0(docname, ".", segnum)
+                segid <- stats::ave(docname == docname, docname, FUN = cumsum)
+                docid <- paste0(docname, ".", segid)
```
kbenoit: This means that not just duplicated docnames, but also all docnames, get a numerical suffix. Is this what we want?
```r
corpus(letters[1:3], docnames = c("A", "A", "B"), unique_docnames = FALSE) %>%
    docnames()
## [1] "A.1" "A.2" "B.1"
```

Would this not be better?
```r
## [1] "A.1" "A.2" "B"
```

R/docvars.R
```diff
@@ -125,32 +128,27 @@ subset_docvars <- function(x, i = NULL) {
 
 # internal function to upgrade docvars to modern format
 upgrade_docvars <- function(x, docnames = NULL) {
-    if (sum(is_system(colnames(x))) == 4) 
+    if (sum(is_system(colnames(x))) == 3)
```
kbenoit: Better to check for an `attr(x, "meta"$system$version` and if this is `NULL` we know it's pre-1.5, or later, we can get a version number. (I think I implemented this already in the `dev-corpus2-fix-1613` branch.) Probably we should write a function that checks which version and returns a value. If we don't find the new corpus structure we can return ">=1.4".

tests/testthat/test-docvars.R
```diff
@@ -467,52 +448,37 @@ test_that("docvars<-.corpus and name uniqueness", {
...
+                     c("debate", "foren", "name", "party"))
     
-    # can NULL one of severeal docvars
-    corp <- corpus(c("A b c d.", "A a b. B c."),
-                   docvars = data.frame(testdv = 10:11, seconddv = letters[1:2]))
-    docvars(corp, "seconddv") <- NULL
-    expect_identical(
-        docvars(corp),
-        data.frame(testdv = 10:11, row.names = docnames(corp))
-    )
```
kbenoit: This test fails, so I added it back, since it should pass.
```r
corp <- corpus(c("A b c d.", "A a b. B c."),
                   docvars = data.frame(testdv = 10:11, seconddv = letters[1:2]))
docvars(corp, "seconddv") <- NULL
docvars(corp)
## data frame with 0 columns and 2 rows
```

kbenoit: Looks fixed, but where did the test go?

kbenoit:
names<-.corpus/tokens and rownames<-.dfm now resolved.

However shouldn't we also change the factor names of `docid_` to reflect changes in these too?

PR #1626, opened 2019-02-22
Removed global std namespace usage from quanteda.h

jtatria:
Qualified std member access in fcm_mt.cpp, kwic_mt.cpp, tokens_chunk_mt.cpp, tokens_segment_mt.cpp


kbenoit:
@koheiw looks pretty straightforward to me, please approve if you are happy with it.

PR #1630, opened 2019-02-25
Update the corpus objects to v2 structure

kbenoit:
* Updates the corpus objects to v2 and verifies that all code, tests, functions etc still work.
* Renames the older test objects to v2
* changes `is_pre15()` to `is_pre2()`
* Removes settings (`settings.R`) and related functions/slots/attributes etc, since these interfered with tests and handling of the new objects. (:cheer: I know...)


kbenoit:
We should add a format section to `data_corpus_irishbudget2010` and `data_corpus_dailnoconf1991` similar to the one in `data_corpus_inaugural`.

kbenoit:
@koheiw can you add a test for `docvars.R:153`?

kbenoit:
@koheiw you good with this?

Issue #1634, opened 2019-03-02
rank and frequency


ManfredBernhard:
Dear Quanteda,
I realised that  identical frequencies in a dfm receive different ranks as shown below.

txtfreq<-textstat_frequency(froschk_dfm)
txtfreq
           feature frequency rank docfreq group
1               und        64    1       1   all
2               der        37    2       1   all
3               sie        35    3       1   all
4               die        30    4       1   all
5                du        20    5       1   all
6            frosch        20    6       1   all
7                er        20    7       1   all
8                in        19    8       1   all
9               als        19    9       1   all
10              ich        17   10       1   all
11              war        15   11       1   all
12              ihr        15   12       1   all
13               es        15   13       1   all
14               da        15   14       1   all
15             aber        14   15       1   all
16               so        14   16       1   all
17              ein        13   17       1   all
18              dem        13   18       1   all

Can you please fix this bug?
Best,
ManfredBernhard

jiongweilua:
Hi @ManfredBernhard ,

Do you mean to say that in this section of the output, they should all be ranked as 5 rather than 5, 6 and 7 ?

```r
feature frequency rank docfreq group
5 du 20 5 1 all
6 frosch 20 6 1 all
7 er 20 7 1 all
```

kbenoit:
Good point. It would be better to replace https://github.com/quanteda/quanteda/blob/master/R/textstat_frequency.R#L92-L93 with a call to `data.table::frank()` and add `ties.method` as an argument. Our existing method is basically the same as `ties.method = "random"`.

However it's easy to override this (although it can be a bit trickier if you have used the `groups` argument, but not much). Just overwrite the `ranks` column of the output with one you've computed on your own using alternative `ties.method` arguments to `rank()`. 

``` r
library("quanteda", warn.conflicts = FALSE)
## Package version: 1.4.9000
## Parallel computing: 2 of 12 threads used.
## See https://quanteda.io for tutorials and examples.

dfmat <- dfm(c("a a b c d d", "a b b c d"))
dfmat
## Document-feature matrix of: 2 documents, 4 features (0.0% sparse).
## 2 x 4 sparse Matrix of class "dfm"
##        features
## docs    a b c d
##   text1 2 1 1 2
##   text2 1 2 1 1

tstat <- textstat_frequency(dfmat)
tstat
##   feature frequency rank docfreq group
## 1       a         3    1       2   all
## 2       b         3    2       2   all
## 3       d         3    3       2   all
[...]

PR #1636, opened 2019-03-03
Add ties.method to textstat_frequency()

kbenoit:
Passes this through to `data.table::frank()`, but uses the `"random"` method for backward compatibility (with how `textstat_frequency()` operated before. 

Open questions:
- Should the default be something else? `"dense"` makes a lot of sense.
- Should we follow existing practice in `base::rank()` and `data.table::frank()` and call this argument `ties.method` or should we follow our general practice of avoiding argument names containing ".", and call this `ties_method`?

Solves #1634.


kbenoit:
@stefan-mueller comments welcome

koheiw:
`min` sounds the best because it is consistent with `termfreq_type = "rank"` in `dfm_trim()`:

https://github.com/quanteda/quanteda/blob/b86b775ea2be4e93f56b6cb1d0865fbb0869c8f4/R/dfm_trim.R#L172

kbenoit:
True, I'd forgotten about that. That makes sense in `dfm_trim()` though because we are using the rank as a cutoff. When the rank is returned as information, then "min" is definitely not my favourite.

stefan-mueller:
Looks good! I also tested various **ggplot2** examples for all values for `ties.method` and the plotting works fine.

koheiw:
I am fine to let users to choose the method to break ties, but are you sure that the name of the argument is `ties.method` instead of `ties_method`?

kbenoit:
How about this then:
* call it `ties_method`
* which of these rankings makes more sense, for the vector [9, 9, 20]:
    - 1, 1, 3 ("min": ties ranked equivalently, but recording the next value as the actual ordinal place)
    - 1, 1, 2 ("dense": ties ranked equivalently, but recording the next value as the next ordinal place)
    - 2, 1, 3 ("random", as one of the two possible outcomes)
    - 1, 2, 3 ("first", using the arbitrary order in which it is sorted before ranking)

I don't think there is a strong case to be made for "average", "max", or "last" but of course the user could supply these.

kbenoit:
OK, I changed the argument name to `ties_method` and made `"min"` the default. I think this makes most sense, and it's consistent now with `termfreq_type = "rank"` in `dfm_trim()`.

Issue #1640, opened 2019-03-07
Quanteda 1.4.1 loses tokens when input data is big


Astelix:
I got different results from several analyses with Quanteda 1.3.4 and 1.4.1. The problem seems to stem from version 1.4.1 seems to loose tokens when the input data is big.

To reproduce the following example you need my data file (which I dont't want to upload here).

``` r
sessionInfo()
#> R version 3.5.1 (2018-07-02)
#> Platform: x86_64-w64-mingw32/x64 (64-bit)
#> Running under: Windows 8.1 x64 (build 9600)
#> 
#> Matrix products: default
#> 
#> locale:
#> [1] LC_COLLATE=German_Germany.1252  LC_CTYPE=German_Germany.1252   
#> [3] LC_MONETARY=German_Germany.1252 LC_NUMERIC=C                   
#> [5] LC_TIME=German_Germany.1252    
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> loaded via a namespace (and not attached):
#>  [1] compiler_3.5.1  magrittr_1.5    tools_3.5.1     htmltools_0.3.6
#>  [5] yaml_2.2.0      Rcpp_1.0.0      stringi_1.3.1   rmarkdown_1.11 
#>  [9] highr_0.7       knitr_1.21      stringr_1.4.0   xfun_0.5       
#> [13] digest_0.6.18   evaluate_0.13

library(tidyverse)
#> Warning: Paket 'tidyverse' wurde unter R Version 3.5.2 erstellt
#> Warning: Paket 'ggplot2' wurde unter R Version 3.5.2 erstellt
#> Warning: Paket 'tibble' wurde unter R Version 3.5.2 erstellt
#> Warning: Paket 'tidyr' wurde unter R Version 3.5.2 erstellt
#> Warning: Paket 'readr' wurde unter R Version 3.5.2 erstellt
#> Warning: Paket 'dplyr' wurde unter R Version 3.5.2 erstellt
#> Warning: Paket 'stringr' wurde unter R Version 3.5.2 erstellt
#> Warning: Paket 'forcats' wurde unter R Version 3.5.2 erstellt
library(quanteda)
#> Warning: Paket 'quanteda' wurde unter R Version 3.5.2 erstellt
#> Package version: 1.4.1
#> Parallel computing: 2 of 8 threads used.
#> See https://quanteda.io for tutorials and examples.
#> 
[...]

PR #1641, opened 2019-03-08
Add Unicode normalization

koheiw:
#1640

### DIFF
--- a/R/corpus.R
+++ b/R/corpus.R
@@ -149,6 +149,7 @@ corpus.character <- function(x, docnames = NULL,
     unused_dots(...)
     name <- names(x)
     x[is.na(x)] <- ""
+    x <- stri_trans_nfc(x)
     
     # convert the dreaded "curly quotes" to ASCII equivalents
     x <- stri_replace_all_fixed(x,


kbenoit:
This looks fine but what if rather than hard-wiring this, we gave the user a bit more control by adding an option
```r
quanteda_options(unicode_normalization = "nfc")
```
where the legal values would be those implemented in **stringi**, namely nfc, nfd, nfkc, nfkc_casefold, and nfkd?

koheiw:
All the method call `corpus.character()` including `tokens.character()`:
https://github.com/quanteda/quanteda/blob/b86b775ea2be4e93f56b6cb1d0865fbb0869c8f4/R/tokens.R#L185-L187

I do not really understand the difference between nfc, nfd, nfkc, nfkc_casefold but I think nfc does very little (make characters look the same actually the same in Unicode).

kbenoit:
Quite right on the corpus call from tokens! I thought I saw an example where this was skipped. Maybe I should return my coffee to the vendor since I think they slipped me a decaf... 😕 

On the Unicode normalizations, I don't know much about the differences either, so we should probably educate ourselves more before hardwiring the choice of nfc. Or, we just ask @gagolews... 😃 

In https://github.com/quanteda/readtext/issues/146 however I found that nkfc solved a ligature problem but that nfc did not.

``` r
txt <- "substantial investments in human and ﬁnancial 
resources for the eﬀect of earlier diagnosis of waﬄes."

# ligatures not normalized
cat(stringi::stri_trans_nfc(txt))
## substantial investments in human and ﬁnancial 
## resources for the eﬀect of earlier diagnosis of waﬄes.

# ligatures normalized
cat(stringi::stri_trans_nfkc(txt))
## substantial investments in human and financial 
## resources for the effect of earlier diagnosis of waffles.
```

koheiw:
There is no harm to add the option if there is a use-case, but characters for `pattern` are always normalized by `stri_trans_nfc()`, so we should use this in corpus too until we add the option.

gagolews:
BTW, it's not uncommon to call a series of normalizations, e.g., `nfkc; nfc`

gagolews:
you can do that with a single call to `stri_trans_general(str, "nfkc; nfc") `

Issue #1653, opened 2019-03-26
Current (v1.4.1) CRAN release failing on Solaris


kbenoit:
https://www.r-project.org/nosvn/R.check/r-patched-solaris-x86/quanteda-00install.html

@koheiw Seems to relate to `tokens_chunk_mt.cpp`:

```
tokens_chunk_mt.cpp: In function ‘quanteda::Texts chunk(quanteda::Text&, quanteda::UintParam&, int, int)’:
tokens_chunk_mt.cpp:16:45: error: call of overloaded ‘ceil(std::vector<unsigned int>::size_type)’ is ambiguous
     chunks.reserve(ceil(tokens.size() / step));
                                             ^
In file included from /usr/include/math.h:15:0,
                 from /opt/csw/include/c++/5.2.0/cmath:44,
                 from /home/ripley/R/Lib32/Rcpp/include/Rcpp/platform/compiler.h:100,
                 from /home/ripley/R/Lib32/Rcpp/include/Rcpp/r/headers.h:59,
                 from /home/ripley/R/Lib32/Rcpp/include/RcppCommon.h:29,
                 from /home/ripley/R/Lib32/Rcpp/include/Rcpp.h:27,
                 from quanteda.h:4,
                 from tokens_chunk_mt.cpp:1:
/opt/csw/lib/gcc/i386-pc-solaris2.10/5.2.0/include-fixed/iso/math_iso.h:188:21: note: candidate: long double std::ceil(long double)
  inline long double ceil(long double __X) { return __ceill(__X); }
                     ^
/opt/csw/lib/gcc/i386-pc-solaris2.10/5.2.0/include-fixed/iso/math_iso.h:159:15: note: candidate: float std::ceil(float)
  inline float ceil(float __X) { return __ceilf(__X); }
               ^
/opt/csw/lib/gcc/i386-pc-solaris2.10/5.2.0/include-fixed/iso/math_iso.h:75:15: note: candidate: double std::ceil(double)
 extern double ceil __P((double));
               ^
*** Error code 1
make: Fatal error: Command failed for target `tokens_chunk_mt.o'
```

kbenoit:
@jeroen has set up a VM containing Solaris at https://github.com/jeroen/solarisvm 

He tells me that It has two versions of R, and that the one in `/opt/csw/bin/R` should support **Rcpp**.

[...]

PR #1655, opened 2019-03-28
Cast to double to address #1653

koheiw:
I remember that we had similar issue in `tokens_select_mt.cpp` etc. earlier.


jeroen:
I confirm this fixes the issue:

Screen Shot 2019-03-28 at 1 56 16 PM

Issue #1658, opened 2019-03-31
Tests with R-devel are failing


koheiw:
https://win-builder.r-project.org/incoming_pretest/quanteda_1.4.3_20190329_031431/Windows/00check.log
https://win-builder.r-project.org/incoming_pretest/quanteda_1.4.3_20190329_031431/Debian/00check.log

koheiw:
This should be related to  change in sample() in R 3.6:
https://cran.r-project.org/doc/manuals/r-devel/NEWS.html

We could add `seed` and other functions argument to guarantee reproducibility:
https://stackoverflow.com/questions/48626086/same-seed-different-os-different-random-numbers-in-r

Alternatively pass `...` to `sample()`

PR #1659, opened 2019-04-01
Fixes dfm_sample and a broken URL CRAN errors

kbenoit:
Solves #1658 

Verified on r-hub builder for Debian Linux, R-devel, GCC.


koheiw:
How about `corpus_sample()`?

kbenoit:
Probably because we do not test for a specific outcome, `corpus_sample()` tests are not affected.

PR #1664, opened 2019-04-03
Change output of textstat_simil/dist() to data.frame of pairs

koheiw:
This PR:

* Changes the output format of `textstat_simil()` and `textstat_dist()` to a data.frame of pairwise items (documents or features).
* Implements `as.matrix()` methods and make sure that `as.dist()` also works.
* Implements `min_simil` option for `textstat_simil()`.
* Removes some methods that produced irregular results (e.g. "faith" and "kullback").

Solves #1675.


koheiw:
@kbenoit  tell me how you like the behavior, before changing `textstat_dist` too.

R/textstat_simil.R
```diff
@@ -358,3 +379,37 @@ as.matrix.simil <- function(x, diag = 1.0, ...) {
...
     df
 }
+
+#' Coerce a textstat_simil or textstat_dist to a sparse Matrix format
+#' 
+#' Coerce a \link{textstat_simil} or \link{textstat_dist} to a sparse \pkg{Matrix} format.
+#' @param x input textstat object
+#' @keywords textstat internal
+#' @export
+as.Matrix <- function(x) {
```
kbenoit: The more **Matrix** package-like call would be to extend S4 methods for `as(x, "dgCMatrix")` and `as(x, "dsTMatrix")`.  This function is a bit indeterminate. My guess is that those wishing to coerce to these methods will be comfortable with the S4 functions extending **Matrix**.

Also note that this
```r
> textstat_simil(dfmat, method = "cosine", margin = "documents", tri = TRUE, min_simil = 0.50) %>%
+     as.Matrix()
5 x 5 sparse Matrix of class "dgCMatrix"
           2001-Bush 2005-Bush 2009-Obama 2013-Obama 2017-Trump
2001-Bush          1 0.5204355  0.5411649  0.5561972          .
2005-Bush          . 1.0000000  .          0.5163644          .
2009-Obama         . .          1.0000000  0.6373318          .
2013-Obama         . .          .          1.0000000          .
2017-Trump         . .          .          .                  1
```
runs afoul of the issues in https://github.com/quanteda/quanteda/issues/1455 for both the lower triangle and the thresholded out values, suggesting that these are zeroes when they are not.

A better approach would be an extension of `as(x, "dsTMatrix")` that avoids the intermediate dense matrix object here:
```r
> textstat_simil(dfmat, method = "cosine", margin = "documents", tri = FALSE) %>%
+     as.matrix() %>%
+     as("dsTMatrix")
5 x 5 sparse Matrix of class "dsTMatrix"
           2001-Bush 2005-Bush 2009-Obama 2013-Obama 2017-Trump
2001-Bush  1.0000000 0.5204355  0.5411649  0.5561972  0.4518935
2005-Bush  0.5204355 1.0000000  0.4575297  0.5163644  0.4349030
2009-Obama 0.5411649 0.4575297  1.0000000  0.6373318  0.4481950
2013-Obama 0.5561972 0.5163644  0.6373318  1.0000000  0.4546945
2017-Trump 0.4518935 0.4349030  0.4481950  0.4546945  1.0000000
```
and to disallow any sparse object method via **quanteda** at least that contains anything that looks like a zero. So we would only support the sparse symmetric version and we would have to guarantee that the diagonal is 1.0.

R/textstat_simil.R
```diff
@@ -358,3 +379,37 @@ as.matrix.simil <- function(x, diag = 1.0, ...) {
...
+#' @method as.matrix simil_pairwise
+#' @keywords textstat internal
+as.matrix.simil_pairwise <- function(x, ...) {
+    as.matrix(as.Matrix(x))
+}
+
+#' @export
+#' @method as.list simil_pairwise
+#' @keywords textstat internal
+as.list.simil_pairwise <- function(x, ...) {
```
kbenoit: Let's keep the full existing `as.list()` functionality, regardless of the value of `tri/upper`, i.e.
```r
as.list.simil_pairwise(x, sorted = TRUE, n = NULL, ...)
```

R/textstat_simil.R
```diff
@@ -344,11 +364,12 @@ as_dist <- function(x, method, call, diag = FALSE, upper = FALSE) {
 #' @export
 #' @method as.matrix simil
 #' @keywords textstat internal
-as.matrix.simil <- function(x, diag = 1.0, ...) {
+as.matrix.simil <- function(x, diag = 1.0, upper = TRUE, ...) {
```
kbenoit: We should also return any values thresholded out from `min_simil` as `NA` in the matrix. The 0's below should be `NA`, not 0.
```r
> textstat_simil(dfmat, method = "cosine", margin = "documents", tri = FALSE, min_simil = 0.50) %>%
+     as.matrix()
           2001-Bush 2005-Bush 2009-Obama 2013-Obama 2017-Trump
2001-Bush  1.0000000 0.5204355  0.5411649  0.5561972          0
2005-Bush  0.5204355 1.0000000  0.0000000  0.5163644          0
2009-Obama 0.5411649 0.0000000  1.0000000  0.6373318          0
2013-Obama 0.5561972 0.5163644  0.6373318  1.0000000          0
2017-Trump 0.0000000 0.0000000  0.0000000  0.0000000          1
```

kbenoit:
The fact that the Appveyor test failed above (but passed before) indicates there might be something stochastic or otherwise weird happening in the C++ code, so we should run some repetition tests to make sure this is entirely stable.

R/textstat_simil.R
```diff
@@ -13,10 +13,11 @@
 #'   \code{"features"} for word/term features.
 #' @param method method the similarity or distance measure to be used; see
 #'   Details.
-#' @param upper whether the upper triangle of the symmetric \eqn{V \times V}
```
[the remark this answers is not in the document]
koheiw: `upper` can be renamed `lower` as the new object only have upper triangle by default. I am fine with changing `fcm` in this way.

R/textstat_simil.R
```diff
@@ -13,10 +13,11 @@
...
-#'   matrix is recorded. Only used when \code{value = "dist"}.
-#' @param diag whether the diagonal of the distance matrix should be recorded. .
```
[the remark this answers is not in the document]
koheiw: `as.dist()` can have `diag` but resulting object of `textstat_simil()` should have all the information. I am strongly against making up values for `as.matrix()` and `as.Matrix()` because diagonals are not always 1.0 or 0 in some measures.

R/textstat_simil.R
koheiw: Need to find an efficient way to fill up a matrix with NA.

R/textstat_simil.R
koheiw: Users can sort the output just by `tstat2[order(tstat2$similarity),]` because it is a `data.frame`

R/textstat_simil.R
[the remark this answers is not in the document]
kbenoit: I think it makes more sense to use the argument as we currently use `tri` in `fcm()`, which is to restrict the output to the upper triangle only. So `upper = TRUE` means we return just the upper triangle, otherwise we return all A-B, B-A pairs and in their matrix form, the transpose of the upper triangle, in other words the full symmetric matrix.

So for `fcm()`, we change `tri` to `upper`, which returns just the upper triangle, which is the same behaviour as `tri`. (We trap `tri` and print a deprecation warning, but change the argument name in the function.)

For this branch of `textstat_simil()`, we restore the `upper` argument but change its meaning from the previous `dist`-like functionality that appeared to return just the lower triangle, to the behaviour similar to `fcm()`.

R/textstat_simil.R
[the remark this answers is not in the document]
kbenoit: Hmmm, I thought `simil(A, A)` always 1.0 but it's not for the `method = "faith"`. Because I think it should be, I filed #1675. If that's not right, then we definitely do not want to make up any values.

If we can exclude the diagonals from the output pairs, then we save as many pairs as their are ndoc or nfeat.

R/textstat_simil.R
kbenoit: Fine for the data.frame, but why change the behaviour of `as.list()`?

R/textstat_simil.R
kbenoit: Yes I wish there were such a way! But non-recorded values can only have one value, otherwise they'd need to be recorded.

R/textstat_simil.R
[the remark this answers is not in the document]
kbenoit: Solved in 4a052caf, by removing the faith measure. Now all simil(A,A) = 1.0.

R/textstat_simil.R
kbenoit: Fixed in d94cbd7d.

R/textstat_simil.R
kbenoit: Fixed in d94cbd7d.

R/textstat_simil.R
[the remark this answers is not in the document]
kbenoit: I added back the the `upper = TRUE` as a default. It works like this:

``` r
library("quanteda")
mt <- dfm(corpus_subset(data_corpus_inaugural, Year > 2006)) %>%
  dfm_trim(min_termfreq = 10)

tstat <- textstat_simil(mt, upper = TRUE, diag = TRUE)
tstat
##    document1  document2 similarity
## 1 2013-Obama 2009-Obama  0.9681414
## 2 2017-Trump 2009-Obama  0.9295036
## 3 2017-Trump 2013-Obama  0.9152045
## 4 2009-Obama 2009-Obama  1.0000000
## 5 2009-Obama 2013-Obama  0.9681414
## 6 2013-Obama 2013-Obama  1.0000000
## 7 2009-Obama 2017-Trump  0.9295036
## 8 2013-Obama 2017-Trump  0.9152045
## 9 2017-Trump 2017-Trump  1.0000000
as.matrix(tstat)
##            2009-Obama 2013-Obama 2017-Trump
## 2009-Obama  1.0000000  0.9681414  0.9295036
## 2013-Obama  0.9681414  1.0000000  0.9152045
## 2017-Trump  0.9295036  0.9152045  1.0000000
as.list(tstat)
## $`2009-Obama`
## 2009-Obama 2013-Obama 2017-Trump 
##  1.0000000  0.9681414  0.9295036 
## 
## $`2013-Obama`
## 2013-Obama 2009-Obama 2017-Trump 
##  1.0000000  0.9681414  0.9152045 
## 
## $`2017-Trump`
## 2017-Trump 2009-Obama 2013-Obama 
##  1.0000000  0.9295036  0.9152045

tstat <- textstat_simil(mt, upper = TRUE, diag = FALSE)
tstat
##    document1  document2 similarity
## 1 2013-Obama 2009-Obama  0.9681414
## 2 2017-Trump 2009-Obama  0.9295036
## 3 2017-Trump 2013-Obama  0.9152045
## 5 2009-Obama 2013-Obama  0.9681414
## 7 2009-Obama 2017-Trump  0.9295036
## 8 2013-Obama 2017-Trump  0.9152045
as.matrix(tstat) # always include the diagonal of simil(A,A) = 1.0
##            2009-Obama 2013-Obama 2017-Trump
## 2009-Obama  1.0000000  0.9681414  0.9295036
## 2013-Obama  0.9681414  1.0000000  0.9152045
## 2017-Trump  0.9295036  0.9152045  1.0000000
as.list(tstat) # (A,A) pairs not included
## $`2009-Obama`
## 2013-Obama 2017-Trump 
##  0.9681414  0.9295036 
## 
## $`2013-Obama`
## 2009-Obama 2017-Trump 
##  0.9681414  0.9152045 
## 
## $`2017-Trump`
## 2009-Obama 2013-Obama 
##  0.9295036  0.9152045

tstat <- textstat_simil(mt, upper = FALSE, diag = TRUE)
tstat
##    document1  document2 similarity
## 1 2009-Obama 2009-Obama  1.0000000
## 2 2009-Obama 2013-Obama  0.9681414
## 3 2013-Obama 2013-Obama  1.0000000
## 4 2009-Obama 2017-Trump  0.9295036
## 5 2013-Obama 2017-Trump  0.9152045
## 6 2017-Trump 2017-Trump  1.0000000
as.matrix(tstat) # upper is NA
##            2009-Obama 2013-Obama 2017-Trump
## 2009-Obama  1.0000000         NA         NA
## 2013-Obama  0.9681414  1.0000000         NA
## 2017-Trump  0.9295036  0.9152045          1
as.list(tstat) # same as with upper = TRUE
## $`2009-Obama`
## 2009-Obama 2013-Obama 2017-Trump 
##  1.0000000  0.9681414  0.9295036 
## 
## $`2013-Obama`
## 2013-Obama 2009-Obama 2017-Trump 
##  1.0000000  0.9681414  0.9152045 
## 
## $`2017-Trump`
## 2017-Trump 2009-Obama 2013-Obama 
##  1.0000000  0.9295036  0.9152045
as.dist(as.matrix(tstat)) # works
##            2009-Obama 2013-Obama
## 2013-Obama  0.9681414           
## 2017-Trump  0.9295036  0.9152045

tstat <- textstat_simil(mt, upper = FALSE, diag = FALSE)
tstat
##    document1  document2 similarity
## 2 2009-Obama 2013-Obama  0.9681414
## 4 2009-Obama 2017-Trump  0.9295036
## 5 2013-Obama 2017-Trump  0.9152045
as.matrix(tstat) # upper is NA
##            2009-Obama 2013-Obama 2017-Trump
## 2009-Obama  1.0000000         NA         NA
## 2013-Obama  0.9681414  1.0000000         NA
## 2017-Trump  0.9295036  0.9152045          1
as.list(tstat) # same as with upper = TRUE
## $`2009-Obama`
## 2013-Obama 2017-Trump 
##  0.9681414  0.9295036 
## 
## $`2013-Obama`
## 2009-Obama 2017-Trump 
##  0.9681414  0.9152045 
## 
## $`2017-Trump`
## 2009-Obama 2013-Obama 
##  0.9295036  0.9152045
as.dist(as.matrix(tstat)) # works
##            2009-Obama 2013-Obama
## 2013-Obama  0.9681414           
## 2017-Trump  0.9295036  0.9152045
```

R/textstat_simil.R
kbenoit: Restored the previous behaviour, although if `diag = FALSE`, then the coercion to list omits identity pairs.

5bf1269

kbenoit:
**Executive summary**

This was a balance between compatibility with the previous version but while also retaining the advantages of "sparsity" (by forming only pairs) including the ability to exclude pairs based on a threshold. 

With this change, and once we do the same for `textstat_dist()`, it means that all `textstat_*()` functions will return a data.frame, where the first two columns identify the units (documents or features).

- `teststat_simil()` now outputs a pairwise list, and the second column is the "selection" column, if any. This is a data.frame classed also as `textstat_simil`, and has attributes reflecting the settings. The columns are now named `document1`, `document2` or `feature1`, `feature2` depending on the `margin`.
- `min_simil` defaults to zero, meaning all are included; if higher, then any pair not meeting the minimum value is not part of the pairwise list. In coercion to matrix, a value not included in the pairwise list is marked `NA`.
- `upper = TRUE` is retained for compatibility with the previous function, and if `TRUE`, outputs the reverse of the pairs in the pairwise list, e.g. (A, B) and (B, A). If `FALSE`, then just (B, A) is output. When converting to a matrix, then the upper triangle is `NA` if the textstat_simil object was built with `upper = FALSE`.
- `diag = TRUE` is retained for compatibility, and includes the (A,A) pairs in the pairwise list if `TRUE`. However in the `as.matrix()` method, the diagonal is _always_ included, as 1.0. Because I removed the weird "faith" measure, it is now true for all of our similarity measures that simil(A,A) = 1.0, and the result has a lower bound of 0.
- `as.list()` works as before, and the `upper` has no effect on this, but `diag = TRUE` excludes the same-value similarity that is always 1.0.
- There is no exposed coercion to a **Matrix** package object, but the user can do this easily using e.g. `as(as.matrix(x), "dgCMatrix:")`.

R/textstat_simil.R
```diff
@@ -105,22 +123,48 @@ textstat_simil.dfm <- function(x, selection = NULL,
...
+    # eliminate identical pairs when diag = FALSE
+    if (!diag) {
+        result <- result[result[1] != result[2], ]
     }
-    result
+    
+    # replace x and y with margin names
+    names(result)[1:2] <- paste0(stri_sub(margin, 1, -2), 1:2)
+    
+    class(result) <- c("textstat_simil", "data.frame")
```
koheiw: Output from `textstat_collocations` is `c('collocations', 'textstat', 'data.frame')`, so this should be `c('simil', 'textstat', 'data.frame').

koheiw:
Handling of diagonals need to be much cleaner in all the functions, otherwise they will be bug prone. 
```
> mt <- dfm(corpus_subset(data_corpus_inaugural, Year > 1980))
> mt <- dfm_trim(mt, min_termfreq = 10)
> out <- textstat_simil(mt, c(2, 4), margin = "documents", diag = FALSE)
 Error in Ops.factor(left, right) : level sets of factors are different 
```
Here, we are implementing what `Matrix` (one of the best packages in R) does using a inferior object (data.frame). It cannot be easy and highly risky. This is why I stopped developing this branch.

koheiw:
My suggest is to change the output object to a (dense) `Matrix` without supporting `min_simil`. Only few people ever use that option, but we expose many users to potential bugs if we change to data.frame. This huge (potential) cost is overwhelming the benefit in my view.

R/textstat_simil.R
kbenoit: See #1690. Classing these as `simil` and `dist` would collide with return values from `stats::dist()` and `proxy::dist()`.

The subsetting seems to preserve the attributes fine. And conveniently, it does not print them.
```r
> dfmat <- dfm(corpus_subset(data_corpus_inaugural, Year > 2000), 
+              remove_punct = TRUE, remove = stopwords("english"))
> str(textstat_simil(dfmat, method = "cosine", margin = "documents")[1:3, ])
Classes ‘textstat_simil’ and 'data.frame':	3 obs. of  3 variables:
 $ document1 : Factor w/ 5 levels "2001-Bush","2005-Bush",..: 1 2 3
 $ document2 : Factor w/ 5 levels "2001-Bush","2005-Bush",..: 1 1 1
 $ similarity: num  1 0.52 0.541
 - attr(*, "method")= chr "cosine"
 - attr(*, "min_simil")= num 0
 - attr(*, "diag")= logi TRUE
 - attr(*, "upper")= logi TRUE
> textstat_simil(dfmat, method = "cosine", margin = "documents")[1:3, ]
   document1 document2 similarity
1  2001-Bush 2001-Bush  1.0000000
2  2005-Bush 2001-Bush  0.5204355
3 2009-Obama 2001-Bush  0.5411649
```

kbenoit:
I fixed the bug above by making the factor levels identical. On the output format, we have agreed that we would output this - and the other textstat objects - as data.frames. See #1690 however for my analysis of how well we have been implementing this.

On `min_simil`: The whole impetus for this change was the feature requests via #1210, so I am very surprised that you suggest dropping it. I think it works nicely now that we output just the pairs.

kbenoit:
In solving #1690 in a separate PR, we can tidy up the uniformity issue (and return `textstat_entropy()` values for instance as a data.frame) but for the reasons I explain in that issue, I'd favor the class name as `textstat_simil` since this avoids S3 method name collisions with other packages.

koheiw:
Let me withdraw #1210. If there are advanced users who ask me about computing similarity in large corpora, I will refer them to **proxyC**. I am fine with any class labels as far as they are named systematically.

kbenoit:
I think this is a valuable feature, and the alternative is 25% slower and more memory-intensive. Without being able to set a threshold, we have to score everything and then sort and slice off the top ones from a list. Why would we not want to enable this?

```r
library("quanteda")
data_dfm_sotu <- dfm(data_corpus_sotu)

nuclsimil <- textstat_simil(data_dfm_sotu,
  margin = "features",
  selection = c("nuclear"),
  method = "cosine",
  min_simil = 0.90,
  diag = FALSE
)
nuclsimil
##        feature1 feature2 similarity
## 1    developing  nuclear  0.9118093
## 2    commitment  nuclear  0.9106826
## 4        sector  nuclear  0.9022622
## 5 proliferation  nuclear  0.9051895
## 6       airline  nuclear  0.9022165
## 7      rhodesia  nuclear  0.9040306

# alternative
nuclsimilall <- textstat_simil(data_dfm_sotu,
  margin = "features",
  selection = "nuclear",
  method = "cosine",
  diag = FALSE
)
as.list(nuclsimilall, sorted = TRUE, n = 7)
## $nuclear
##    developing    commitment proliferation      rhodesia        sector 
##     0.9118093     0.9106826     0.9051895     0.9040306     0.9022622 
##       airline    capability 
##     0.9022165     0.8922810

  minsimil = textstat_simil(data_dfm_sotu,
    margin = "features",
    selection = c("nuclear"),
    method = "cosine",
    min_simil = 0.90,
    diag = FALSE
  ),
  allsimil = textstat_simil(data_dfm_sotu,
    margin = "features",
    selection = "nuclear",
    method = "cosine",
    diag = FALSE
  ) %>%
    as.list(sorted = TRUE, n = 7),
  times = 30,
  unit = "relative"
)
## Unit: relative
##      expr      min       lq     mean median      uq      max neval
##  minsimil 1.000000 1.000000 1.000000 1.0000 1.00000 1.000000    30
##  allsimil 1.246645 1.250244 1.281888 1.2648 1.27871 1.475134    30
```

kbenoit:
In 661b1c0 I've now aligned the behaviour of `textstat_dist()` with that of `textstat_simil()`. There is no thresholding argument for `textstat_dist()` but this makes sense since distance scales differ and have no upper bound.

I removed Kullback-Liebler distance as an option so that we only support "conservative" distance measures that obey the basic distance axioms, namely dist(A, A) = 0 and dist(A, B) = dist(B, A). (KL violates the latter, and the current implementation violates the former - see #1693.)

Once we merge this PR we can close a lot of the dist/simil-related issues and see what is left to solve in this area. For instance, we can replace `textstat_proxy()` with the equivalent functions in **proxyC**, and if users want more aggressive dist or simil measures, they can use that package directly on the dfm instead of the `textstat_()` wrappers.

koheiw:
Performance is not my main concern but please check the difference between `textstat_proxy()` and `textstat_simil()`.

kbenoit:
You're right, the factorization and some of the other steps are causing a big performance hit. Let me work with this a bit more.

kbenoit:
Replaced by #1694.

PR #1686, opened 2019-05-12
Make behaviour of kwic.corpus() and kwic.tokens() identical

### LINKED ISSUE #1684: textplot_xray() from kwic() based on dictionary pattern
I thought that in https://github.com/quanteda/quanteda/pull/1515, we fixed kwics to plot (in `textplot_xray()`) by dictionary key, not value. But it appears to be faceting on values, not keys.

``` r
library("quanteda")
## Package version: 1.4.4

# one facet per value
dict <- dictionary(list(
  preps = c("to", "of", "for"),
  articles = c("the", "an", "an")
))
corp <- corpus_subset(data_corpus_inaugural, Year %in% c(1945, 1865))
kwicdict <- kwic(corp, pattern = dict)
textplot_xray(kwicdict)
```

``` r

# one facet per key
tokens(corp) %>%
  tokens_lookup(dictionary = dict, exclusive = FALSE, capkeys = TRUE) %>%
  kwic(pattern = char_toupper(names(dict))) %>%
  textplot_xray()
```

Is that what we intended? The second form is not likely to be easy for users to apply.

koheiw:
Fix #1684

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: quanteda
-Version: 1.4.4
+Version: 1.4.5
 Title: Quantitative Analysis of Textual Data
 Description: A fast, flexible, and comprehensive framework for 
     quantitative text analysis in R.  Provides functionality for corpus management,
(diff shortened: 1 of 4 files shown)


kbenoit:
Thanks @koheiw !

Issue #1693, opened 2019-05-21
textstat_dist(x, method = "kullback") returns invalid values


kbenoit:
## Describe the bug

Kullback distance cannot be less than 1, but it appears as negative in the example below in a great number of cases.

## Reproducible code

``` r
library("quanteda")
## Package version: 1.4.3

mt <- dfm(corpus_subset(data_corpus_inaugural, Year > 1980))

distmat <- textstat_dist(mt, margin = "features", method = "kullback") %>%
  as.matrix()
sum(distmat < 0)
## [1] 619162
distmat[1:10, 1:10]
##              senator   hatfield            ,          mr            .
## senator    0.0000000  1.3862944 -0.347928096  0.44927141 -0.341245230
## hatfield   1.3862944  0.0000000 -0.264871697 -0.35434990 -0.255595351
## ,         -0.3479281 -0.2648717  0.000000000  1.03262918  0.008365354
## mr         0.4492714 -0.3543499  1.032629178  0.00000000 -0.293036148
## .         -0.3412452 -0.2555954  0.008365354 -0.29303615  0.000000000
## chief     -0.3182571 -0.2986266  0.511131414 -0.07763809  0.488288516
## justice   -0.2996782 -0.1666596  0.533622093  0.23838327  0.491809698
## president -0.2843137 -0.2900466  0.138794635 -0.27086534  0.125356594
## vice      -0.2207594 -0.3579323  0.498428209 -0.04896630  0.495687240
## bush      -0.2506796 -0.2986266  0.495314623  0.01894322  0.534843458
##                  chief      justice  president        vice        bush
## senator   -0.318257084 -0.299678162 -0.2843137 -0.22075942 -0.25067957
## hatfield  -0.298626578 -0.166659608 -0.2900466 -0.35793228 -0.29862658
## ,          0.511131414  0.533622093  0.1387946  0.49842821  0.49531462
## mr        -0.077638091  0.238383266 -0.2708653 -0.04896630  0.01894322
## .          0.488288516  0.491809698  0.1253566  0.49568724  0.53484346
## chief      0.000000000 -0.000926755 -0.1492864  0.06591290  0.00000000
## justice   -0.000926755  0.000000000  0.1880273  0.47439294  0.39567601
## president -0.149286445  0.188027346  0.0000000  0.39111046  0.60782149
## vice       0.065912897  0.474392944  0.3911105  0.00000000 -0.03844919
[...]

Issue #1541, opened 2019-01-05
textstat_dist/simil() should return appropriate values for zero-variance inputs


kbenoit:
We discussed this in #1530 and established that:

* For some measures, such as cosine similarity, the similarity of any vector with a vector of all zeros should be `NaN`, since it involves 0/0. Here, **proxy** just has it wrong. (In Python's numpy `spatial.distance.cosine()` however this correctly produces a warning and returns `nan`.) As I noted in comments in `test-textstat_simil.R`, I think the way that we currently and **proxy** compute edice is also wrong for a pair of zero-only vectors.

There are skipped tests for this in 
https://github.com/quanteda/quanteda/blob/c27aeab9c83b02c1f971093aeb784a14b6f122a2/tests/testthat/test-textstat_simil.R#L225-L342 but accepted the PR while skipping those tests. Some of those tests might be wrong in their expectations, for instance we need to consider very carefullly, apart from what is established in **proxy**, what the expectations for Jaccard, eJaccard, and Dice should be in these edge cases.

Note also that some of our benchmark tests against e.g. **proxy** were shown to be wrong, such as 
https://github.com/quanteda/quanteda/pull/1530#issuecomment-451014579.

Issue #1540, opened 2019-01-05
textstat_dist/simil() should return NA for empty feature dfms


kbenoit:
We discussed this in #1530 and established that:

* Definitionally, dist or simil for an empty dfm (documents in rows but no features) should be NA for the document margin, and produce an error for the features margin (since by design it should be impossible to have an "empty" dfm with features but no documents).

There are skipped tests for this in 
https://github.com/quanteda/quanteda/blob/c27aeab9c83b02c1f971093aeb784a14b6f122a2/tests/testthat/test-textstat_simil.R#L180-L253 but accepted the PR while skipping those tests.

Note also that some of our benchmark tests against e.g. **proxy** were shown to be wrong, such as 
https://github.com/quanteda/quanteda/pull/1530#issuecomment-451014579.

koheiw:
See `?getOption("matprod")`:

> matprod:
a string selecting the implementation of the matrix products %*%, crossprod, and tcrossprod for double and complex vectors:
> 
> "internal"
uses an unoptimized 3-loop algorithm which correctly propagates NaN and Inf values and is consistent in precision with other summation algorithms inside R like sum or colSums (which now means that it uses a long double accumulator for summation if available and enabled, see capabilities).
> 
> "default"
uses BLAS to speed up computation, but to ensure correct propagation of NaN and Inf values it uses an unoptimized 3-loop algorithm for inputs that may contain NaN or Inf values. When deemed beneficial for performance, "default" may call the 3-loop algorithm unconditionally, i.e., without checking the input for NaN/Inf values. The 3-loop algorithm uses (only) a double accumulator for summation, which is consistent with the reference BLAS implementation.
> 
> "blas"
uses BLAS unconditionally without any checks and should be used with extreme caution. BLAS libraries do not propagate NaN or Inf values correctly and for inputs with NaN/Inf values the results may be undefined.
> 
> "default.simd"
is experimental and will likely be removed in future versions of R. It provides the same behavior as "default", but the check whether the input contains NaN/Inf values is faster on some SIMD hardware. On older systems it will run correctly, but may be much slower than "default".

kbenoit:
This is still unresolved, although it works for "cosine" and "correlation", it returns zeros for the other types. See the skipped tests in `test-textstat_simil.R` and `test-textstat_dist.R`.

PR #1694, opened 2019-05-23
Change output of textstat_simil/dist to sparse Matrix format

### LINKED ISSUE #1210: Add min threshold to textstat_dist()?

### LINKED ISSUE #1469: Conversion to Matrix to dist is talking too much time
More than half of the execution time of `textstat_simil()` is coming from object conversion :unamused: 
```r
> quanteda_options(threads = 8)
> mt <- dfm(data_corpus_inaugural)
+   textstat_simil_old(mt, margin = "features"),  
+   #textstat_simil(mt, margin = "features"),
+   #textstat_simil(mt, margin = "features", min_simil = 0.9),
+   textstat_simil(mt, margin = "features"),
+   textstat_proxy(mt, margin = "features", rank = 3),
+   #qlcMatrix::cosSparse(mt),
+   times = 10
+ )
Unit: milliseconds
                                              expr        min         lq       mean     median         uq        max neval
       textstat_simil_old(mt, margin = "features")  7251.1911  7261.6533  7481.1757  7381.5864  7471.9465  8520.6013    10
           textstat_simil(mt, margin = "features") 11935.1468 12059.3998 12184.3832 12215.5913 12300.8472 12398.1135    10
 textstat_proxy(mt, margin = "features", rank = 3)   549.5476   557.0611   565.2102   561.9871   571.3744   585.1749    10
```
screenshot_20181025_123351

### LINKED ISSUE #1472: Call for feedback on textstat_proxy()

### LINKED ISSUE #1675: textstat_simil(proxy, method = "faith") should return 1.0 for similarity of a document with itself
All other measures of similarity return 1.0 for the similarity of a vector with itself. Except "Faith", which appears to penalize similarity as identical vectors grow in length. I suspect this is an error in the formula, unless I have misunderstood [Faith, D. P. (1983). Asymmetric binary similarity measures. Oecologia, 57(3), 287-290.](https://link.springer.com/article/10.1007/BF00377169). 

In Faith (1983) it states that the formula is (a + d/2) / n, where a is (TRUE, TRUE) for a match and d is (FALSE, FALSE). For a vector with itself, a = n and d = 0, which should evaluate to 1.0. But this is not what happens.

``` r
library("quanteda")
## Package version: 1.4.4

set.seed(100)
x <- rpois(1000, lambda = 3)
dfmat <- as.dfm(matrix(c(x, x), nrow = 2, byrow = TRUE))
textstat_simil(dfmat, method = "faith")
##       x     y similarity
## 1 text2 text1      0.972
## 2 text1 text1      0.972
## 3 text1 text2      0.972
## 4 text2 text2      0.972

dfmat <- as.dfm(matrix(c(x[1:300], x[1:300]), nrow = 2, byrow = TRUE))
textstat_simil(dfmat, method = "faith")
##       x     y similarity
## 1 text2 text1      0.975
## 2 text1 text1      0.975
## 3 text1 text2      0.975
## 4 text2 text2      0.975
```

### LINKED ISSUE #1689: Unstable behaviour in textstat_simil()
In rare cases, that cannot be reproduced by setting the seed, but can be discovered by re-running the code below, some values are different.

``` r
library("quanteda")
# matrix for testing
mt <- dfm(corpus_subset(data_corpus_inaugural, Year > 1980))
mt <- dfm_trim(mt, min_termfreq = 10)
res <- textstat_simil(mt, c(2, 4), margin = "documents")

set.seed(10)
resv <- list()
for (i in 1:1000) {
  resv[[i]] <- textstat_simil(mt, c(2, 4), margin = "documents")$similarity
}
rescols <- do.call(cbind, resv)

plot(rescols[12, ], type = "b", ylim = c(.8, 1))
```

``` r
plot(rescols[1, ], type = "b", ylim = c(.8, 1))
```

kbenoit:
This PR:

* Changes the output format of `textstat_simil()` and `textstat_dist()` to a specially classed `textstat_simildist` S4 object that inherits from sparse **Matrix** classes. There are child classes of this depending on whether it is a simil or dist object, whether the y dimension has been selected, and whether `min_simil > 0`.

    Class structure:

| object                      | based on      | `selection` | `min_simil` |
|-----------------------------|---------------|:-----------:|:-----------:|
| `textstat_simildist`        | dsparseMatrix |      -      |      -      |
| `textstat_simil`            | dsTMatrix     |      𝗑      |      0      |
| `textstat_simil_sel`        | dgTMatrix     |      ✓      |      0      |
| `textstat_simil_sparse`     | dsCMatrix     |      𝗑      |     > 0     |
| `textstat_simil_sel_sparse` | dgCMatrix     |      ✓      |     > 0     |
| `textstat_dist`             | dsTMatrix     |      𝗑      |      -      |
| `textstat_dist_sel`         | dgTMatrix     |      ✓      |      -      |

* Verifies that `as.dist()` and `as.matrix()` work on all of these objects, the former from **base** and the latter from the **Matrix** package functions for the parent sparse Matrix classes.

* Removes the `diag` and `upper` arguments from the `textstat_simil/dist()` functions. (However: We should probably trap and deprecate these via `...`.)

* Implements `as.list(x, sorted = TRUE, n = NULL, diag = FALSE, ...)` for the output objects, where `diag = TRUE` means that same-pair values (which are always either 0.0 or 1.0) are included.

* Implements `as.data.frame(x, diag = FALSE, upper = FALSE)` to convert the output object to a pairwise list that is identical the object output class that we developed for #1664. Setting `upper = TRUE` means that both (A, B) and (B, A) pairs are included.

* Implements `min_simil` option for `textstat_simil()`.

* Removes some methods that produced irregular results (e.g. "faith" and "kullback").

Closes #1675 
Closes #1469
Closes #1472 
Closes #1210 
Closes #1689
Closes #1675 

Still need to check: #1540, #1541 

Need to solve separately, although no longer an issue for `textstat_simil`: #1693


koheiw:
There are too many classes (reminds me of `dfmSparse` and `dfmDense`). If you wrap a Matrix by a list, you only need a single S3 class for a simil/dist object.

kbenoit:
But then you would need to write methods for every matrix operation for that list wrapper class, instead of using the ones already common to Matrix classes. It looks complex but the user will never need this information. Classing it this way preserves the symmetry (saving half the space) from just making them all dgCMatrix. 

So: it’s as simple as a normal matrix for the user, but gives us the hidden control we need.

I like the validators and more robust slots from the S4 classes too.

koheiw:
I thought you want to stop people from doing matrix operation (as in data.frame), although we can define `as.Matrix()` to make it possible.  I prefer Matrix objects without custom class labels.

koheiw:
The best solution would be to ask a third-party package to do the conversation to data.frame. https://github.com/tidymodels/broom/issues/470

koheiw:
I am note sure why `update-simil-Matrix` was merged. I did not mean to do this by #1699...

Issue #1461, opened 2018-10-24
fcm(ordered = TRUE) is not working properly


koheiw:
I expect ordered fcm to be asymmetric, but it is not. Also we should set `tri = FALSE` automatically when `ordered = TRUE`.
```r
> fcm(c("a b c", "a b c"), "window", window = 2, ordered = TRUE)
Feature co-occurrence matrix of: 3 by 3 features.
3 x 3 sparse Matrix of class "fcm"
        features
features a b c
       a 0 2 2
       b 0 0 2
       c 0 0 0
> fcm(c("a b c", "a b c"), "window", window = 2, ordered = TRUE, tri = FALSE)
Feature co-occurrence matrix of: 3 by 3 features.
3 x 3 sparse Matrix of class "fcm"
        features
features a b c
       a 0 2 2
       b 0 0 2
       c 0 0 0
```

kbenoit:
Not sure I understand here, since both of those matrices are in fact asymmetric, and correct for what we intended for `ordered = TRUE`.

The lower triangle is all 0 because no column feature in the lower part comes after its row feature. If we add a new feature to the second document, it does show up correctly (and asymmetrically):
```r
> fcm(c("a b c", "d a b c"), "window", window = 2, ordered = TRUE)
Feature co-occurrence matrix of: 4 by 4 features.
4 x 4 sparse Matrix of class "fcm"
        features
features a b c d
       a 0 2 2 0
       b 0 0 2 0
       c 0 0 0 0
       d 1 1 0 0
```

The documentation already states
> `tri` if `TRUE` return only upper triangle (including diagonal). Ignored if `ordered = TRUE`.

PR #1707, opened 2019-06-12
Issue 1461

koheiw:
I fully inspected the code for #1461, and made changes where improvement is possible.

### DIFF
--- a/tests/testthat/test-fcm.R
+++ b/tests/testthat/test-fcm.R
@@ -260,10 +260,17 @@ test_that("fcm works the same for different object types", {
     expect_identical(fcm(txt), fcm(tokens(txt)))
 })
 
-test_that("fcm expects warning for a wrong weights length", {
+test_that("fcm expects error for wrong weight or window", {
     txt <- c("a a a b b c", "a a c e", "a c e f g")
-    expect_warning(fcm(tokens(txt), context = "window", window = 2, count = "weighted", weights = c(1,2,3)),
-                 "weights length is not equal to the window size, weights are assigned by default!")
+    toks <- tokens(txt)
+    expect_error(fcm(toks, context = "window", window = 0),
+                "window size must be at least 1")
+    expect_error(fcm(toks, context = "window", window = 2, weight = 0.1,
+                     count = "weighted", weights = c(1,2,3)),
+                 "weights length must be equal to the window size")
+    expect_error(fcm(toks, context = "window", window = 2, weight = c(0.1, 0.2, 0.3),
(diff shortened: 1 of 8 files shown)


koheiw:
I did not change the tests, but I am not sure about the diagonal counts.

For example, I think A-A should be 4 instead of 2

https://github.com/quanteda/quanteda/blob/8b7d2235fb13a10ecd6ebf40cdf778e67fa77e0d/tests/testthat/test-fcm.R#L34-L45

koheiw:
I removed `span_sentence` because it was not dong anything. We can add this provide this functionality but, in principle, users should segment documents in to sentences beforehand.

tests/testthat/test-fcm.R
```diff
@@ -1,31 +1,42 @@
 context('Testing fcm*.R')
 
-test_that("compare the output feature co-occurrence matrix to that of the text2vec package", {
-    skip_if_not_installed("text2vec")
-    library("text2vec")
-    
+# test_that("compare the output feature co-occurrence matrix to that of the text2vec package", {
```
kbenoit: Our diagonals are different but does it make sense to remove this comparison? (They match, except the diagonal.)

If we do remove this code then we can remove **text2vec** from `Suggests:`.

kbenoit:
Moving a comment here from another issue where I pasted it by mistake! Just for the record, since I am happy with the removal of this argument.

That's true, I had forgotten we did not implement `span_sentence = FALSE`. I think the idea was to make it easier than this:

``` r
library("quanteda")
## Package version: 1.4.3
txt <- "A b c d.  A a b."

fcm(txt, context = "window", size = 3, span_sentence = FALSE)
## Warning in fcm.tokens(tokens(x), ...): spanSentence = FALSE not yet
## implemented

corpus(txt) %>%
  tokens(remove_punct = TRUE) %>%
  tokens_tolower() %>%
  fcm(context = "window", size = 3)
##         features
## features a b c d
##        a 1 3 1 1
##        b 0 0 1 1
##        c 0 0 0 1
##        d 0 0 0 0

# compare to:
tokens(txt, remove_punct = TRUE) %>%
  tokens_tolower() %>%
  fcm(context = "window", size = 3, span_sentence = TRUE)
##         features
## features a b c d
##        a 3 5 3 3
##        b 0 1 2 2
##        c 0 0 0 1
##        d 0 0 0 0

```



But I agree this ought to be a pre-processing step, separate from `fcm()`.

The lack of a `tolower` option in `fcm()` suggests we should add one, or consider removing it from `dfm()`. Maybe it's time to consider pruning some of the shortcut options in `dfm()`...

Issue #1375, opened 2018-06-02
Nested (overlapped) elements to tokens_lookup()


R01010010R:
tokens_lookup() returns more than value if a dictionary has overlapping elements (see https://github.com/koheiw/newsmap/issues/13#issuecomment-393322971).

There is a workaround which works.
1. Use `tokens_lookup` with the option `valuetype = 'glob'`.
2. Check if the tokens also return at least one of the values when I use `tokens_lookup` with the option `valuetype = 'fixed'`. And then drop the other values from (1).

However, an option such as the one @koheiw suggested might be more efficient and easier to implement.

plugrafico:
Can dictionaries by assembled so that categories follow logical commands to ignore hits by neighboring categories?
For instance (just exemplifying, not really sure how the logical expressions should look like):
```
dict <- dictionary(list(
guinea_bissau="guinea bissau", 
equatorial_guinea="equatorial guinea",
guinea="guinea(NOT 'guinea_bissau' NOT 'equatorial_guinea')")

**Comment by kbenoit** [comment]
Really should be a SO question, but anyway:

The easiest way would be to compound the tokens you want treated as single items. You can also look up sequences and match them using dictionaries.

``` r
library("quanteda")
## Package version: 1.4.3

txt <- c("I spent one guinea while visiting Equatorial Guinea and then went to Guinea Bissau")
toks <- tokens(txt)

# method 1: compound the tokens first
tokens_compound(toks, phrase(c("equatorial guinea", "guinea bissau")))
## tokens from 1 document.
## text1 :
##  [1] "I"                 "spent"             "one"              
##  [4] "guinea"            "while"             "visiting"         
##  [7] "Equatorial_Guinea" "and"               "then"             
## [10] "went"              "to"                "Guinea_Bissau"

# method 2: define just the compound word dictionaries
dict <- dictionary(list(
  Equatorial_Guinea = "equatorial guinea",
  Guinea_Bissau = "guinea bissau"
))
tokens_lookup(toks, dict, exclusive = FALSE)
## tokens from 1 document.
## text1 :
##  [1] "I"                 "spent"             "one"              
##  [4] "guinea"            "while"             "visiting"         
##  [7] "EQUATORIAL_GUINEA" "and"               "then"             
## [10] "went"              "to"                "GUINEA_BISSAU"
```

Set `capkeys = FALSE` to turn off the capitalization.

plugrafico:
Thanks for clearing that, @kbenoit

PR #1708, opened 2019-06-13
Add overlap argument for #1375

koheiw:


koheiw:
Not entirely sure about the name of the argument, but `overlap = "global"` offers desired behavior.

```r
test_that("overlap function is working", { 
    dict <- dictionary(list(
        'AS' = c("American Samoa", "American Samoan*", "Pago Pago"),
        'WS' = c("Samoa", "Samoan*", "Apia"),
        'VG' = c("British Virgin Islands", "Virgin Island*", "Road Town"),
        'GB' = c("UK", "United Kingdom", "Britain", "British", "Briton*", "Brit*", "London"),
        'US' = c("United States", "US", "American*", "Washington", "New York")
    ))
    txt <- c(
        'British Virgin Islands is a British overseas territory',
        'Samoa is an independent state',
        'American Samoa is in the South Pacific'
    )
    toks <- tokens(txt)
    expect_equal(
        as.list(tokens_lookup(toks, dict, overlap = "local")),
        list(text1 = c("VG", "GB", "GB"), 
             text2 = c("WS"), 
             text3 = c("AS", "US", "WS"))
    )
    expect_equal(
        as.list(tokens_lookup(toks, dict, overlap = "global")),
        list(text1 = c("VG", "GB"), 
             text2 = c("WS"), 
             text3 = c("AS"))
    )
    expect_equal(
        as.list(tokens_lookup(toks, dict, overlap = "none")),
        list(text1 = c("VG", "GB", "GB", "VG", "GB", "GB"), 
             text2 = c("WS"), 
             text3 = c("AS", "US", "WS"))
    )
})
```

koheiw:
I decided to hide "none" until someone asks us. I also changed to ignore partial overlap when "global".

```r
>     toks <- tokens("Virgin Islands are near Dominica and the Dominican Republic")
>     dict <- dictionary(list("VG" = "Virgin Islands",
+                             "VI" = "Virgin Islands",
+                             "DM" = "Dominica*",
+                             "DO" = "Dominican Republic"))
> as.list(tokens_lookup(toks, dict, overlap = "global"))
$text1
[1] "VG" "VI" "DM" "DO"

> as.list(tokens_lookup(toks, rev(dict), overlap = "global"))
$text1
[1] "VI" "VG" "DM" "DO"
```

kbenoit:
That's true, I had forgotten we did not implement `span_sentence = FALSE`. I think the idea was to make it easier than this:

``` r
library("quanteda")
## Package version: 1.4.3
txt <- "A b c d.  A a b."

fcm(txt, context = "window", size = 3, span_sentence = FALSE)
## Warning in fcm.tokens(tokens(x), ...): spanSentence = FALSE not yet
## implemented
## Feature co-occurrence matrix of: 6 by 6 features.
## 6 x 6 sparse Matrix of class "fcm"
##         features
## features A b c d . a
##        A 1 3 2 2 3 1
##        b 0 0 2 2 3 2
##        c 0 0 0 1 1 1
##        d 0 0 0 0 2 1
##        . 0 0 0 0 1 2
##        a 0 0 0 0 0 0

corpus(txt) %>%
  tokens(remove_punct = TRUE) %>%
  tokens_tolower() %>%
  fcm(context = "window", size = 3)
##         features
## features a b c d
##        a 1 3 1 1
##        b 0 0 1 1
##        c 0 0 0 1
##        d 0 0 0 0
```

But I agree this ought to be a pre-processing step, separate from `fcm()`.

The lack of a `tolower` option in `fcm()` suggests we should add one, or consider removing it from `dfm()`. Maybe it's time to consider pruning some of the shortcut options in `dfm()`...

koheiw:
I did not do anything with the exclusive mode in this PR, but I can do something similar if you want. 

On `match_overlap = c("separate", "order")`, we made lookup functions independent of the order of the values in dictionaries, but dependent on the order of tokens. Making the lookup function dependent on the order of values requires redesigning the entire C+ code+, so let's not do this in this PR. 

On `match_nested = c("all", "longest")`, priority to longer pattern is only in the `global` mode, not in the `local` mode. If `match_nested = "all"` then, we have to record the same keys for nested patterns, but I don't think this is what we want.

kbenoit:
I think the behaviour should be the same irrespective of the value of `exclusive`, unless you think there is a good reason for treating things differently when `exclusive = FALSE`.

`match_overlap`: I didn't mean to imply that we should do anything with the order of dictionary values, but as you have implemented this PR, we would have the behaviour that overlapping sequences are counted separately (the default of `"separate"`). Matching the first overlap and not the second would be the behaviour when `match_overlap = "order"`, which I think is easy to restore. (Dictionary value order should never matter though!)

The idea behind `match_nested = c("all", "longest")` was an attempt to rename local and global. Longest = global, and all = local. In my interpretation these are about nesting though and not overlap. Did I misunderstand the "local" part? (Possible!)

koheiw:
The current rule for multiple match is (1) ignore all the full and partial overlaps within the same key when "local" and "global" and (2) ignore only full overlaps (nesting) for different keys when "global".

koheiw:
`match_overlap` is possible but let's do it separately. I am not sure when people use this.

koheiw:
It looks like this now
```r
> tokens_lookup(toks, dict, overlap = "global", exclusive = FALSE)
tokens from 1 document.
text1 :
 [1] "Congress"   "passed"     "a"          "MONEY"      ","          "signed"     "by"        
 [8] "PRES"       ","          "to"         "strengthen" "the"        "LAW"        "."         

> tokens_lookup(toks, dict, overlap = "global", exclusive = TRUE)
tokens from 1 document.
text1 :
[1] "money" "pres"  "law" 
```

kbenoit:
> The current rule for multiple match is (1) ignore all the full and partial overlaps within the same key when "local" and "global" and (2) ignore only full overlaps (nesting) for different keys when "global".

(1) is what we have always wanted and never questioned, since within the same key, the rule is to count a match to the key if any of its values matches the token. 
```r
> tokens(txt) %>%
+     tokens_lookup(dict, exclusive = FALSE)
tokens from 1 document.
text1 :
[1] "X"   "and" "X" 
```

Let's leave `match_overlap` out for now and just keep this behaviour without an option to change it.

kbenoit:
The only open question then is what to call the added option. It's more nesting than overlap. `overlap = "global"` doesn't clearly state to me what it does. I'm not 100% happy with my counterproposal but it's more descriptive.

```r
match_nested = c("all", "longest")
```
where "all" replaces "local" and "longest" replaces "global". WDYT?

koheiw:
Even if it is "global",  "local" behavior do not change, so it can be TRUE or FALSE unless we have a plan to add more modes. For example, `ignore_nested`.

kbenoit:
What would other modes be?  Maybe: `match_nested = "shortest"` to match the shortest first, not the longest? Which probably makes little sense.

Boolean version: `match_nested = FALSE`  = "local", `TRUE` = "global".

That's still less than clearly descriptive though. This is a hard example of how to name things!

(Something like) `match_nested = c("all", "longestfirst")` seems more clear and allows for additional modes, if we can think of any down the road that make sense.

kbenoit:
Can you also check the C++ code block that has no tests and that is causing the codecov/patch check to fail?

koheiw:
We should make it clear that the longest-first rule apply to values for different keys. It applies to values in the same key but they simply do not notice as the output (keys) is the same. "all" sounds like the "none" mode that I dropped earlier.

kbenoit:
Agreed - then what should we call this argument and its values?

koheiw:
The most descriptive name is `ignore_overlap = c("within-keys", "across-dictionary")`. 

Then, `overlap_levels = c("key", "dictionary")`?

If it confusing with `levels`, `overlap_detection` or `overlap_rules`.

kbenoit:
I think this confuses overlap with duplicate matching. Duplicate matching within key should never result in multiple lookups. Matching multiple keys should always result in lookups for each key, because dictionaries are unordered. It’s up to a user to enforce unique key matching, although we could help this by adding warnings. So I don’t think we need to change anything about duplicate matching. 

Nesting occurs when one pattern matches another, say key1 = “a b”, key2 = “abc”. One behaviour is to match both (“local”) and the other is to match the longer (“global”). This could be considered a special case of duplicate matching but it’s different because there is the possibility of a priority, which there is not with simple duplicates. And this is what I think we should focus on in this PR. 

Overlap is a third thing which in my mind means: key1 = “a b”, key2 = “b c”, and tokens are “a b c”. Two options are return both or return the first. I’m okay with leaving this for now to the default behaviour which is to match both. But my point earlier is that this is separate from nesting and can be handled separately. 

Anyway TMI maybe but it helps to try to be as clear as possible. Happy to have a chat about this if that would be simpler.

koheiw:
😕

koheiw:
Actually, we can control several things independently

local: none/full overlap/partial overlap
global: none/full overlap/partial overlap

See `any_of` (partial overlap) and `all_of` (full overlap) in the code:
https://github.com/quanteda/quanteda/blob/3a5cf7a3cd24c650c01895248f8503f25733895d/src/tokens_lookup_mt.cpp#L36-L63

but those will users mad.

kbenoit:
Decided: Rename to `nested_keys = c("all", "longest")`, to correspond to PR behaviour of "local" and "global", respectively.

Issue #1663, opened 2019-04-03
Separate textmodel() functions into a new package


kbenoit:
**quanteda** is becoming very large, and it might be time to modularize the packages into a "quanteda-verse" set of packages that split its functionality. This would operate similar to the **tidyverse** package, which loads all of the packages in the tidyverse, i.e.
```r
> library("tidyverse")
── Attaching packages ─────────────────────────────────────────── tidyverse 1.2.1 ──
✔ ggplot2 3.1.0       ✔ purrr   0.3.2  
✔ tibble  2.1.1       ✔ dplyr   0.8.0.1
✔ tidyr   0.8.3       ✔ stringr 1.4.0  
✔ readr   1.3.1       ✔ forcats 0.4.0  
── Conflicts ────────────────────────────────────────────── tidyverse_conflicts() ──
✖ dplyr::filter() masks stats::filter()
✖ dplyr::lag()    masks stats::lag()
```
This would involve:

| package                   | depends on            | function                                                                                             |
|:--------------------------|:----------------------|:-----------------------------------------------------------------------------------------------------|
| **quanteda**              |                       | loads all quanteda packages                                                                          |
| **quanteda.core**         |                       | corpus, tokens, dfm functions, `dictionary()`, `kwic()` -- basically, everything not specified below |
| **quanteda.textmodel**    | **quanteda**          | all `textmodel_*()` functions                                                                        |
| **quanteda.textstat**     | **quanteda**          | all `textstat_*()` functions                                                                         |
| **quanteda.textplot**     | **quanteda**          | all `textplot_*()` functions                                                                         |
| **quanteda.dictionaries** | **quanteda**          | (already exists)                                                                                     |
| **quanteda.stm**          | **quanteda**, **stm** | could reimplement the infrastructure needed for **stm**                                              |

**Advantages**

- size management
- isolation of potential errors and easier maintenance
- easier extensibility, especially for others wishing to build functionality based on **quanteda.core** in their own packages but who do not need all of the stat or plot functions

kbenoit:
Interim/experimental/transitional solution for v2:

- move the `textmodel_*()` functions to a **quanteda.textmodel** package, and move the draft **quanteda.classifiers** package functions into **quanteda.textmodel**.
- quanteda would import **quanteda.textmodel**, and re-export (`stopwords::stopwords`-style or `magrittr::`%?%`-style) all of the `textmodel_*()` functions from **quanteda.textmodel**.
[...]

PR #1724, opened 2019-06-30
Move all the header files to allow access from other packages

koheiw:
In preparation for #1663


koheiw:
If header files are in "include", other packages can use them by `LinkingTo: quanteda`

PR #1729, opened 2019-07-05
Return valid kwic object even when there is no match

koheiw:
`kwic()` returns data.frame without columns, which causes error in the xlay plot. I cannot build the pkgdown website because of this bug,


koheiw:
Because of this bug, I cannot build the pkgdown website...

Issue #1734, opened 2019-07-15
Assining docvars to corpus is extremely slow


koheiw:
For no reason, whole data.frame is converted to a matrix...

https://github.com/quanteda/quanteda/blob/657115bb78422e67a5e3b60b05ddd498793d7cb6/R/docvars.R#L113

PR #1735, opened 2019-07-15
Fix docvar assignment inconsistency

koheiw:
I do not want to spend time to fix to relatively new bug #1734, so just disallowed assigning a vector (we do not need this functionality anyway). `docvars` is a mess. We should switch to v2.0 soon.

### DIFF
--- a/R/docvars.R
+++ b/R/docvars.R
@@ -99,70 +99,19 @@ docvars.kwic <- function(x) {
 
 #' @export
 "docvars<-.corpus" <- function(x, field = NULL, value) {
-    if (is.dfm(value))
-        value <- convert(value, to = "data.frame")[, -1, drop = FALSE]
-    if ("texts" %in% field)
-        stop("You should use texts() instead to replace the corpus texts.")
-    if (is.null(field)) {
-        if (is.null(value)) {
-            newdv <- data.frame(row.names = docnames(x))
-        } else {
-            newdv <- data.frame(value, stringsAsFactors = FALSE, check.names = FALSE)
-            # uses the V1, V2 etc scheme
-            #names(newdv) <- names(as.data.frame(as.matrix(value)))
-        }
-    } else {
-        newdv <- docvars(x)
-        newdv[field] <- value
-    }
-    meta_names_not_in_newdv <- setdiff(grep("^_.+", names(documents(x)), value = TRUE), names(newdv))
-    documents(x) <- cbind(documents(x)[, c("texts", meta_names_not_in_newdv), drop = FALSE], newdv)
+    x$documents <- set_docvars(x$documents, field, value)
     return(x)
 }
 
 #' @export
 "docvars<-.tokens" <- function(x, field = NULL, value) {
-    if (is.dfm(value))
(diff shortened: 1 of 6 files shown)


koheiw:
I probably stop maintaining v1.5 now.

kbenoit:
I don't understand what was the bug in #1734. Was it just that it was slow? That's easily fixed without changing the default naming behaviour.

The conversion was originally a lazy way to get the default behaviour of adding V1, V2, etc to the columns of a data.frame with no names. The fix proposed here changes the behaviour in that it now assigns integers as names. What it does not do is to disallow assigning a vector to a docvar.

```r
> corp <- corpus(c("A b c d.", "A b c."))
> docvars(corp) <- c("dv1.1", "dv1.2")
> docvars(corp)
      value
text1 dv1.1
text2 dv1.2
> 
> docvars(corp) <- 1
> docvars(corp)
      value
text1     1
text2     1
```

Since it was laziness (mine!) that led to the code referenced in #1734 just to get the data.frame naming defaults, I would be happy to fix it now to provide the same behaviour without the matrix conversion, but I want to make sure first that I understand what you meant by "bug".

koheiw:
The line you added converts entire data.frame `value` to a matrix, which takes long time and changes types of vectors. 
```
names(as.data.frame(as.matrix(value))) 
```

kbenoit:
@koheiw see f8a0a12, I restored the earlier behaviour but without the unnecessary conversion to `matrix`. It's a bit messy in there but all working as expected.

R/docvars.R
```diff
@@ -239,19 +239,31 @@ set_docvars <- function(dvars, field = NULL, value)  {
...
-            } else {
+            if (is.null(value)) {
                 result <- dvars_system
+            } else {
+                if (is.matrix(value))
+                    value <- as.data.frame(value, stringsAsFactors = FALSE)
+                if (is.data.frame(value)) {
+                    if (nrow(value) != nrow(dvars))
+                        stop(message_error("docvar_mismatch"), call. = FALSE)
+                    if (any(missingnames <- nzchar(names(value), keepNA = TRUE))) {
```
koheiw: I prefer to raise error for data.frame without names. Why do we need to auto-generate names?

koheiw: I would be better to drop all the unnamed or duplicated columns than to automatically renaming.

kbenoit: I'm ok with raising an error, but 
1.  This is the current behaviour, and
2. It's more robust for users, especially those who assign a matrix as value. (Note that the coercion from matrix format still takes place, but only if `value` is input as a matrix.)

R/docvars.R
```diff
@@ -239,19 +239,31 @@ set_docvars <- function(dvars, field = NULL, value)  {
...
+                        names(value)[which(is.na(missingnames))] <- 
+                            paste0("V", seq_len(sum(is.na(missingnames))))
+                    }
+                    result <- cbind(dvars_system, value)
```
koheiw: If we add check for column names in line 258, assignment of matrix without names will be also blocked.

PR #1739, opened 2019-07-25
Implement ... passing in ntokens/ntype.tokens

### LINKED ISSUE #1706: ntoken.tokens() is not passing through the `...` argument
The `ntoken.tokens()` is not passing through the `...` arguments in the way that `ntoken.corpus()` is. We should fix this to make it consistent. Same for `ntype.tokens()`.

``` r
library("quanteda")
#> Package version: 1.4.3

# ntoken.corpus()
min(ntoken(data_corpus_inaugural))
#> [1] 147
min(ntoken(data_corpus_inaugural, remove_punct = TRUE, remove_symbols = TRUE))
#> [1] 135

# ntoken.tokens(), no arguments
min(ntoken(toks))
#> [1] 147
min(lengths(tokens(toks)))
#> [1] 147

# ntokens.tokens(), arguments
min(ntoken(toks, remove_punct = TRUE, remove_symbols = TRUE))
#> [1] 147
min(lengths(tokens(toks, remove_punct = TRUE, remove_symbols = TRUE)))
#> [1] 135

# same for ntype.tokens()
min(ntype(toks, remove_punct = TRUE, remove_symbols = TRUE))
#> [1] 96
min(lengths(lapply(tokens(toks, remove_punct = TRUE, remove_symbols = TRUE), unique)))
#> [1] 91
```

<sup>Created on 2019-06-09 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

kbenoit:
Fixes #1706

### DIFF
--- a/R/nfunctions.R
+++ b/R/nfunctions.R
@@ -54,7 +54,7 @@ ndoc.tokens <- function(x) {
 #' # number of features
 #' nfeat(dfm(corpus_subset(data_corpus_inaugural, Year > 1980), remove_punct = FALSE))
 #' nfeat(dfm(corpus_subset(data_corpus_inaugural, Year > 1980), remove_punct = TRUE))
-nfeat <- function(x) {  # TODO: nfeature has to deplicate one day
+nfeat <- function(x) {
     UseMethod("nfeat")
 }
 
@@ -68,14 +68,6 @@ nfeat.dfm <- function(x) {
     x <- as.dfm(x)
     ncol(x)
 }
-
-# nfeat.tokens <- function(x) {
-#     if (attr(x, 'padding')) {
-#         length(types(x)) + 1
-#     } else {
-#         length(types(x))
-#     }
-# }
 
 #' Defunct form of nfeat
 #' 
@@ -184,7 +176,7 @@ ntype.corpus <- function(x, ...) {
 ntype.dfm <- function(x, ...) {
     check_dots(list(...))
     x <- as.dfm(x)
-    ## only returns total non-zero features
+    # only returns total non-zero features
     result <- as.integer(rowSums(x > 0))
     names(result) <- docnames(x)
     result
@@ -225,7 +217,7 @@ nsentence.default <- function(x, ...) {
 
 #' @export
 nsentence.character <- function(x, ...) {
-    upcase <- 
+    upcase <-
         try(any(stringi::stri_detect_charclass(x, "[A-Z]")), silent = TRUE)
     if (!is.logical(upcase)) {
         # warning("Input text contains non-UTF-8 characters.")
--- a/R/nfunctions.R
+++ b/R/nfunctions.R
@@ -152,7 +152,10 @@ ntoken.character <- function(x, ...) {
 
 #' @export
 ntoken.tokens <- function(x, ...) {
-    lengths(x)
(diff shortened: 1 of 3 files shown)


R/nfunctions.R
```diff
@@ -176,15 +174,17 @@ ntype.corpus <- function(x, ...) {
...
+    # only returns total non-zero features
     result <- as.integer(rowSums(x > 0))
     names(result) <- docnames(x)
     result
 }
 
 #' @export
 ntype.tokens <- function(x, ...) {
+    if (length(list(...))) x <- tokens(x, ...)
     vapply(unclass(x), function(y) length(unique(y[y > 0])), integer(1))
```
koheiw: I think you should use this code for ntype.tokens()
```r
# nfeat.tokens <- function(x) {
#     if (attr(x, 'padding')) {
#         length(types(x)) + 1
#     } else {
#         length(types(x))
#     }
# }
```

kbenoit: That would be fine except `ntype()` is vectorized over documents. The behaviour should be:
``` r
library("quanteda")
txt <- c(
  d1 = "a b c a b c",
  d2 = "a b c d e"
)
toks <- tokens(txt)

ntype(toks)
## d1 d2 
##  3  5
```
Also I don't think we should count the pad as a "type". So this is correct:
```r
ntype(tokens_remove(toks, "c", padding = TRUE))
## d1 d2 
##  2  4
```

koheiw: What function should people use to get the size of vocabulary in the entire corpus (tokens object)? Like `length(types(toks))`?

kbenoit: That would be my recommendation.

So this thread is resolved and we can merge the PR?

koheiw: If unique types in tokens is equivalent to features in dfm, then `ntype() == nfeat()`. This is why I asked the questions.

kbenoit: For clarity I’ll elaborate stuff we already know, but it answers the nfeat question. Here goes:

By our terminology and our implementation, tokens and types mean all words and unique words respectively, but these are attributes of documents. So, `ntoken()` and `ntype()` always return a named vector of the total tokens and total types, with each element corresponding to a document.

`tokens` objects have no “features”, only a dfm has features. That’s why we have not implemented `nfeat.tokens`. `nfeat()` only works on a dfm, and returns the number of columns. (Just as `ndoc.dfm()` returns the number of rows/documents.)

If someone wanted the unique types across a tokenised corpus, they could either use what you suggest in https://github.com/quanteda/quanteda/pull/1739#discussion_r307352760 or we could think of a new function as a shortcut. The existing nfunctions however are entirely consistent. Implementing a new one is not part of this PR.

koheiw: How about `ntoken(x, unique =TRUE)`?

kbenoit: I think you mean `ntype(x, unique = TRUE)` but that's a not part of this issue and therefore not part of this PR, so we should be discussing and implementing it separately. 

Not good practice to hold up a specific fix because we see possible features to add when digging through the code for the specific fix.

koheiw: Ah, I though you are adding `ntype` in this PR that I even did not know existed.

PR #1741, opened 2019-07-25
Require y to be non-constant in textmodel_nb()

### LINKED ISSUE #1711: prediction on a constant crashes predict.textmodel_nb()
``` r
library(quanteda)
## Registered S3 methods overwritten by 'ggplot2':
##   method         from 
##   [.quosures     rlang
##   c.quosures     rlang
##   print.quosures rlang
## Package version: 1.4.5
## 
## 
##     View
irb_dfm <- dfm(quanteda::data_corpus_irishbudget2010)
model <- textmodel_nb(irb_dfm, y = docvars(irb_dfm)[["debate"]], distribution = "multinomial")
predict(model, irb_dfm)
## Error in names(nb.predicted) <- docnames(data): 'names' attribute [14] must be the same length as the vector [1]
```
Although it makes no sense to predict on a constant, we should at least return the prediction equal to the constant value and not have the function crash.

kbenoit:
Fixes #1711.

### DIFF
--- a/R/textmodel_nb.R
+++ b/R/textmodel_nb.R
@@ -114,7 +114,7 @@ textmodel_nb.dfm <- function(x, y, smooth = 1,
     call <- match.call()
     
     y <- as.factor(y)
-    if (var(as.numeric(y), na.rm = TRUE) == 0) stop("y cannot be constant")
+    if (stats::var(as.numeric(y), na.rm = TRUE) == 0) stop("y cannot be constant")
     
     temp <- x[!is.na(y),]
     class <- y[!is.na(y)]
--- a/R/textmodel_nb.R
+++ b/R/textmodel_nb.R
@@ -115,7 +115,9 @@ textmodel_nb.dfm <- function(x, y, smooth = 1,
     distribution <- match.arg(distribution)
     call <- match.call()
     
-    y <- factor(y)
+    y <- as.factor(y)
+    if (length(unique(na.omit(y))) == 1) stop("y cannot be constant")
+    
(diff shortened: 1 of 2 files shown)


R/textmodel_nb.R
```diff
@@ -115,7 +115,9 @@ textmodel_nb.dfm <- function(x, y, smooth = 1,
     distribution <- match.arg(distribution)
     call <- match.call()
     
-    y <- factor(y)
+    y <- as.factor(y)
+    if (length(unique(na.omit(y))) == 1) stop("y cannot be constant")
```
koheiw: How about `var(as.numeric(y), na.rm = TRUE)) == 0`?

kbenoit: OK, that's most efficient (although the differences are tiny). As long as we don't get into any numerical precision issues, then we should be ok.
```r
> x <- as.factor(sample(letters[1:3], size = 1000000, replace = TRUE))
> microbenchmark(unique = (length(unique(na.omit(x))) == 1), 
+                var = var(as.numeric(x), na.rm = TRUE) == 0,
+                dupl = all(duplicated(x)[-1L]))
Unit: milliseconds
   expr       min        lq     mean    median       uq       max neval
 unique 15.457192 17.864760 22.85690 19.102567 22.57854  66.27518   100
    var  6.526918  7.920738 14.83654  9.182764 15.35607 221.80587   100
   dupl 13.403758 16.811788 23.87359 18.691379 25.27883 227.15093   100
```

PR #1745, opened 2019-08-01
Add Yule's I to textstat_lexdiv()

chainsawriot:
Yule's I is another textual diversity measure. Although it is similar to Yule'K, it is not the same. It has recently been used in the article by Tolochko & Boomgaarden (2019).

Tolochko, P., & Boomgaarden, H. G. (2019). Determining Political Text Complexity: Conceptualizations, Measurements, and Application. International Journal of Communication, 13, 21.


kbenoit:
Thanks @chainsawriot !

Issue #1743, opened 2019-07-30
Non-exclusive tokens lookup is wrong when there are paddings


koheiw:
```r
> toks <- tokens("a x c y d") %>% 
+   tokens_remove(c("x", "y"), padding = TRUE)
> toks
tokens from 1 document.
text1 :
[1] "a" ""  "c" ""  "d"

> tokens_lookup(toks, dictionary(list("aa" = "a", "bb" = "b")))
tokens from 1 document.
text1 :
[1] "aa"

> tokens_lookup(toks, dictionary(list("aa" = "a", "bb" = "b")), exclusive = FALSE)
tokens from 1 document.
text1 :
[1] "AA" "AA" "c"  "AA" "d" 
```

PR #1773, opened 2019-11-23
Fix ID shifting of paddings in document without keyword values

koheiw:
Further  fix for #1743. Let's marge this before #1772 so that I can marge this to v2.


koheiw:
Merged to dev-v2

Issue #1623, opened 2019-02-21
More tests of robustness


kbenoit:
Especially for:

- [x] `is.corpus()`
- [x] `is.tokens()`
- [x] `is.dfm()`
- [x] `as.corpus()`
- [x] `as.tokens()`
- [x] `as.dfm()`

and should be done with both the new and older versions of all objects. Older versions are in `tests/data/pre15objects`.

PR #1793, opened 2019-12-04
Test object robustness

kbenoit:
I'm creating this as an early pull request that will require more development to pass.

It checks the `is.*()` and `as.*()` methods, per #1623, for the core object classes. It's a good test or robustness since 
- the `as.*()` methods should input an old or a new object and always output a new object. This requires that all of the required attributes or slots be present.
- the `is.*()` methods should work accurately for both pre-v2 and >= v2 objects.

At the time of the creation of the PR, not all of the tests are working, and will require some code changes in order to make the breaking tests pass. For instance the `meta` attribute does not appear to be created with `as.tokens.tokens()`.


kbenoit:
@koheiw This is related to attribute copying in https://github.com/quanteda/quanteda/pull/1795, and the tests currently breaking (see the comments in the new test code) will pass if the attribute issues are fixed. This relates to https://github.com/quanteda/quanteda/pull/1795 - can you fix it so that the tests pass? Thanks!

koheiw:
I added upgrading functions, but not sure what to do with this warning.
```r
> load("tests/data/pre_v2_objects/data_dfm_pre2.rda")
Warning message:
Not a validObject(): no slot of name "unit" for this object of class "dfm" 
```
I added validator but is is not triggered when the object is loaded. Any idea @kbenoit ?
```r
setValidity("dfm", function(object) {
    if (is_pre2(object))
        warning("Upgrade the object for v2.0 using as.dfm()")
})
```

kbenoit:
The warning is raised because we added the “unit” slot recently to the dfm class, but this is not a slot present in the pre-v2 dfm objects. So when it’s read without being passed first to as.dfm() for an upgrade, we get the S4 warning. We should figure out a way to fix this if possible.

koheiw:
Shouldn't R call `validObject()` and then the custom validator?

kbenoit:
I think it’s only called when an object is created or copied.

PR #1795, opened 2019-12-11
Issue 1612

koheiw:
This PR is too improve consistency of handling attributes. Only behavior change is `cbind()` allows duplicated features.


tests/testthat/test-docnames.R
```diff
@@ -26,14 +26,14 @@ test_that("docnames<- works with corpus, tokens and dfm (#987)", {
...
-    dfmat <- as.dfm(data_dfm_lbgexample)
-    dfmat@Dimnames$docs <- seq_len(ndoc(dfmat))
-    expect_identical(
-        dimnames(as.matrix(textstat_simil(dfmat)))[[1]],
-        as.character(1:6)
-    )
-})
+# test_that("dfm docnames cannot be made integer before textstat_simil", {
+#     dfmat <- as.dfm(data_dfm_lbgexample)
+#     dfmat@Dimnames$docs <- seq_len(ndoc(dfmat))
```
kbenoit: No one should be doing this, but the check for dimnames being character only happen when something tries to modify the object, as apparently `textstat_simil()` does. This causes a run-time error even though the assignment of the integer dimnames itself violates the class/slot type declaration. A flaw of S4, true, but is it something we could or should trap?

kbenoit: Same error with `docnames(dfmat) <- 1:6`

koheiw: We could run `set_dfm_dimnames(x) <- dimnames(x)` in `as.dfm()` to covert number to characters but I am not sure if we need to save people who try to destroy their own objects...

koheiw: Good idea to coerce input to character in `docnames()`.

koheiw:
Travis is falling to install libraries.

kbenoit:
Looks good, I will approve if/when Travis finishes the checks. :+1:

kbenoit:
fcm indexing appears to be failing:

```r
x <- fcm(tokens(c("this contains lots of stopwords",
                  "no if, and, or but about it: lots"),
                remove_punct = TRUE))
x[]
##  Error in x[] : Subscript out of bounds 
```

Issue #1791, opened 2019-12-03
dfm/fcm indexing with single index argument behaves differently from other matrix indexing


kbenoit:
These behaviours are inconsistent:

``` r
library("quanteda")
## Package version: 1.9.9000
## Parallel computing: 2 of 12 threads used.
## See https://quanteda.io for tutorials and examples.
## 
## Attaching package: 'quanteda'
## The following object is masked from 'package:utils':
## 
##     View

data_dfm_lbgexample["R1"]
## Document-feature matrix of: 1 document, 37 features (54.1% sparse).
as.matrix(data_dfm_lbgexample)["R1"]
## [1] NA
as(data_dfm_lbgexample, "dgTMatrix")["R1"]
## Error in if (any(as.logical(i))) {: missing value where TRUE/FALSE needed

data_fcm_lbgexample <- fcm(data_dfm_lbgexample)
data_fcm_lbgexample["A"]
## Feature co-occurrence matrix of: 1 by 37 features.
as.matrix(data_fcm_lbgexample)["A"]
## [1] NA
as(data_fcm_lbgexample, "dgTMatrix")["R1"]
## Error in if (any(as.logical(i))) {: missing value where TRUE/FALSE needed
```

koheiw:
This should be easy to solve if we managed to find distinguish between `["A",]` and `["A"]` in S4 methods. I have no clue so far.

koheiw:
https://github.com/cran/Matrix/blob/c8ca9f0a455a420142ddba4e4b4df22c03b45238/R/Csparse.R#L185-L196

PR #1796, opened 2019-12-12
Raise error when only single index is provided for #1791

koheiw:


koheiw:
`matrix` behaves like a numeric vector with single indices. Let's stick with `Matrix`
```r
> mat <- matrix(1:9, nrow = 3)
> mat[5]
[1] 5
> mat["aa"]
[1] NA
```

Issue #1797, opened 2019-12-12
Upgrade `tokens_compound()`


koheiw:
In relation to #1503, we can add two arguments to `tokens_compound()`.

-  `non_sequence`: compound any adjacent tokens in `pattern`.
-  `window`: compound neighbors of tokens in `pattern`

Consider tokens 
```r
toks <- as.tokens(list("a", "b", "c", "d", "e"))
```
Then,
```r
tokens_compound(toks, pattern = c("a", "d", "b"), non_sequence = TRUE)
```
will produce  `c("a_b", "c", "d", "e")`. This can be used to generate ngrams without removing stopwords before hand, for example.

```r
tokens_compound(toks, pattern =  c("a", "d", "b"), window = 1)
``` 
will produce  `c("a_b_c", "d_e")`. This can be used to tokenize negations better.

koheiw:
It is easy to implement these but the remaining question is what do we do with paddings... `tokens_select()` includes paddings in its window, but we don't want to ignore paddings here.

kbenoit:
OK, I've reviewed PR #1798, will put the comments here.

First, what exactly is the use case here?

Second, below, is that what you intended for the pads? Seems to make sense to me but it does "remove" the pad through absorption.

``` r
library("quanteda")
## Package version: 1.9.9000
## Parallel computing: 2 of 12 threads used.
## See https://quanteda.io for tutorials and examples.
## 
## Attaching package: 'quanteda'
## The following object is masked from 'package:utils':
## 
##     View

toks <- as.tokens(list(d1 = letters[1:10]))

tokens_compound(toks, phrase("d e"))
## tokens from 1 document.
## d1 :
## [1] "a"   "b"   "c"   "d_e" "f"   "g"   "h"   "i"   "j"
tokens_compound(toks, phrase("d e"), window = 1)
## tokens from 1 document.
## d1 :
## [1] "a"       "b"       "c_d_e_f" "g"       "h"       "i"       "j"
tokens_compound(toks, phrase("d"), window = 1)
## tokens from 1 document.
## d1 :
## [1] "a"     "b"     "c_d_e" "f"     "g"     "h"     "i"     "j"

# should the pad be removed (by being absorbed into right of "d")
tokens_remove(toks, "e", padding = TRUE) %>%
  tokens_compound("d", window = 1)
## tokens from 1 document.
## d1 :
## [1] "a"   "b"   "c_d" "f"   "g"   "h"   "i"   "j"

[...]

PR #1798, opened 2019-12-12
Dev compound

koheiw:
Add `window` to compound. See #1797.

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -53,8 +53,8 @@ qatd_cpp_tokens_chunk <- function(texts_, types_, size, overlap) {
     .Call(`_quanteda_qatd_cpp_tokens_chunk`, texts_, types_, size, overlap)
 }
 
-qatd_cpp_tokens_compound <- function(texts_, compounds_, types_, delim_, join) {
-    .Call(`_quanteda_qatd_cpp_tokens_compound`, texts_, compounds_, types_, delim_, join)
+qatd_cpp_tokens_compound <- function(texts_, compounds_, types_, delim_, join, window_left, window_right) {
+    .Call(`_quanteda_qatd_cpp_tokens_compound`, texts_, compounds_, types_, delim_, join, window_left, window_right)
 }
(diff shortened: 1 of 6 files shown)


kbenoit:
See #1797. If we adopt this, I think we would need
- a more clear description of the `window` argument in the help
- an example showing a simple use case. Why would someone need or want this?

koheiw:
This is a way to avoid patterns like `phrase("* * *")` which is near impossible because the exponential increase in number of fixed combinations. Main usage will be compounding hyphens `c("x", "-", "y")` without `*` and negations `c("not", "x")` or `c("not", "y")`. I don't think many people use this but does not harm anyone either.

koheiw:
I have it smarter to avoid removing padding. The main purpose of this function is in https://github.com/quanteda/quanteda/blob/f2fa75354d3222ab1a8c10a01e0fa6ff667b8d68/R/tokens.R#L515-L516

koheiw:
``` r
require(quanteda)
#> Package version: 1.9.9000
#> Parallel computing: 2 of 4 threads used.
#> 
#> 
#>     View
quanteda_options(threads = 4)

identical(
  as.list(tokens_compound(toks, phrase("not *"))),
  as.list(tokens_compound(toks, "not", window = c(0, 1)))
)
#> [1] TRUE

  tokens_compound(toks, phrase("not *")),
  tokens_compound(toks, "not", window = c(0, 1)),
  times = 10
)
#> Unit: milliseconds
#>                                            expr       min        lq      mean
#>          tokens_compound(toks, phrase("not *")) 305.83083 311.09162 346.93148
#>  tokens_compound(toks, "not", window = c(0, 1))  76.57474  78.73434  90.42414
#>     median      uq      max neval
#>  344.98429 365.615 425.5151    10
#>   90.04657 100.535 105.1423    10
```

<sup>Created on 2019-12-14 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Issue #1503, opened 2018-11-24
Consider switching to post-processing of special tokens


koheiw:
Post-processing is as fast as current pre-processing:
```r
ndoc(corp)
##  [1] 84599
system.time({
toks <- stri_split_boundaries(texts(corp), 
                              type = "word", skip_word_none = FALSE) %>% 
        as.tokens() %>% 
        tokens_compound(phrase(c("# *", "@ #", "* -", "- *")), concatenator = "") %>% 
        tokens_remove("^[\\p{Z}\\p{C}]+$", valuetype = "regex")
})
##   user  system elapsed 
## 127.989  11.377  85.952 
system.time({
toks <- tokens(corp)
})
##    user  system elapsed 
##  98.810   2.519  77.285
```

kbenoit:
That's a brilliant idea, and would finally give us the option to separate the core token segmenter function from our own preferred handling of segmented tokens. This would make it possible (finally!) to address #276. Other "tokenizers" are faster than `tokens(x, what = "word")` but only because of our special handling of things like Twitter punctuation characters.

koheiw:
If we are going to do minimal pre-processing and provide a handful of post-processing functions (`tokens.tokens()`, `tokens_split()`, `tokens_compound()`, `tokens_select()` with a new position argument), users can do tag extraction on tokens, and we can kill `corpus_segment()`, which is one of the ugliest functions we have.

koheiw:
If we really want to do better to add `window` to `tokens_compound()` because
```r
tokens_compound(list(c("#", "*"), c("@", "*")), concatenator = "") # twitter
tokens_compound(c("*", "-"), c("-", "*")), concatenator = "") # hyphen
```
is the same as 
```r
tokens_compound(c("#", "@"), window = c(0, 1), concatenator = "") # twitter
tokens_compound("-", window = c(1, 1), concatenator = "") # hyphen
```
`windows` is better because we can avoid the used of "*".

koheiw:
Post-processing of twitter tags is 3 times ( `(129 - 47) / (75 - 47)`) faster than current pre-processing.
```r
require(quanteda)
require(stringi)
require(lubridate)
quanteda_options(threads = 8)

corp <- readRDS("~/Documents/Sputnik/Data/data_corpus_tweets.RDS")

split <- function(x) {
  stri_split_boundaries(x, type = "word", skip_word_none = FALSE) %>% 
    as.tokens()
}

post <- function(x) {
[...]

Issue #1446, opened 2018-10-07
tokens(x) stumbles on interior punctuation


kbenoit:
Punctuation in the middle of a word should be preserved, but it is not.
```r
txt <- "holy s%$t, quanteda is a f***ing great"

tokens(txt, remove_punct = FALSE)
## tokens from 1 document.
## text1 :
##  [1] "holy"     "s"        "%"        "$"        "t"        ","        "quanteda"
##  [8] "is"       "a"        "f"        "*"        "*"        "*"        "ing"     
## [15] "great"   
tokens(txt, remove_punct = TRUE)
## tokens from 1 document.
## text1 :
## [1] "holy"     "s"        "t"        "quanteda" "is"       "a"        "f"       
## [8] "ing"      "great" 
```

Changing the boundary method helps but still cannot distinguish the end of word punctuation:
```r
txt <- "holy s%$t, quanteda, is a f***ing great"
tokens(txt, what = "fasterword", remove_punct = TRUE)
## tokens from 1 document.
## text1 :
## [1] "holy"      "s%$t,"     "quanteda," "is"        "a"         "f***ing"  
## [7] "great"   
```
True  for Twitter characters too.  The first is correct, but when  we try to remove Twitter characters, the problem arises:
```r
txt <- "I'm @kenbenoit on Twitter, #rstats don't u B a h@ter!"
tokens(txt, remove_twitter = FALSE, remove_punct = TRUE)
## tokens from 1 document.
## text1 :
## [1] "I'm"        "@kenbenoit" "on"         "Twitter"    "#rstats"    "don't"     
## [7] "u"          "B"          "a"          "h@ter"

tokens(txt, remove_twitter = TRUE, remove_punct = TRUE)
tokens from 1 document.
text1 :
## [1] "I'm"       "kenbenoit" "on"        "Twitter"   "rstats"    "don't"     "u"        
## [8] "B"         "a"         "h"         "ter"
```

koheiw:
We could use stringi's RBBI to preserve internal punctuations as mentioned in #896

kbenoit:
Excellent idea, RBBI is in the current CRAN version of **stringi** now (v1.2.4).

Issue #1477, opened 2018-10-29
Inconsistency between remove_url and remove_twitter


koheiw:
In relation to #1465, I think that there is an inconsistency between `remove_url` and `remove_twitter`. Why whole "@koheiw" and "#quanteda" are not remove, while "http://google.com" is deleted?
```r
> tokens("http://google.com aaa", remove_url = TRUE)
tokens from 1 document.
text1 :
[1] "aaa"

> tokens("@koheiw #quanteda bbb", remove_punct = TRUE, remove_twitter = TRUE)
tokens from 1 document.
text1 :
[1] "koheiw"   "quanteda" "bbb"
```
I expect the second to be
```r
[1] "bbb"
```

koheiw:
We could add `remove_tag` to delete whole tags, and deprecate `remove_twitter`.

amatsuo:
In my understanding, the use of `remove_twitter` is for analyzing hashtags (and sometimes people's names) along with normal nouns. For example, we can combine `#boris` and `boris` into one. For that purpose, the current implementation is useful. 

I don't see much reason to have a functionality of removing hashtags and mentions entirely, as these are quite useful for many purposes and removing them with `tokens_remove` is super easy.

kbenoit:
The name is misleading but the function is sound: it means preserve the special Twitter punctuation characters when `remove_punct = TRUE, remove_twitter = FALSE`. @amatsuo and I relied on this a lot in the Brexit study to keep #brexit and @strongerin (e.g.) as hashtags and usernames while removing other punctuation.

We used the `remove_twitter` to keep the name consistent, so that everything could start with `remove_...` and also not to have a longer compound argument name such as `remove_twitter_punct`.

In your example to keep only "bbb" then it's easiest to do it separately:

```r
> tokens("@koheiw #quanteda bbb") %>%
      tokens_remove(pattern = "^[#@].+$", valuetype = "regex")
tokens from 1 document.
text1 :
[1] "bbb"

```

koheiw:
I am doing `tokens_remove(pattern = "^[#@].+$", valuetype = "regex")` currently. 

In order to make `what = "word"` and `what = "fasterword"` consistent (#1465) without loosing flexibility, we have to distinguish between keeping "@" and "#" with following word ("kohei" or "quanteda") and (2) removing "@" and "#". `remove_twitter` actually does only 1 and `remove_punct` does 2. We can call it  `separate_twitter_punct` (and make it independent of `remove_punct`).

[...]

PR #1801, opened 2019-12-13
Change to post processing

koheiw:
- remove preserve_special
- add special handling for hyphens and twitter using `tokens_compound()`

This is mainly for #1503, but can also address #1446


koheiw:
It is failing tests due to #1477. `remove_hyphens` needs to be `split_hyphens` in order to gain the same result from different tokens functions, and tokenization mode (word/fasterword/fastetword).

R/tokens.R
```diff
@@ -716,7 +721,7 @@ serialize_tokens <- function(x, types_reserved = NULL, ...) {
 #' # stemming
 #' toks2 <- tokens("Stemming stemmed many word stems.")
 #' unclass(toks2)
-#' unclass(quanteda:::tokens_recompile(tokens_wordstem(toks2)))
+#' unclass(quanteda:::tokens_recompile(stem(toks2)))
```
kbenoit: Must have been a mistake - I will change it back.

kbenoit:
It's failing a number of tests now because of changes in how hyphens are processed. In former behaviour, `remove_hyphens = TRUE, remove_punct = FALSE ` both split a hyphenated word _and_ kept the hyphen. In the PR, it removes the inner hyphen.

So this passed before the PR, but fails in the PR (from `test-tokens.R:381`):
```r
    expect_equal(as.character(tokens(txt, remove_hyphens = TRUE, remove_punct = FALSE)[[1]]),
                 c("a", "b", "-", "c", "d", ".", "!"))
``` 

For consistency, I think we should keep the older behaviour, and consider making the change later as a policy decision once we work out the full consistency of all of the token processing rules, across different tokenization "engines".

koheiw:
My plan is to call the same internal function from `tokens.tokens()` and `tokens.corpus()` to make their behavior strictly the same. Let's decided what is the desired behavior, or leave this branch until we become clear about this. I think v2.0 is the best timing to switch to new behavior, although it would not be to different.

My suggestion is adding
-  `separate_hyphens`
- `separate_twitter`
and useing `remove_punct` for removing separated hyphens or Twitter tags.

Currenct `remove_hyphens = TRUE` will be the same as `separate_hyphens = TRUE` and `remove_punct = TRUE)`. The difference is that it becomes possible keeping hyphens as separate token "-" when `separate_hyphens = FALSE` and `remove_punct = FALSE`. The same for `remove_twitter`.

koheiw:
The new argument can be `compound_hyphens` or `join_hyphens` to make it clear that users can do it manually using `tokens_compound()`

kbenoit:
Not foolproof though...
```
> tokens("Pre- and post-processing") %>%
+     tokens_compound("-", window = 1)
tokens from 1 document.
text1 :
[1] "Pre_-_and"        "post-processing"
```

koheiw:
Good point :thinking:

kbenoit:
Other packages do this differently, although I prefer to match our previous behaviour than adopt this rule (which I don't think makes sense):
```r
> tm::removePunctuation("I can't zig-zag.")
[1] "I cant zigzag"
> tm::removePunctuation("I can't zig-zag.", preserve_intra_word_contractions = TRUE)
[1] "I can't zigzag"
> tm::removePunctuation("I can't zig-zag.", preserve_intra_word_dashes = TRUE)
[1] "I cant zig-zag"
```

Overall, I think this would be a good case for 
- adding a new argument called `separate_hyphens = FALSE` which turns "zig-zag" into "zig", "-", "zag", just as does `remove_hyphens = TRUE` does now;
- deprecate `remove_hyphens`
- not automatically splitting on the first tokenization step, since it's probably faster to split post-tokenization, rather than to concatenate post-tokenization, _and_ it avoids the problem [above](https://github.com/quanteda/quanteda/pull/1801#issuecomment-565841054).

Can we implement the new processing workflow for v2 and still match this behaviour?
``` r
library("quanteda")
## Package version: 1.5.2

txt <- "Ex-post, @quantedainit #rstats."

tokens(txt, remove_punct = TRUE, remove_twitter = TRUE, remove_hyphens = TRUE)
## tokens from 1 document.
## text1 :
## [1] "Ex"           "post"         "quantedainit" "rstats"
tokens(txt, remove_punct = TRUE, remove_twitter = TRUE, remove_hyphens = FALSE)
## tokens from 1 document.
## text1 :
## [1] "Ex-post"      "quantedainit" "rstats"
tokens(txt, remove_punct = TRUE, remove_twitter = FALSE, remove_hyphens = TRUE)
## tokens from 1 document.
## text1 :
## [1] "Ex"            "post"          "@quantedainit" "#rstats"
tokens(txt, remove_punct = TRUE, remove_twitter = FALSE, remove_hyphens = FALSE)
## tokens from 1 document.
## text1 :
## [1] "Ex-post"       "@quantedainit" "#rstats"
tokens(txt, remove_punct = FALSE, remove_twitter = TRUE, remove_hyphens = TRUE)
## Warning in tokens_internal(texts(x), ...): remove_twitter reset to FALSE when
## remove_punct = FALSE
## tokens from 1 document.
## text1 :
## [1] "Ex"            "-"             "post"          ","            
## [5] "@quantedainit" "#rstats"       "."
tokens(txt, remove_punct = FALSE, remove_twitter = TRUE, remove_hyphens = FALSE)
## Warning in tokens_internal(texts(x), ...): remove_twitter reset to FALSE when
## remove_punct = FALSE
## tokens from 1 document.
## text1 :
## [1] "Ex-post"       ","             "@quantedainit" "#rstats"      
## [5] "."
tokens(txt, remove_punct = FALSE, remove_twitter = FALSE, remove_hyphens = TRUE)
## tokens from 1 document.
## text1 :
## [1] "Ex"            "-"             "post"          ","            
## [5] "@quantedainit" "#rstats"       "."
tokens(txt, remove_punct = FALSE, remove_twitter = FALSE, remove_hyphens = FALSE)
## tokens from 1 document.
## text1 :
## [1] "Ex-post"       ","             "@quantedainit" "#rstats"      
## [5] "."
```

koheiw:
In `tokens.corpus()`, we are file in this way
```r
> stri_split_boundaries("Pre- and post-processing", type = "word", skip_word_none = FALSE) %>% 
+   as.tokens() %>% 
+   tokens_remove("^[\\p{Z}\\p{C}]+$", valuetype = "regex", padding = TRUE) %>% 
+   tokens_compound("-", window = 1, concatenator = "")
tokens from 1 document.
text1 :
[1] "Pre-"            ""                "and"             ""                "post-processing"
```

but not sure in `tokens.tokens()`...

kbenoit:
Ah right, we actually split then preserve now. So we already get the above case wrong, unless we use the whitespace splitter, in which case the behaviour is inconsistent.
```
> tokens("Pre- and post-processing", remove_hyphens = FALSE)
tokens from 1 document.
text1 :
[1] "Pre"             "-"               "and"             "post-processing"

> tokens("Pre- and post-processing", remove_hyphens = FALSE, what = "fasterword")
tokens from 1 document.
text1 :
[1] "Pre-"            "and"             "post-processing"
```

For total consistency we would not use any of the `stri_split_boundaries()` options at all, but just split whitespace, and then do the operations in post, but this might limit our applicability to non-English or at least non-Western languages.

I'm not sure of the solution, but we should be maximising flexibility to choose different tokenisation engines, while also ensuring consistency of our `remove_*` options in applying to them after splitting.

Worth a discussion probably, or solve it in January.

koheiw:
We can get closer.
``` r
require(stringi)
require(quanteda)
#> Package version: 1.9.9004
#> Parallel computing: 2 of 4 threads used.
#> 
#> 
#>     View

txt <- "Ex-post, @quantedainit #rstats."

tokens2 <- function(x, remove_punct, remove_twitter, remove_hyphens) {
  result <- as.tokens(stri_split_boundaries(x, type = "word", skip_word_none = FALSE))
  result <- tokens_remove(result, "^[\\p{Z}\\p{C}]+$", valuetype = "regex", padding = TRUE)
  if (!remove_twitter)
    result <- tokens_compound(result, "^[#@]$", window = c(0, 1), valuetype = "regex", concatenator = "")
  #if (remove_twitter)
  #  result <- tokens_remove(result, "^[#@]$", window = c(0, 1), valuetype = "regex")
  if (!remove_hyphens)
    result <- tokens_compound(result, "-", window = 1, valuetype = "regex", concatenator = "")
  if (remove_punct)
    result <- tokens_remove(result, "^[\\p{P}]+$", valuetype = "regex")
  return(result)
}
tokens2(txt, remove_punct = TRUE, remove_twitter = TRUE, remove_hyphens = TRUE)
#> tokens from 1 document.
#> text1 :
#> [1] "Ex"           "post"         ""             "quantedainit" ""            
#> [6] "rstats"
tokens2(txt, remove_punct = TRUE, remove_twitter = FALSE, remove_hyphens = TRUE)
#> tokens from 1 document.
#> text1 :
#> [1] "Ex"            "post"          ""              "@quantedainit"
#> [5] ""              "#rstats"
tokens2(txt, remove_punct = TRUE, remove_twitter = FALSE, remove_hyphens = FALSE)
#> tokens from 1 document.
#> text1 :
#> [1] "Ex-post"       ""              "@quantedainit" ""             
#> [5] "#rstats"
tokens2(txt, remove_punct = FALSE, remove_twitter = TRUE, remove_hyphens = TRUE)
#> tokens from 1 document.
#> text1 :
#>  [1] "Ex"           "-"            "post"         ","            ""            
#>  [6] "@"            "quantedainit" ""             "#"            "rstats"      
#> [11] "."
tokens2(txt, remove_punct = FALSE, remove_twitter = FALSE, remove_hyphens = TRUE)
#> tokens from 1 document.
#> text1 :
#> [1] "Ex"            "-"             "post"          ","            
#> [5] ""              "@quantedainit" ""              "#rstats"      
#> [9] "."
tokens2(txt, remove_punct = FALSE, remove_twitter = FALSE, remove_hyphens = FALSE)
#> tokens from 1 document.
#> text1 :
#> [1] "Ex-post"       ","             ""              "@quantedainit"
#> [5] ""              "#rstats"       "."
```

<sup>Created on 2019-12-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #1803, opened 2019-12-13
Fix tokens

koheiw:
This PR improves handling of Unicode control characters that sometimes leads to DFM crash.


koheiw:
Travis is failing here too

Issue #1411, opened 2018-08-04
dfm.tokens() does not have case_insensitive for select


koheiw:
## Describe the bug

`dfm()` has the `select` and `valuetype` arguments but it does not have the `case_insensitive` argument. I there is no reason for not having it. I guess we just forgot to add this.

## Reproducible code

Please paste minimal code that reproduces the bug. If possible, please upload the data file as `.rds`.
```r
dfm(tokens("a A"), select = "a", case_insensitive = FALSE)
# Document-feature matrix of: 1 document, 1 feature (0% sparse).
# 1 x 1 sparse Matrix of class "dfm"
#       features
# docs    a
#   text1 2

# Warning message:
# Argument case_insensitive not used. 
```

PR #1812, opened 2019-12-18
Add case_insensitive to dfm

koheiw:
#1411


R/dfm.R
```diff
@@ -28,6 +28,7 @@
 #'   separately, or using [tokens_lookup()] on the tokenized text
 #'   before calling `dfm`.
 #' @inheritParams valuetype
+#' @param case_insensitive ignore the case of dictionary values if `TRUE`
```
kbenoit: will this affect matches for patterns specified with `select` as well?

koheiw: This is a wrong description that I compiled from other function...

kbenoit: We should probably review this across functions for consistency. I just noticed that `dfm_select()` also refers to dictionaries.

If we add this to the man page for valuetype we could put it all in one place, and @inheritParams when needed as with valuetype. I would describe then as:
```r
@param case_insensitive logical; if `TRUE`, ignore case when matching a `pattern` [dictionary] values
```

PR #1832, opened 2020-01-13
Add summary metadata

kbenoit:
Add functions to compute and store summary information inside the (system) metadata of a corpus. This allows corpora to have the default summary information pre-computed and stored, and then retrieved rather than computed on the fly for `summary.corpus()`.

The summary information is stored in `meta(x, "summary", type = "system")`, but can be:
- set using `add_system_summary(x, ...)`
- retrieved using `get_system_summary(x, ...)`

When `summary.corpus()` is called, this information is returned rather than calling the internal function `summarize_texts()` to compute them. If the names of the documents of `x` are not identical to the summary document names, this is automatically recomputed. In a future revision, we should consider adding this as a form of document metadata, so that if a corpus is subset or sliced, we don't have to recompute everything. Note: This would have been a perfect use for our now-deprecated `metadoc()` quantities.

Also adds the summary information to the three system corpora.


koheiw:
It is generally a good idea to cache data, but there are issues related to dots:
```r
> head(summary(corp, remove_punct = FALSE))
             Text Types Tokens Sentences Year  President FirstName
1 1789-Washington   625   1538        23 1789 Washington    George
2 1793-Washington    96    147         4 1793 Washington    George
3      1797-Adams   826   2578        37 1797      Adams      John
4  1801-Jefferson   717   1927        41 1801  Jefferson    Thomas
5  1805-Jefferson   804   2381        45 1805  Jefferson    Thomas
6    1809-Madison   535   1263        21 1809    Madison     James
> head(summary(corp, remove_punct = TRUE))
             Text Types Tokens Sentences Year  President FirstName
1 1789-Washington   617   1430        23 1789 Washington    George
2 1793-Washington    91    135         4 1793 Washington    George
3      1797-Adams   819   2318        37 1797      Adams      John
4  1801-Jefferson   711   1726        41 1801  Jefferson    Thomas
5  1805-Jefferson   799   2166        45 1805  Jefferson    Thomas
6    1809-Madison   531   1175        21 1809    Madison     James
```
I cannot do this on this branch:
```r
> summary(corp, remove_punct = TRUE)
Error in get_summary_metadata(object, ...) : 
  unused argument (remove_punct = TRUE)
```
I am fine to remove `...` entirely although it reduces usefulness of the summary stat. It might be  a good change because it prevent people from using the summary stat in analysis (there are many :grimacing:  ). In order to give an accurate summary on tokens, we probably need `summary.tokens()`.

kbenoit:
OK I fixed the `...` passing, works now. Although I agree that this should probably almost never be used.

Most users will not use this functionality anyway, but we can add it to system corpora and consider it a first draft for later caching of descriptive information. Mainly I need it now to avoid a performance hit on the gui app development caused by Shiny not doing much garbage cleaning after tokenizing words and sentences on the fly just for the purposes of summarizing the corpus. We can figure out ways to reimplement this later, as long as summary takes cached counts, it won't change the API (just as this PR does not change the `summary.corpus()` API).

koheiw:
How about just do this in your GUI system?
```
quanteda:::meta_system(corp, "summary") <- summary(corp, n = ndoc(x), showmeta = FALSE)
```
so we do not add `add_summary_metadata`? I am fine to export `meta_system()` if this is useful for system development?

kbenoit:
I did that at first, but it was not robust.
- some corpora don't have it, then it returns `NULL`
- there were no helper functions to add this (consistently, robustly) to the metadata
- and by encapsulating this in internal functions, it means we can change its functionality later but use the same add/get functions.

The solution here provides a more consistent and robust solution while not changing anything that existing users would notice. (Except it's much faster now with `summary.corpus()` for objects where the summary info is already stored.)

koheiw:
I am fine with `get_summary_metadata(object, ...)` as it does not change how `summary.corpus()` works unless users set the meta field. 

If you want other people to use caching or you continue to use the function in a long run, let's make something like `"cache_summary<-"` instead of add-hoc `add_system_summary()`, because it will be difficult to change the name of functions once your online system is deployed. I also think `meta_system(corp, "summary") <- `  is flexible and robust enough for system development.

kbenoit:
I'm happy not to export this pair of functions if that reassures you. The app doesn't call them in any way, it's all done through `summary.corpus()`. 

I did use `add_summary_metadata()` to add the summary info to the three built-in corpus objects, you are right that it's just a convenience wrapper to a simple function call - for now. But to make this more functional, for instance we would add something to update it by subsetting after some operation that affects the documents (i.e. `corpus_subset()`, `corpus_reshape()`, `[.corpus`, etc.). I get a huge gain just from the simple functionality even without this so it's still worth it.

So if the change you suggest is to remove `@export` I'm fine with that, then these functions are entirely internal.

Issue #1830, opened 2020-01-13
negative indexing fails on tokens objects


kbenoit:
``` r
library("quanteda")
## Package version: 1.9.9007
toks <- tokens(c("a b c", "x y z", "s d e f"))

# works
toks[c(1, 3)]
## tokens from 2 documents.
## text1 :
## [1] "a" "b" "c"
## 
## text3 :
## [1] "s" "d" "e" "f"

# works
toks[c(TRUE, FALSE, TRUE)]
## tokens from 2 documents.
## text1 :
## [1] "a" "b" "c"
## 
## text3 :
## [1] "s" "d" "e" "f"

# fails
toks[-2]
## Error in `[.tokens`(toks, -2): Subscript out of bounds
```

PR #1834, opened 2020-01-14
Issue 1830

koheiw:
Fix negative indexing using a dummy named-numeric vector #1830

### DIFF
--- a/tests/testthat/test-corpus.R
+++ b/tests/testthat/test-corpus.R
@@ -26,22 +26,6 @@ test_that("test show.corpus", {
 
 })
 
-test_that("test c.corpus", {
-    corp <- c(data_corpus_inaugural[1:2],
-              data_corpus_inaugural[3:5],
-              data_corpus_inaugural[6:10])
-
-    expect_equivalent(
-        corp,
-        data_corpus_inaugural[1:10]
-    )
-
-    expect_equal(
-        docvars(corp),
-        docvars(data_corpus_inaugural[1:10])
-    )
-
-})
 
 test_that("test corpus constructors works for kwic", {
     kw <- kwic(data_char_sampletext, "econom*")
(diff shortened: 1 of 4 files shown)


tests/testthat/test-corpus.R
```diff
@@ -544,3 +544,27 @@ test_that("as.corpus correctly sets metadata on pre-v2 corpus", {
         substring(as.character(Sys.Date()), 1, 10)
     )
 })
+
+test_that("corpus indexing works as expected", {
+  corp <- corpus(c(d1 = "one two three", d2 = "four five six", d3 = "seven eight"))
+  
+  expect_equal(corp[[1]], "one two three")
```
koheiw: Do we want the character vector to be named?
```r
expect_equal(corp[[1]], c(d1 = "one two three"))
```

kbenoit: I think so. Should be same as:
```r
> texts(corp[1])
             d1 
"one two three" 
```
Sound right to you?

kbenoit: I also think that if we follow the behaviour in https://github.com/quanteda/quanteda/wiki/indexing_core_objects (and I think we should), then `[[.corpus` should behave exactly as `texts.corpus()` does and therefore
- returned a named character
- allow multiple indexing.

The second is not true right now:
```r
> corp <- corpus(c(d1 = "one two three", d2 = "four five six", d3 = "seven eight"))
> corp[[c(1, 2)]]
Error in corp[[c(1, 2)]] : 
  attempt to select more than one element in vectorIndex
```

koheiw: I notice that `[[.corpus` is not defined anywhere! We can allow multiple indexing but it is really unusual in R
```r
> l <- as.list(letters)
> l[[1:3]]
Error in l[[1:3]] : recursive indexing failed at level 2
> 
> letters[[1:3]]
Error in letters[[1:3]] : 
  attempt to select more than one element in vectorIndex
```

kbenoit: OK, so the `[[` is consistent behaviour for a vector, which returns the unnamed contents of a single element.

```r
> c(d1 = "a", d2 = "b")[1]
 d1 
"a" 
> c(d1 = "a", d2 = "b")[[1]]
[1] "a"
```
In < v2, "[[" acts like `$.corpus` in >= v2 for returning a docvar. It's probably simpler and more consistent if we just disable `[[.corpus` and restrict `[[` to list-based objects (tokens, dictionaries).

We already cause potential confusion by hijacking (:grin:) `$`, but that's the only non-standard operator redefinition. I'd be reluctant to change `[[` as well. So: I'd say either leave it as is, or even disable it altogether.

koheiw: Let's leave it as is for now.

PR #1843, opened 2020-01-15
Internal functions for adding extended corpus-level summary information

kbenoit:
The idea here is to compute and cache (stored as `"summary_extended"` in system metadata for the corpus) overall information about the number and types of tokens, the distribution of token lengths of documents, and the frequencies of the top (100) words for use in a summary wordcloud. This will allow instant access rather than having to compute them.

Needed for quanteda.gui development (to produce a one-page quantitative and visual summary overview of a corpus) but will not be used in any active way in the R package.

The method for computing this is entirely internal and extends the previous internal method `add_summary_metadata()` through a new argument `extended = FALSE`.


kbenoit:
Yes I was thinking the same. I'll probably do that very soon, and when I do I can move this internal function to that package. But good to have it in for now, until then. Thanks.

PR #1857, opened 2020-01-26
Substantial improvement and simplification of tokens approach

### LINKED ISSUE #276: Shift tokenisation to the tokenisers package
This will involve

1. Comparing behaviours to see what is implemented in `quanteda::tokenize()` that is not currently available in **tokenizers**.  
2. Comparing performance, as a benchmark in `benchmarks/`.  Based on initial tests, **tokenizers** is faster, but could be due to the extra handling implemented in **quanteda** for the additional behaviours  - such as preserving Twitter characters otherwise deleted as punctuation - since both approaches are based on the same **stringi** calls.  
3. Replacing the core calls in `quanteda::tokens()` (new name) with calls to the **tokenizers** package.

### LINKED ISSUE #1446: tokens(x) stumbles on interior punctuation

### LINKED ISSUE #1464: tokens(x, remove_punct) behaviour is inconsistent for what = "fast(er|est)word"
Split from #1447.

Punctuation attached to the end or beginnings of words are not removed in `what = "fasterword"` or `what = "fastestword"`.

```r
> tokens("This, a 'test' of will.i.am.", what = "word", remove_punct = TRUE)
tokens from 1 document.
text1 :
[1] "This"      "a"         "test"      "of"        "will.i.am"

> tokens("This, a 'test' of will.i.am.", what = "fasterword", remove_punct = TRUE)
tokens from 1 document.
text1 :
[1] "This,"      "a"          "'test'"     "of"         "will.i.am."

> tokens("This, a 'test' of will.i.am.", what = "fastestword", remove_punct = TRUE)
tokens from 1 document.
text1 :
[1] "This,"      "a"          "'test'"     "of"         "will.i.am."
```

### LINKED ISSUE #1477: Inconsistency between remove_url and remove_twitter

### LINKED ISSUE #1503: Consider switching to post-processing of special tokens

### LINKED ISSUE #1854: New tokens() infrastructure
### Plan for major changes to `tokens()` constructor in **quanteda** v2

These major changes are designed to simplify the code and its maintenance in **quanteda**, to allow users to work with other (external) tokenizers, and to improve consistency across the tokens processing options.  

They will resolve the following tokens-related issues: #276 #1446 #1464 #1477 #1503

Changes include:

1.  A new method `tokens.list(x, ...)` constructs a `tokens` object from named list of characters, allowing users to tokenize texts using some other function (or package) such as `tokenize_words()`, `tokenize_sentences()`, or `tokenize_tweets()` from the **tokenizers** package, or the list returned by `spacyr::spacy_tokenize()`.  This will now be the recommended method for constructing a tokens object.
    
2. `tokens.tokens()` will include only `remove_*` and `split_*` rules that require an intervention; absent this intervention, nothing will happen to the tokens. To override any processing, we would have a new `as_is = TRUE` option as the default for `tokens.tokens()`

3. `tokens.character()` and `tokens.corpus()` will still work, and will call a new "default" word tokenizer that calls `stringi::stri_split_boundaries(x, type = "word")` to separate everything, including whitespace.
    
    This will then be passed to the `tokens.tokens()`, which will not remove anything by default, but will by default re-join split items for which the defaults are all `split_xxx = FALSE`.

    See the code how these are currently implemented.

4.  The main challenge in this "split everything, rejoin what is needed" are URLs and email addresses, since these are variable in length. We will need a new C++ function to rejoin URLs using this approach.

5.  When older arguments are passed, including `remove_twitter` and `remove_hyphens`  then these are acted on but a deprecation warning is issued.

6.  The `ngrams`, `skip`, and `concatenator` arguments are removed - these should be done in a second stage after forming unigrams, not in the `tokens()` call.

7.  Add `split_currency = FALSE` argument.

kbenoit:
#### Summary

**quanteda** v2 implements major changes to the `tokens()` constructor.  These are designed to simplify the code and its maintenance in **quanteda**, to allow users to work with other (external) tokenizers, and to improve consistency across the tokens processing options.  Changes include:

-  A new method `tokens.list(x, ...)` constructs a `tokens` object from named list of characters, allowing users to tokenize texts using some other function (or package) such as `tokenize_words()`, `tokenize_sentences()`, or `tokenize_tweets()` from the **tokenizers** package, or the list returned by `spacyr::spacy_tokenize()`.  
    
-  All `remove_*` options to `tokens()` now remove them from tokens objects by calling `tokens.tokens()`, after constructing the object.  "Pre-processing" is now  actually post-processing using `tokens_*()` methods internally, after a conservative tokenization on token boundaries. This both improves performance and improves consistency in handling special characters (e.g. Twitter characters) across different tokenizer engines. (#1503, #1446, #1801)    
    
-  Internal tokenizers are re-implemented in `tokenizers.R`, which input character and return a named list.  These have `split_*` options that can prevent the word tokenizer from splitting infix hyphens or social media tag punctuation characters (such as `#hashtags`).  To maintain consistency with current behaviour, an internal tokenizer accessible as `what = "word1"` is provided.
    
-  The option `remove_twitter` has been replaced in `tokens()` by `split_tags`, which preserves valid social media hashtags and usernames (using Twitter rules for validity) rather than removing the `#` and `@` punctuation characters if `remove_punct = TRUE`.
    
- The option `remove_hyphens` is removed and deprecated, but replaced by `split_hyphens`.  This preserves infix (internal) hyphens rather than splitting them.

Closes #1854
Closes #276 
Closes #1446 
Closes #1464 
Closes #1477 
Closes #1503
    
#### Overall new approach and philosophy

-  All tokens options are therefore _intervention_ options, to split or remove things that by default are not split or removed.  `tokens.tokens()` will remove what is found, but cannot "undo" a removal -- for instance it cannot replace missing punctuation characters if these have already been removed.

-  `tokens()` is a tokens _constructor_, which will call a tokenizer by default if fed character or corpus, but it's not designed to be a tokenizer. It can remove or split things, but only if the information was there to begin with. Our new function has no tokenization code in `tokens.R`, just tokens construction and custom cleaning up functions. 

- The included tokenizers in `tokenizers.R` all take character input, and output a list. These are completely independent of `tokens()`. The `what = ` argument in `tokens()` maps to these different tokenizers, but any other list can be used (so any other tokenizer, such as the ones from the **tokenizers** package, can be used if a user prefers).

#### Some performance issues

The loop across 10,000 documents is gone, but when/if restored it belongs in the `tokens.R` code and should be independent of the tokenizer function used as input. It will probably need to be in the `tokens.corpus` function. Restoring it in a new PR should be simple enough, if we think it's needed, but does not need to be in this PR.


kbenoit:
### Performance comparisons

It's slower, but this is something we can improve a lot within the new structure. The main hit comes from tokenizing twice, in effect once to preserve URLs, and a second time to split on boundaries. Our task to improve this should be:

- detailed profiling to identify the bottlenecks
- a faster implementation that preserves URLs
- using ICU `type` redefinition to stop splitting infix hyphens and tags, so we don't have to do the substitution


**This PR:**

``` r
library("quanteda")
## Package version: 1.9.9011

data_corpus_sotu <- as.corpus(data_corpus_sotu)

  word = tokens(data_corpus_sotu, what = "word"),
  word1 = tokens(data_corpus_sotu, what = "word1"),
  tokenizers = as.tokens(tokenizers::tokenize_words(texts(data_corpus_sotu),
    lowercase = FALSE,
    strip_punct = FALSE
  )),
  times = 10
)
## Unit: seconds
##        expr      min       lq     mean   median       uq      max neval
##        word 3.468142 3.676383 3.798956 3.714741 3.796351 4.871748    10
##       word1 2.635238 2.797787 2.897512 2.908115 3.013894 3.079931    10
##  tokenizers 1.344340 1.398721 1.541328 1.582230 1.657638 1.701044    10
```

**previous approach (current master branch)**

``` r
library("quanteda")
## Package version: 1.9.9009

data_corpus_sotu <- as.corpus(data_corpus_sotu)

  word = tokens(data_corpus_sotu, what = "word"),
  tokenizers = as.tokens(tokenizers::tokenize_words(texts(data_corpus_sotu),
    lowercase = FALSE,
    strip_punct = FALSE
  )),
  times = 10
)
## Unit: seconds
##        expr      min       lq     mean   median       uq      max neval
##        word 1.385352 1.445255 1.563910 1.478707 1.575289 2.240371    10
##  tokenizers 1.304198 1.330463 1.470152 1.350683 1.506135 2.171366    10
```

koheiw:
I don't have time to review properly now, but noticed that you are still using `preserve_special()`.

I found a regex with look-ahead and look-behind might be useful, although I am not sure about its speed:
```r
txt2 <- "Mr. Vane-Tempest-Stewart and Mrs. Brown are very happy. So am I, Prof. Smith., #quanteda @qi"
stri_replace_all_regex(txt2, "(?<![-#@])\\b(?!-)", "_") 
# [1] "_Mr_. _Vane-Tempest-Stewart_ _and_ _Mrs_. _Brown_ _are_ _very_ _happy_.\n_So_ _am_ _I_, 
# _Prof_. _Smith_., #quanteda_ @qi_"
```
Why do you try?

kbenoit:
Yes exactly!  We figure out ways to improve the tokenizer performance, if we want it to be smart - since the non-smart one in `tokenizers::tokenize_words()` is fast but dumb (destroys tags, URLs, hyphens, etc). 

The key point is that we now separate the tokenizer functions (segmenters) from the tokens constructor functions, and can optimize each independently. This also lets us use any tokenizer we want (including from other packages, like **spacyr** or **tokenizers**).

kbenoit:
Fixed in af4b798.

koheiw:
I like the behavior of new `word` but am reluctant to ship this until we solve the performance issue. How about handling URLs in the same way as tags and hyphens by giving them unique special token ID?

``` r

require(stringi)
hyphen <- "(\\w)\\p{Pd}+"
tag <- "[@#][\\p{L}\\p{N}]+"
url <- "https?:\\/\\/(www\\.)?[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-z]{2,4}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)"

txt <- "See #quanteda reference: http://quanteda.io/references @kohei7 #quanteda"
special <- unlist(stri_extract_all_regex(txt, paste(hyphen, tag, url, sep = "|")))
special <- unique(special)
id <- paste0("_", seq_along(special), "_") # just for demo
#id <- paste0("\uF0000", seq_along(special), "\uF0000") # private user area should be used
stri_replace_all_fixed(txt, special, id, vectorize_all = FALSE)
#> [1] "See _1_ reference: _2_ _3_ _1_"
```

kbenoit:
I thought of that too, and it's certainly something we could implement, but is it faster?

I'm ok with shipping the new "word" tokenizer even if slower, since I think once the dust has settled on this complex new release, it will be easier to focus just on tokenizer performance (and maybe get input from the **tokenizers** team). For instance maybe the solution is to have a specialist tokenizer just for URLs and tags, and one that does not need to handle them at all otherwise. I also think we can work with Marek to improve the rules for stringi word boundaries, maybe to avoid having the preserve the special characters such as infix hyphens and tags. All this will be much easier once the new approach is implemented and there is a clear separation between `tokens()` and the choice of segmenter ("tokenizer").

So I would say: If we can improve performance now, relatively easily, let's do it; if not, let's get the new infrastructure released and work on performance next. Remember a user can always drop to "word1" to get the old performance (or nearly) and behaviour. 

What could be worth looking at also here is that right now, the "word(1)" tokenizers return the substituted special characters which then need restoring in the `tokens.corpus()` code. This means our tokenizers are not fully modular. We should probably move this back to the `tokenize_word()` function. I probably did it `tokens()`-side because I thought replacing the types and re-compiling is faster. I think modularity is more important though. Think of it this way: We want tokenizers that could be put directly into the **tokenizers** package and still work for us and for anyone else.

koheiw:
I think this would be faster because it does not use many `lapply`. We must figure out which way is the best do before shipping. I don't have time now, but I always do this. This is how I made the package faster.

kbenoit:
Fine but what if we merge first and optimise second? Before release? Then I can work on the pre release checks at the same time.

R/tokenizers.R
```diff
@@ -0,0 +1,170 @@
...
+
+    # -------- apply ICU word tokenizer to everything not indexed for preservation
+    if (verbose) catm("...applying second stage smart tokenization\n")
+    x[!index_url] <- stri_split_boundaries(x[!index_url], type = "word")
+
+    # -------- convert the vector back to a list
+    if (verbose) catm("...tidying up\n")
+    x <- split(x, cut(seq_along(x), docindex, include.lowest = FALSE, labels = named))
+    # handle nested lists from 2nd stage stri_split_boundaries() call
+    x <- lapply(x, unlist)
```
koheiw: I don't think you need these slow  commands for word1 tokenization.

koheiw:
Optimization should be done before merging to keep the master branch still usable. It is fine to have experimental functions but should not be the default option. Until optimization is completed, the old tokenizer should remain "word" (and the roughly same speed and more) and new tokenizer should be "word-v2" or something.

kbenoit:
I’ll try to get around to this tomorrow, not able to to today. I wouldn’t worry about our master branch this week though - it’s a daily moving target. What matters is what we release!

kbenoit:
How about this:
``` r
library("quanteda")
## Package version: 1.9.9011
## 
## 
##     View
  word = tokens(data_corpus_sotu, what = "word1"),
  word1 = tokens(data_corpus_sotu, what = "word"),
  tword = tokenizers::tokenize_words(data_corpus_sotu) %>% as.tokens(),
  times = 5, unit = "relative"
)
## Unit: relative
##   expr      min       lq     mean   median       uq      max neval
##   word 2.664794 2.793485 2.746498 2.716197 2.770019 2.784319     5
##  word1 3.576932 3.534299 3.817012 3.542952 3.727471 4.663146     5
##  tword 1.000000 1.000000 1.000000 1.000000 1.000000 1.000000     5
```

How we could improve this further:
- use the stri_split_boundaries option to remove separators at the splitting operation, if `remove_separators = TRUE` as it almost always will be. (currently this is removed in `tokens_remove()` calls in `tokens.tokens()`.
- move the split_tags = FALSE option to "word1" - so that "regular text" is not handled like "Internet text". Our "word1" would then just be for net text. Regular speeches without @usernames #hashtags or URLs don't need the special handling.

Note that the current and possible improvements do not separate our tokenizer since it still needs the restore_special() in tokens.corpus(), but this is MUCH faster than doing it in the list-based tokenizer - about 3-4 times faster. But doing it our way does not preclude sending a separately (externally) tokenized list to `tokens.list()` or `as.tokens()` since this will not call the special handlers. (See above.)

R/tokenizers.R
kbenoit: Fixed in https://github.com/quanteda/quanteda/pull/1870

PR #1861, opened 2020-02-09
Add meta functions for dictionary2 objects

### LINKED ISSUE #1827: Add metadata to dictionary objects
Dictionaries need the same metadata as corpus objects, to record the source, license, notes, etc. I propose we add the same user and system slots to the `dictionary2` object class, and define a `meta.dictionary2()` method that works just as it does for corpus objects.

kbenoit:
- Extends all meta fns to dictionary2 objects
- Adds tests
- Improves documentation for dictionary2 objects
- Consolidates meta.R functions using fewer code
- Improves as.list.dictionary2()
- Lints files involved in these updates.

Closes #1827


kbenoit:
@koheiw I added two new slots (valuetype, meta) but we could move valuetype into meta$properties. We should also move concatenator into meta$properties.

I’d suggest merging this PR into master first, since it makes dictionary2 objects consistent with our other objects, then merging the new master with this PR in it, into your working branch for the properties revisions.

kbenoit:
OK, I think I’ll just remove valuetype from this PR, since it’s not used (yet) and we can bring it back once we get the `meta$properties` working.

Worth studying how I consolidated into one function the meta attr/slots (where an object attribute is the same as an S4 slot when it shares the slot name) for all of the object types, since this will make implementing the `meta$properties` easier.

Issue #1845, opened 2020-01-17
Add new category of object metadata: properties


kbenoit:
We currently have a problem with maintaining consistency across the attributes and slots, and after a lengthy discussion, @koheiw and I settled on the following scheme for v2:

1. `meta$system` remains exactly the same as it is now, which is _always_ updated when the object is created. This means calling `meta_system_defaults()` to get information from the current state of the user's system, and is never affected by the function call or the state of the object itself. 

    Note that this means that the `@version` slot for a dfm is removed, since this is already part of the `meta$system` list.

2. `meta$user` is unchanged.

3. Add a third "type" to the `meta` attribute/slot for all core objects: **properties**.

    This means moving the current slots (dfm) and attributes (corpus and tokens) from top-level slots or attributes, to list elements of `meta$properties`. This will include `unit`, `weightTf`, `weightDf`, `smooth`, `ngrams`, `skip`, and `concatenator`.

4. `dictionary2` class objects will also get the same structure of `meta` slot, with the "concatenator" slot moving to meta$properties. The addition of user-level metadata for dictionary objects will solve #1827.

In this way, objects now will carry a minimal set of slots and attributes, namely

| Attribute/slot | corpus | tokens | dfm | dictionary |
|----------------|:------:|:------:|:---:|:----------:|
| docvars        |    ✓   |    ✓   |  ✓ (slot)  |            |.
| meta           |    ✓   |    ✓   |  ✓ (slot)  |      ✓ (slot)    |
| types          |        | ✓      |     |            |
| padding      |        | ✓      |     |            |

(plus base object slots such as "names" or those inherited from **Matrix**, in the case of a dfm)

koheiw:
Started working on this. My understanding was this for corpus
```r
    structure(x,
              names = names,
              class = "corpus",
              properties = list(
                  "version" = unlist(utils::packageVersion("quanteda")),
                  "unit" = unit
                  ),
              meta = list(
                  "system" = meta_system_defaults(source),
                  "user" = meta
                  ),
              docvars = docvars)
```
The difference from your outline is that `properties` is the one of the three top-level slots. We could make them to two, but `meta` should be under `properties` in this case, because the latter has a greater importance.

koheiw:
How about this?
```r
structure(x,
          names = names,
          class = "corpus",
          meta = list("system" = meta_system_defaults(source),
                      "object" = list(unit, skip, ngram, etc),
                      "user" = list()),
          docvars = docvars)
```

[...]

Issue #1828, opened 2020-01-09
Remove units from dfm slot


kbenoit:
In https://github.com/quanteda/quanteda/commit/9ab6f29a91397ccb85c70306a5a8234e063abdfc#diff-0f9641f8839509fc76ca3c9c590156ed we added a `units` slot to dfm objects, but this is now causing problems:

- for packages with older dfm objects, it raises an S4 error, e.g.
```r
> str(RNewsflow::rnewsflow_dfm)
Formal class 'dfm' [package "quanteda"] with 16 slots
  ..@ settings    : list()
  ..@ Dim         : int [1:2] 1754 6968
  ..@ Dimnames    :List of 2
  .. ..$ docs    : chr [1:1754] "35573532" "35573536" "35573539" "35620308" ...
  .. ..$ features: chr [1:6968] "person.688" "noun.1516" "location.119" "organization.323" ...
  ..@ weightTf    :List of 3
  .. ..$ scheme: chr "count"
  .. ..$ base  : NULL
  .. ..$ K     : NULL
  ..@ weightDf    :List of 5
  .. ..$ scheme   : chr "unary"
  .. ..$ base     : NULL
  .. ..$ c        : NULL
  .. ..$ smoothing: NULL
  .. ..$ threshold: NULL
  ..@ smooth      : num 0
  ..@ ngrams      : int 1
  ..@ skip        : int 0
  ..@ concatenator: chr "_"
  ..@ version     : int [1:3] 1 3 0
  ..@ docvars     :'data.frame':	1754 obs. of  6 variables:
  .. ..$ docname_  : chr [1:1754] "35573532" "35573536" "35573539" "35620308" ...
  .. ..$ docid_    : Factor w/ 1754 levels "35573532","35573536",..: 1 2 3 4 5 6 7 8 9 10 ...
  .. ..$ segid_    : int [1:1754] 1 1 1 1 1 1 1 1 1 1 ...
  .. ..$ date      : POSIXct[1:1754], format: "2013-06-01 05:00:00" "2013-06-01 05:00:00" "2013-06-01 05:00:00" "2013-06-01 18:33:00" ...
  .. ..$ source    : chr [1:1754] "Print NP 2" "Print NP 2" "Print NP 2" "Online NP 1" ...
  .. ..$ sourcetype: chr [1:1754] "Print NP" "Print NP" "Print NP" "Online NP" ...
[...]

PR #1866, opened 2020-02-13
Dev property

### LINKED ISSUE #1264: Need a way to record the valuetype for pattern matching in a dictionary
Here's a case where our default `valuetype  = "glob"` assumption would be broken:

```r
data(data_dictionary_HuLiu, package = "quanteda.dictionaries")
data_char_HuLiu <- unlist(as.list(data_dictionary_HuLiu))

# detect what could appear to be a glob pattern
asterisks_index <- stringi::stri_detect_fixed(data_char_HuLiu, "*")
data_char_HuLiu[asterisks_index]
## negative468  negative811 negative1560 negative3765 
##  "bull****"       "d*mn"       "f**k"       "sh*t"

# but it's a literal
damn_index <- stringi::stri_detect_regex(data_char_HuLiu, "^d.*mn")
data_char_HuLiu[damn_index]
## negative811 negative816 negative817 negative818 negative819 negative820 negative821 
##      "d*mn"      "damn"  "damnable"  "damnably" "damnation"    "damned"   "damning" 
```

### Solution

Record this a slot called `@valuetype`. This will be the default for matching on a dictionary unless the user overrides it by specifying an explicit `valuetype` for any matching operation.

koheiw:
Add `meta$object` to record object-related information for #1845. Main changes are 
- `build_*` and `rebuild_*` functions that mimic S4's `new()` method
- Related internal functions to handle the meta attribute
- Functions that require `concatenator` and `ngram` values.
- Improve consistency in how functions are written.

Solves #1845 
Solves #1828 
Closes #1264


R/utils.R
```diff
@@ -134,23 +134,25 @@ pattern2list <- function(pattern, types, valuetype, case_insensitive,
         }
         attr(result, "pattern") <- match(names(result), names(temp))
     } else {
-        if (length(pattern) == 0) return(list())
-        if (is.dictionary(pattern)) {
-            temp <- flatten_dictionary(pattern, levels)
+        if (length(x) == 0) return(list())
+        if (is.dictionary(x)) {
+            x <- as.dictionary(x)
+            attrs <- attributes(x)
```
kbenoit: `attrs` is not used in this function

R/dfm_weight.R
```diff
@@ -69,7 +73,7 @@ dfm_weight <- function(
     scheme = c("count", "prop", "propmax", "logcount", "boolean", "augmented", "logave"),
     weights = NULL,
     base = 10,
-    K = 0.5,
```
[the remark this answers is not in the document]
kbenoit: I can live with this though - probably not many people will have ever used this parameter.

R/dfm_weight.R
```diff
@@ -241,6 +241,23 @@ dfm_smooth.dfm <- function(x, smoothing = 1) {
 #' feature occurs more than a given frequency threshold.  (The default threshold
 #' is  zero, meaning that any feature occurring at least once in a document will
 #' be counted.)
+#' @param x a [dfm]
```
kbenoit: I had to restore these because although the argument names are the same as in `dfm_weight()`, the documentation entries are different (except `threshold`).

kbenoit:
This is ready to go IMO but first I'd like you to:
- check the unused variables queries above
- can you document the internal functions in meta.R a bit better?

R/corpus-methods-base.R
```diff
@@ -193,24 +193,23 @@ tail.corpus <- function(x, n = 6L, ...) {
 `+.corpus` <- function(c1, c2) {
     c1 <- as.corpus(c1)
     c2 <- as.corpus(c2)
+    attrs1 <- attributes(c1)
+    attrs2 <- attributes(c2)
```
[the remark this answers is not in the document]
koheiw: I wasn't entirely sure what to do with meta from `c2`. We can block combining corpora with different `unit`, attach user meta from `c1` and `c2`, etc.

R/dfm.R
```diff
@@ -299,9 +322,11 @@ dfm.dfm <- function(x,
                     verbose = quanteda_options("verbose"),
                     ...) {
 
+    unused_dots(...)
+    
     x <- as.dfm(x)
     valuetype <- match.arg(valuetype)
-    check_dots(list(...))
+    attrs <- attributes(x)
```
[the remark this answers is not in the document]
koheiw: I will take this out.

R/docvars.R
```diff
@@ -141,7 +141,9 @@ subset_docvars <- function(x, i = NULL) {
 }
 
 # Reshape docvars keeping variables that have the same values within groups
-group_docvars <- function(x, group) {
+group_docvars <- function(x, group = NULL) {
```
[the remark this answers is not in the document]
koheiw: Right. `docnum` is not longer used.

R/corpus-methods-base.R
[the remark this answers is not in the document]
kbenoit: not really sure what is best here, but for now, I'd say match previous behaviour.

R/utils.R
koheiw: Right.

R/tokens.R
```diff
@@ -280,22 +280,9 @@ tokens.corpus <- function(x,
...
-            unlist(x, recursive = FALSE), 
-            types = attr(x[[length(x)]], "types"),
-            what = what, 
-            ngram = 1L, 
-            skip = 0L,
-            concatenator = "_",
-            docvars = docvars,
-            meta = attrs[["meta"]]
-    )
+    return(result)
```
koheiw: We have to use `build_tokens()` always to make sure that object has all the meta fields.

kbenoit: yes but the previous call to as.tokens() uses build_tokens()

R/meta.R
```diff
@@ -115,44 +152,241 @@ check_valid_meta_object <- function(x) {
...
+        } else if (length(inherit) && m %in% names(inherit)) {
+            if (!identical(class(default[[m]]), class(inherit[[m]])))
+                stop("Invalid inherit object for ", m)
+            default[[m]] <- inherit[[m]]
+        }
+    }
+    return(default)
+}
+
+#' Internal functions to access meta filed in a list of attributes
```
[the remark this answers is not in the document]
koheiw: Done

R/tokens.R
```diff
@@ -248,9 +248,8 @@ tokens.corpus <- function(x,
     if ("remove_twitter" %in% names(dots)) {
         .Defunct(msg = "'remove_twitter' is defunct, use 'what = \"word2\"' instead.")
     }
-    unused_dots(dots)
+    check_dots(dots, c(names(formals(tokens))))
```
koheiw: I changed all the `check_dots()` to simpler `unused_dots()` in all the functions. We should choose either.

kbenoit: unused_dots() is fine, but I needed to be able to change (...) in the parent function (for trapping deprecated arguments) and then send the resulting (modified) list to the check_dots(). You cannot use list(...) in the child fn if it's already been evaluated - then it's a list one level deep. So just to get this branch working, for now, I added back the previous fn. But you are right we only should use one.

R/tokens.R
kbenoit: but fix it as you need to, whatever gets it working for now, we can tidy up later.

koheiw:
I noticed that the blocked serialization has been removed. We should restore this, otherwise it becomes impossible to tokenize lerge texts with limited RAM.
https://github.com/quanteda/quanteda/blob/40de44e973fdf7deccb769b0ff9e4900ff75f8c8/R/tokens.R#L453-L479

kbenoit:
Makes sense, but let’s add it back as a separate fix after merging this monster. We can clean this up, get it merged, and then discuss the best place to batch the tokens construction.

koheiw:
Done!

Issue #1638, opened 2019-03-07
add corpus_sort() or an option for controlling grouping order


kbenoit:
## Requested feature

What if we want to reshape a set of documents into single "document" but sorted, say by chronological order, and these were never reshaped into smaller documents in the first place?

For instance, what if I had a set of speeches across times, and wanted to plot usages of a term across time for all speeches as a single "document", so that I I could plot a lexical diversity graph for all speeches as a "document", by date?

We would need a way to "group" or reshape the documents into a larger document, while controlling the order. This could be done either with a `corpus_sort()` and then have the grouping documents use that order - which is how for  instance `texts(x, groups = "something")`, by using the existing ordering - or by adding an argument for sort order.

koheiw:
I you used `dfm_group()` with factor properly ordered levels, you can do it. I make DFMs where documents group of all news published on the same day in this manner.

kbenoit:
True but if we wanted a lexical diversity plot, we'd have to group texts in order and then tokenize them.

koheiw:
Then, how about using `tokens_group()`? Do you want `corpus_group()`? I think better to compute things on dfms or tokens if possible.

kbenoit:
Closed because this already works via indexing (`[`), and is also possible via `quanteda.tidy::arrange()`.

PR #1867, opened 2020-02-13
Remove group for NA values of groups variable

### LINKED ISSUE #1859: group operations crash on NA in docvars e.g. dfm_group()
`dfm_group()` crashes when a docvar has `NA`. Apparently we never considered what would/should happen if a grouping docvar contains missing. Options seem simple through,
1. "drop" - Ignore NA documents and drop them from the grouped dfm.
2. "group" - Group the documents together as if `NA` were a value to use, grouping them and assigning this new "document" `NA` on the docvar used to group. In the document name, this would simply be `NA` (after we check that there is not already a document with that name).
3. "keep" - Keep the `NA` valued documents on the docvar unchanged, with their original docnames but not grouped. 

Maybe a new argument as in `table()`: `useNA = c("group", "drop", "keep")` with group as the default?

This would also affect `texts()` and the (internal) `tokens_group()`.

``` r
library("quanteda")
## Package version: 1.9.9011

corp <- corpus(c("Doc 1", "Doc 1b", "Doc2", "Doc 3 with NA", "Doc 4, more NA"),
  docvars = data.frame(factorvar = c("Yes", "Yes", "No", NA, NA))
)
summary(corp)
## Corpus consisting of 5 documents, showing 5 documents:
## 
##   Text Types Tokens Sentences factorvar
##  text1     2      2         1       Yes
##  text2     2      2         1       Yes
##  text3     1      1         1        No
##  text4     4      4         1      <NA>
##  text5     5      5         1      <NA>

dfm(corp[1:3], groups = "factorvar")
## Document-feature matrix of: 2 documents, 4 features (50.0% sparse) and 1 docvar.
##      features
## docs  doc 1 1b doc2
##   No    0 0  0    1
##   Yes   2 1  1    0
dfm(corp, groups = "factorvar")
## Error in qatd_cpp_is_grouped_numeric(as.numeric(x), group): Invalid groups
```

kbenoit:
Any group whose value is NA is dropped from the grouped object.

- fixes #1859 (`dfm_group()`)
- also updates `tokens_group()`
- makes `texts(x, groups = ...)` consistent
- updates `?groups` documentation

Made as a PR to the dev-property branch because it is simpler than sorting out merge conflicts after we merge `dev-property`.


kbenoit:
I think we discussed that originally and decided against it on the basis that a corpus should remain an original archive of source texts, and any redefinition of the documents should be done only “downstream”. True that we have `corpus_reshape()` but the units that it creates are syntactically defined and retains the information needed to reshape back to the original documents.

Anyway this is a perfectly sensible feature request but let’s not hold up a PR for another purpose with a feature request we happen to see along the way. My change here was just to make something that already exists more consistent.

kbenoit:
I think the other issue with `corpus_group()` is the same as we encountered for `tokens_group()`, which is that a user needs to define a sequence for concatenation since these are ordered. (See for example https://github.com/quanteda/quanteda/issues/1638). Since grouping a corpus or tokens object is essentially a concatenation operation, sequence matters.

This is exactly why `tokens_group()` is not currently exported.

PR #1870, opened 2020-02-17
Separate word and word2 and hard-wire split_tags

kbenoit:
This is a more efficient change to the built-in word tokenisers for the `new-tokens-approach` branch. 

#### "word"

- besides minimal special processing (to remove for variant selector & whitespace with diacritical marks), this preserves infix hyphens only. It's like the existing `what = "word"` except that it does no special handling for tags.

#### "word2"

- like "word1" except this preserves URLs, email addresses, #hashtags, and @usernames. This is what we would recommend for users that have Internet-speak in their texts.

### Performance comparisons

"word" will never be as fast as `tokenizers::tokenize_words()` because that function just uses the default stringi handling to split words and remove spaces at the same time. Because stringi cannot remove spaces *and* non-word characters like punctuation and symbols separately, and we don't want to remove punctuation and symbols by default, we cannot avail of that speed advantage. Our "fasterword" however does a pretty good job at matching it.

``` r
library("quanteda")
## Package version: 1.9.9011

  tokenize_words = tokenizers::tokenize_words(data_corpus_sotu) %>% as.tokens(),
  what_fasterword = tokens(data_corpus_sotu, what = "fasterword"),
  what_word = tokens(data_corpus_sotu, what = "word"),
  what_word2 = tokens(data_corpus_sotu, what = "word2"),
  times = 5, unit = "relative"
)
## Unit: relative
##             expr      min       lq     mean   median       uq      max neval
##   tokenize_words 1.000000 1.000000 1.000000 1.000000 1.000000 1.000000     5
##  what_fasterword 1.078689 1.025861 1.074574 1.024942 1.085836 1.154596     5
##        what_word 2.706993 2.517935 2.643938 2.437187 2.528844 3.026473     5
##       what_word2 3.598059 3.401355 3.468727 3.270389 3.541447 3.542958     5
```


kbenoit:
I'll merge it and we will assess the damage afterward when the dust settles!

Issue #1871, opened 2020-02-17
Improve the performance of word2 tokenizer


koheiw:
Word2 tokenizer is available now, but we need to make it faster before makig it the default. There is the alternative way to protect URLs (and any other special tokens) from segmentation.

```r
require(quanteda)
require(stringi)

tokenize_word2 <- function(x, split_hyphens = FALSE, verbose = quanteda_options("verbose")) {
    named <- names(x)
    
    # -------- special processing
    # this will not be needed if we can modify the ICU type rules to protect them
    # remove variant selector & whitespace with diacritical marks
    x <- stri_replace_all_regex(x, c("[\uFE00-\uFE0F]", "\\s[\u0300-\u036F]"), "",
                                vectorize_all = FALSE)
    # substitute characters not to split
    x <- quanteda:::preserve_special(x, split_hyphens = split_hyphens, split_tags = FALSE, verbose = verbose)
    # pad URLs with an additional space - for e.g. Japanese text with URLs
    regex_url <- "https?:\\/\\/(www\\.)?[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-z]{2,4}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)"
    x <- stri_replace_all_regex(x, regex_url, " $0 ")
    # make NAs ""
    x[is.na(x)] <- ""
    
    if (verbose) catm("...segmenting tokens\n")
    
    # -------- initial split on white space
    # this and the index_url can be removed if we are able to modify the rules 
    # in type for stri_split_boundaries()
    x <- stri_split_regex(x, "[\\p{Z}\\p{C}]+", omit_empty = FALSE)
    if (verbose) catm("...vectorizing tokens\n")
    doc_lengths <- cumsum(lengths(x))
    docindex <- c(0, doc_lengths)
    x <- unlist(x)
    index_url <- stri_detect_fixed(x, "://")
    
    # -------- second-stage "smart" tokenization
    if (verbose) catm("...applying second stage smart tokenization\n")
    x[!index_url] <- stri_split_boundaries(x[!index_url], type = "word")
    if (verbose) catm("...tidying up\n")
    x <- split(x, cut(seq_along(x), docindex, include.lowest = FALSE, labels = named))
    # handle nested lists from 2nd stage stri_split_boundaries() call
    x <- lapply(x, unlist)
    
    return(structure(x, names = named))
}

tokenize_word_test <- function(x) {
    m <- names(x)
[...]

PR #1876, opened 2020-02-18
Substitute special tokens with IDs in preserve_special2()

koheiw:
As proposed in #1871, this PR replaced the two-staged tokenization with more generalized preserve special, `preserve_special2()` amd `restore_special2()`. These function outside of tokenizers so can be used with user-provided tokenizers.

With this PR, `word2` is closer to the `word`:
```r
> corp_en <- readRDS("~/Dropbox/Brexit/Data/data_corpus_guardian_2016.RDS")
> txt_en <- head(texts(corp_en), 1000)
+     word2_en = tokens(txt_en, "word2"),
+     word_en = tokens(txt_en, "word"),
+     word2_ja = tokens(txt_ja, "word2"),
+     word_ja = tokens(txt_ja, "word"),
+     times = 10
+ )
Unit: milliseconds
     expr       min        lq      mean    median        uq      max neval
 word2_en 1744.1752 1759.4419 1810.5149 1804.0569 1849.0080 1913.644    10
  word_en 1361.0224 1387.4611 1466.3019 1434.5720 1534.2856 1713.264    10
 word2_ja  725.1376  740.6773  813.6235  766.1683  784.4199 1285.556    10
  word_ja  854.9628  860.9563  926.2580  882.9542  929.9040 1249.802    10
```

However, it is failing the tests because of how the two-staged `word2` tokenizer handled separators: all the white spaces are removed in the fist-stage.

This branch's `word2`
```r
> txt <- c(doc1 = "One 2, £ https://qunteda.org one-two.")
> tokens(txt, what = "word2", remove_punct = TRUE, padding = TRUE)
doc1 :
 [1] "One"                 ""                    "2"                   ""                    ""                   
 [6] "£"                   ""                    "https://qunteda.org" ""                    "one-two"            
[11] ""
```
It is great to keep padding for removed punctuation marks but not for the white spaces.  We can either apply `tokens_remove(x, " ", padding = FALSE)` separately or drop " " in `serialize_tokens()` automatically. The latter option is preferred from performance point of view.


kbenoit:
Why don’t we trap the remove_separators in the pre-tokenisation stage and do the higher performance version when - as it will be 99% of the time - remove_separators = TRUE? Then we can flip it to false before it’s passed (again) from tokens_corpus() to tokens_remove().

We don’t want a special token reserved for “ “, because it might be anything: “\n” or “\t” for example. “word2” already always removes it. 

Also, it would not be the end of the world to remove it entirely, and just require a user to use a different tokenizer of their own if they really want spaces as tokens.

koheiw:
I am not sure what you mean by "trap the remove_separators in the pre-tokenisation". Do you agree to remove " " in `serialize_tokens()`?

kbenoit:
I meant that right now, we remove everything in tokens.tokens() (except for word2 where there are no separators to remove). We could take that approach in tokenize_word() by passing it remove_separators and removing them there, before compiling the tokens. Some things are just going to be faster to take out in our tokenizer, rather than in “post” (meaning in tokens.tokens()). Of course `tokens.tokens()` should still be about to remove them all, but if it’s called as a chain starting with tokens.corpus() then we can flip off a removal before this is passed to the final tokens.tokens(), if it has already been removed in the tokenize_word() stage.

We should not hard-wire removal of spaces in serialize_tokens() because we might have a need to make “ “ a token. My view is that yes we reserve “” for a null token, but we should not privilege “ “ (or “\n”, “\t“, etc.). When I use `spacy_tokenize()`, for instance, I get all of those separators and if I want to make them tokens, then I should be able to.

koheiw:
I tested with pre and post-removal of separators. Pre-removal was a bad idea as it requires slow `lapply`. The best way seems to add another `tokens_remove()` to `tokens.tokens()`.

```
> library(quanteda)
+     word2_en = tokens(txt_en, "word2"),
+     word_en = tokens(txt_en, "word"),
+     word2_ja = tokens(txt_ja, "word2"),
+     word_ja = tokens(txt_ja, "word"),
+     times = 10
+ )
Unit: milliseconds
     expr       min        lq      mean    median        uq      max neval
 word2_en 1767.4059 1775.8060 1824.9271 1811.8583 1860.3060 1924.319    10
  word_en 1396.4410 1412.4600 1548.3291 1466.8260 1531.3121 2180.731    10
 word2_ja  764.6381  773.0506  897.0279  797.9992 1012.1505 1274.034    10
  word_ja  895.9842  902.1396  973.0426  917.7579  937.6655 1437.076    10
```

kbenoit:
Not scaling well?

``` r
library("quanteda")
## Package version: 2.0.0

  tokenize_words = tokenizers::tokenize_words(data_corpus_sotu) %>% as.tokens(),
  what_fasterword = tokens(data_corpus_sotu, what = "fasterword"),
  what_word = tokens(data_corpus_sotu, what = "word"),
  what_word2 = tokens(data_corpus_sotu, what = "word2"),
  times = 5, unit = "relative"
)
## Unit: relative
##             expr      min       lq     mean   median       uq      max neval
##   tokenize_words 1.629047 1.741336 1.647931 1.725761 1.639980 1.551682     5
##  what_fasterword 1.000000 1.000000 1.000000 1.000000 1.000000 1.000000     5
##        what_word 3.117787 3.060240 2.829559 3.023437 2.708307 2.461448     5
##       what_word2 2.278114 2.322723 2.074880 2.241550 2.041050 1.695854     5

data_corpus_guardian <- as.corpus(quanteda.corpora::download("data_corpus_guardian"))
  tokenize_words = tokenizers::tokenize_words(data_corpus_guardian) %>% as.tokens(),
  what_fasterword = tokens(data_corpus_guardian, what = "fasterword"),
  what_word = tokens(data_corpus_guardian, what = "word"),
  what_word2 = tokens(data_corpus_guardian, what = "word2"),
  times = 5, unit = "relative"
)
## Unit: relative
##             expr       min        lq      mean    median        uq       max
##   tokenize_words  1.450000  1.445348  1.310506  1.201663  1.193129  1.317673
##  what_fasterword  1.000000  1.000000  1.000000  1.000000  1.000000  1.000000
##        what_word  2.558717  2.521967  2.279342  2.121535  2.162060  2.140848
##       what_word2 14.522001 13.926482 12.637266 11.679306 11.616751 12.104254
##  neval
##      5
##      5
##      5
##      5
```

R/tokenizers.R
```diff
@@ -39,60 +39,35 @@ NULL
 #'   stri_replace_all_regex stri_detect_fixed stri_replace_all_fixed
 #' @export
 tokenize_word <- function(x, split_hyphens = FALSE, verbose = quanteda_options("verbose")) {
-    named <- names(x)
+    
+    m <- names(x)
```
kbenoit: This is more clear than `named`? 🙄

R/tokenizers.R
```diff
@@ -39,60 +39,35 @@ NULL
...
 
     if (verbose) catm("...segmenting tokens\n")
-    structure(stri_split_boundaries(x, type = "word"), names = named)
+    structure(stri_split_boundaries(x, type = "word"), names = m)
 }
 
 #' @rdname tokenize_internal
 #' @importFrom stringi stri_replace_all_regex stri_detect_fixed stri_split_boundaries
 #' @export
 tokenize_word2 <- function(x, split_hyphens = FALSE, verbose = quanteda_options("verbose")) {
```
kbenoit: would have made more sense to create a word3 for comparisons to word2...

R/tokens.R
```diff
@@ -262,6 +262,13 @@ tokens.corpus <- function(x,
                            fasterword = tokenize_fasterword,
                            fastestword = tokenize_fastestword)
     
+    
+    if (what == "word2") {
+        x <- preserve_special2(x, split_hyphens = split_hyphens, 
+                               split_tags = FALSE, verbose = verbose)
```
kbenoit: I had to add `split_tags = FALSE` because otherwise it was always letting tags be split.

R/tokens.R
```diff
@@ -341,17 +351,20 @@ tokens.tokens <-  function(x,
...
+        x <- tokens_remove(x, "^[\\p{Z}\\p{C}]$", valuetype = "regex")
+    
     # removals
     removals <- compile_removals_regex(remove_separators = remove_separators,
                                        remove_punct = remove_punct,
                                        remove_symbols = remove_symbols,
                                        remove_numbers = remove_numbers,
                                        remove_url = remove_url)
-    if (length(removals$regex_to_remove)) {
+    if (length(removals$regex_to_remove) || remove_separators) {
```
kbenoit: A small hack to print the separators removal message, even though it's been removed at line 358.

R/tokens.R
```diff
@@ -465,7 +478,7 @@ compile_removals_regex <- function(remove_separators = FALSE,
     regex_to_remove <- removing_msg <- character()
 
     if (remove_separators) {
-        regex_to_remove <- c(regex_to_remove, "^[\\p{Z}\\p{C}]$")
+    #    regex_to_remove <- c(regex_to_remove, "^[\\p{Z}\\p{C}]$")
         removing_msg <- c(removing_msg, "separators")
```
kbenoit: adding the message but not adding this to the regex restores the message whose absence was causing test failures.

R/tokens.R
```diff
@@ -341,17 +351,20 @@ tokens.tokens <-  function(x,
         x <- tokens_split(x, "\\p{Pd}", valuetype = "regex", remove_separator = FALSE)
     }
     
+    if (remove_separators)
+        x <- tokens_remove(x, "^[\\p{Z}\\p{C}]$", valuetype = "regex")
```
kbenoit: you're sure this is faster than putting it into the removals_regex and calling `tokens_remove()` just once?

koheiw:
It was slower with larger corpora, because it scanned all documents for all special strings, but managed to solve this. With this optimization, `word2` became faster than `word`.

```r
 Unit: seconds
   word_en_1k  1.418906  1.451193  1.646914  1.551313  1.870490  1.951685    10
   word_en_5k  7.545498  7.727867  8.236853  8.347751  8.525282  8.901136    10
  word_en_10k 16.289153 16.485207 16.989550 16.977117 17.269965 17.841597    10
  word2_en_1k  1.136081  1.164554  1.282639  1.247415  1.305638  1.708992    10
  word2_en_5k  5.863662  5.870298  6.274490  6.323972  6.516298  7.054925    10
 word2_en_10k 12.447537 12.899336 13.427562 13.488442 13.959334 14.741451    10
 
 Unit: milliseconds
         expr       min        lq      mean   median         uq       max neval
   word_ja_1k  878.9190  907.5938 1091.9535 1027.054  1288.2462  1397.442    10
   word_ja_5k 4592.7310 4844.2732 5066.1054 5026.551  5263.8228  5646.483    10
  word_ja_10k 9020.1024 9239.1634 9690.9906 9443.367 10173.8194 10785.279    10
  word2_ja_1k  273.8306  284.5448  308.7543  290.818   302.5738   475.694    10
  word2_ja_5k 1379.6971 1404.3934 1504.0059 1495.726  1565.5941  1692.376    10
 word2_ja_10k 2645.1059 2656.7323 2813.7219 2696.457  3005.2131  3206.479    10
```

R/tokens.R
koheiw: It is faster to do it in one-go with `removals_regex` for sure, but I need to add a `tokens_remove()` to avoid paddings for separators when `padding = TRUE`.

R/tokenizers.R
koheiw: I think it is common to use short names for variables with short life-span.

R/tokenizers.R
koheiw: Yes, but the new `word2` is faster than `word` already.

kbenoit:
🎸🎸🎸🎸

koheiw:
Correction: benchmarking with Japanese texts was not accurate due to the issues that I fixed in the latest commit.
```r
+     word_ja_1k = tokens(txt_ja_1k, "word"),
+     word_ja_5k = tokens(txt_ja_5k, "word"),
+     word_ja_10k = tokens(txt_ja_10k, "word"),
+     word2_ja_1k = tokens(txt_ja_1k, "word2"),
+     word2_ja_5k = tokens(txt_ja_5k, "word2"),
+     word2_ja_10k = tokens(txt_ja_10k, "word2"),
+     times = 10
+ )
Unit: milliseconds
         expr       min         lq       mean     median        uq       max neval
   word_ja_1k  932.0972  1036.3290  1213.0805  1117.1693  1446.074  1670.617    10
   word_ja_5k 4984.9916  5257.9704  5525.1840  5463.5975  5808.509  6136.867    10
  word_ja_10k 9716.8197 10037.0383 10967.7669 10375.9061 12187.892 13689.974    10
  word2_ja_1k  800.4299   829.0472   931.7296   880.1505  1022.604  1218.403    10
  word2_ja_5k 4166.2125  4315.3707  4574.0080  4553.1755  4874.337  5021.240    10
 word2_ja_10k 8185.3256  8451.5661  9171.1840  9163.4430  9835.932 10546.645    10
```
This is a reasonable amount  improvement.

R/tokens.R
kbenoit: Right - actually this last line should not remove the pad.
```r
> as.tokens(list(d1 = c("one", " ", "x", "two"))) %>%
+     tokens_remove("x", padding = TRUE) %>%
+     tokens(remove_separators = TRUE)
d1 :
[1] "one" "two"
```
but if we could fix the regex to remove spaces but not pads, we could do it in one expression.

R/tokenizers.R
kbenoit: Let's swap them back then, and have word2 become word, and word become word1 to indicate it's a legacy tokenizer that is inferior but kept just for backward compatibility.

R/tokenizers.R
kbenoit: 😣

kbenoit:
For the future, it would seem like tokenization above a certain number of documents is a perfect task for parallelization. Let's say our batch size is 10,000. We split a corpus into k 10,000-document batches, tokenize and hash (what you call serialize) each batch assigning token indexes starting at (0:(k-1))*10000, and when done, concatenate and call `tokens_recompile()`. A simple version of classic map-reduce.

If we could generalize the function for this, we could supply a tokenizer function that could be pretty much anything, from any package, that returned a named list.

koheiw:
This tokenizer redesigning should open a way to parallelization!

koheiw:
If you are happy with changes in this branch, you can make `word2` the default.

koheiw:
I think the warning does not apply anymore.
```r
> print(tokens(txt_en_1k, what = "word2", remove_separators = FALSE), 1)
Tokens consisting of 1,000 documents.
text119050 :
 [1] "Democratic"   " "            "presidential" " "            "candidate"    " "            "Bernie"      
 [8] " "            "Sanders"      " "            "raised"       " "           
[ ... and 1,019 more ]

[ reached max_ndoc ... 999 more documents ]
Warning message:
In tokens.corpus(corpus(x), what = what, remove_punct = remove_punct,  :
  remove_separators is always TRUE for this type
```

R/tokenizers.R
kbenoit: Done in fcc2e6f

Issue #1878, opened 2020-02-18
build functions for objects are not friendly with inherited classes


kbenoit:
**sentometrics** is breaking in the current `master` because our build functions drop additional class attributes. In 
https://github.com/sborms/sentometrics/blob/master/R/sentocorpus.R#L141-L142
for instance, `sento_corpus()` creates a quanteda corpus and then adds an additional class label (after giving it some docvars that are special for a sento_corpus).

But as you can see below, basic corpus functions that **sentocorpus** uses, such as `corpus_sample()`, erase the added class label. 

More generally, this means that additional package developers cannot create child classes (inasmuch as such a notion exists in the primitive S3 system) of **quanteda** objects without having our functions erase their added attributes. If the objects on which we build **quanteda**'s objects behaved like this, we could never have built our package. This seems like something we should strive to avoid generally.

``` r
library("quanteda")
## Package version: 2.0.0

corp <- corpus(letters[1:3],
  docvars = data.frame(dv = 11:13),
  meta = list(mymeta = "Test meta")
)

mycorp <- corp
class(mycorp) <- c("myclass", class(mycorp))

class(corpus_sample(mycorp, size = 2))
## [1] "corpus"
class(corpus_subset(mycorp, dv == 1))
## [1] "corpus"

docvars(mycorp, "dv2") <- c("one", "two", "three")
class(mycorp)
## [1] "myclass" "corpus"
```

kbenoit:
Our S4 objects suffer the same problem. (Here, it happens in the `build_dfm()` call in `subset_dfm()`.)

``` r
dfmat <- dfm(corp)

setClass("dfmchild",
  slots = c(
    myslot1 = "character",
    myslot2 = "integer"
  ),
  contains = c("dfm")
)

dfmatchild <- as(dfmat, "dfmchild")
dfmatchild@myslot1 <- "test"
dfmatchild@myslot2 <- "42"
## Error in (function (cl, name, valueClass) : assignment of an object of class "character" is not valid for @'myslot2' in an object of class "dfmchild"; is(value, "integer") is not TRUE

dfm_sample(dfmatchild, size = 2) %>%
  is("dfmchild")
## [1] FALSE
dfm_sample(dfmatchild, size = 2) %>%
  slotNames()
## [1] "docvars"  "meta"     "i"        "p"        "Dim"      "Dimnames" "x"       
## [8] "factors"
```

kbenoit:
[...]

Issue #1874, opened 2020-02-18
sentometrics is breaking


kbenoit:
A host of issues arising, need to check and fix.

PR #1879, opened 2020-02-19
Keep class in build  functions

kbenoit:
Solves #1878 for the S3 core objects (corpus, tokens), by allowing the build functions to pass through added class labels for object created from corpus or tokens objects.

Does not solve #1878 for the S4 classes of dfm and dictionary2 objects.

This stops the breaking problems in **sentometrics** in #1874.

Solves part of #1874, along with


koheiw:
I am not sure why we need to accept modified objects. I welcome users to extend our objects and assign a special class, but I also expect them to define their own methods for the class. Accepting modified object to the `build_*` function risks integrity of our objects. This is why modification of S4 is not allowed without formally defining a class, isn't it?

kbenoit:
I think we would both agree that a user (or package) should be able to build a class based on a core **quanteda** object class, but extend it. They would not want to recreate all core functions - especially absolutely basic ones such as `[` (!) for their new object. Instead, the functions we have defined for the parent class (say, a `corpus` object) should work on the child object in the new package.

This is *exactly* what we have done with a dfm for instance: it is a child class of `dgCMatrix`. 

And this is exactly what the **sentomentrics** package does with our `corpus` class: it creates a child class that is a corpus but with a special type of docvar. Before PR #1866, this worked fine, but after the change created `build_corpus()` that constructs a new corpus in response to some modifications, the extra class attribute was dropped. (This would also drop any new attributes, although `sento_corpus()` did not add any.) 

The simple solution is to use `rebuild_*()` instead of `build_*()`. My patches work now for corpus and tokens object, but we need to make slightly deeper changes to replace build with rebuild, and to avoid the S4 `new()` calls for modifying a dfm or dictionary2 object.

To reiterate this was how everything worked before #1866. That PR was designed to consolidate the functions that kept the metadata and other characteristics we need for our objects, but it was not intended to make it difficult or impossible to build on our core classes. So we just need to modify that element of #1866 to keep the good part but unto the (probably unintended) effect on using our functions on derived classes.

Let's merge this PR and put the deeper modification on the changelist for 2.1. It's only affecting **sentometrics** for now but this PR solves that.

Issue #1882, opened 2020-02-21
Need to fix Debian test failures


kbenoit:
This is a really tricky one, since CRAN pre-test, *and* `rhub::check_on_debian()`, generate these errors that **do not occur on any other platform**.  Furthermore, they **do not occur on a local Debian 9 system** that I installed and tested this morning.

It **appears** that the differences occur from different tokenizations. I immediately suspected that Debian/cloud had installed a different version of **stringi**, but then I compiled 1.4.6 locally and re-ran it, and also verified that on my version of Debian, it was also using 1.4.6.

I suspect this from the error messages, for instance that the Dimnames have different lengths 13 v 14. Is there some possibility that the tokenizer is operating differently for some functions on Debian, where they are the same on other OSes?

See 
- https://artifacts.r-hub.io/quanteda_2.0.0.tar.gz-8b14dc7ce8004759897b0ebadae9a1ad/quanteda.Rcheck/tests/testthat.Rout.fail
and 
- https://win-builder.r-project.org/incoming_pretest/quanteda_2.0.0_20200220_101246/Debian/00check.log

Also pasted here:

```
Running the tests in ‘tests/testthat.R’ failed.
Complete output:
  > Sys.setenv("R_TESTS" = "")
  > 
  > library(testthat)
  > library(quanteda)
  Package version: 2.0.0
  Parallel computing: 2 of 10 threads used.
  See https://quanteda.io for tutorials and examples.
  
  Attaching package: 'quanteda'
  
  The following object is masked from 'package:utils':
  
      View
  
  > 
  > Sys.setenv("_R_CHECK_LENGTH_1_CONDITION_" = TRUE)
  > 
  > ops <- quanteda_options()
  > quanteda_options(reset = TRUE)
  > 
  > test_check("quanteda")
  ── 1. Failure: bootstrap_dfm works with character and corpus objects (@test-boot
  bs1[[1]] not equal to dfm(corp).
  Lengths: 39, 42
  Attributes: < Component "Dim": Mean relative difference: 0.07692308 >
  Attributes: < Component "Dimnames": Component "features": Lengths (13, 14) differ (string compare on first 13) >
  Attributes: < Component "Dimnames": Component "features": 8 string mismatches >
  Attributes: < Component "i": Numeric: lengths (17, 19) differ >
  Attributes: < Component "p": Numeric: lengths (14, 15) differ >
  Attributes: < Component "x": Numeric: lengths (17, 19) differ >
  
[...]

PR #1883, opened 2020-02-21
Change the handling of separators per ICU 36

koheiw:
#1882 is probably due to the change in how ICU handles separators. Even if there are other issues, we have to be ready for the change anyway.


koheiw:
I did but `rhub::check_on_debian()` looks broken.

PR #1884, opened 2020-02-21
Remove spacyr

koheiw:
I don't think we should include **spacyr** in **quanteda**. `R/spacyr-methods.R` should be in **spacyr**.

The package is too unreliable for its nature (depends on a non-R program that has little transparency). We have to minimize dependency to make maximize **quanteda** reliability.


kbenoit:
Find a way to define **quanteda** functions to **spacyr** objects without making **spacyr** import **quanteda**, and we can discuss this. Otherwise it's doing little harm - the only second-order package dependency it adds is **reticulate**. You don't need to install the underlying spaCy functions unless you want to use **spacyr**, but then of course you would have them anyway. Our tests only check this locally, not on CRAN or CI.

spaCy  installs just fine via `spacyr::spacy_install()` on Windows, although to avoid having to give the long name of the language model, you will need to run the installation as an Administrator. If there are issues with **spacyr**, we should fix them, not disassociate the package. I use this all the time, and so do the other 11k users who install it each month. I don't want to make **quanteda** a requirement for using the **spacyr** functions since it's designed to be more general-purpose than that. (like **stopwords**)

koheiw:
I don't understand why `R/spacyr-methods.R` cannot be defined in **spacyr**... Just extends **quanteda**'s generics, no?

koheiw:
I do not want to see errors that nothing do with **quanteda**
```
── 2. Error: spacy_parse/tokenize work (@test-spacyr-methods.R#48)  ──────────────
spaCy or language model en is not installed in any of python executables.
Backtrace:
 1. spacyr::spacy_initialize()
 2. spacyr:::set_spacy_python_option(...)
```

kbenoit:
Show me one way that adding **spacyr** to Imports reduces reliability. None of our functions rely on **spacyr** functions - our code is just extending functionality. Indeed, you don't even need **spacyr** configured to run anything in **quanteda** except the local test of `spacy_parse()`. What's more, it adds only one new package installation, **reticulate**, and that package's Imports overlap almost entirely with what we already import (**rappdirs** is the only exception - see https://cran.r-project.org/web/packages/reticulate/index.html). Finally, the code is self-contained and very simple, so it adds no burden to maintenance. (I propose something below at the end, but this will increase our maintenance costs by splitting the functions into two packages.)

If you're annoyed that the local tests fail on your system, you should install it correctly or modify the test code. The error you cite above is because you didn't run `spacy_install()` in Administrator mode on Windows, so it could not create the alias "en" to the language model "en_core_web_sm". `spacy_install(model = "en_core_web_sm")` should work (or you can install it the recommended way).

I don't disagree with the principle of wanting to trim dependencies, but think there are better priorities with actual real gains than **spacyr** (which is one we maintain!).

My position is firmly:
- I don't want to make any change that removes any of the integration functionality we currently have with **spacyr**. In fact I think that the model of how we extend **spacyr** through `spacyr-methods.R` is a good example for others as to how to integrate spaCy functionality without needing to coordinate any changes to the **spacyr** package.
- I don't want **spacyr** to need to import **quanteda**. (It's currently in Suggests.)

**However...***

I think we could remove the Imports _and_ keep the functionality but it would require moving 
https://github.com/quanteda/quanteda/blob/master/R/spacyr-methods.R#L38-L46
to **spacyr**.  The other functions methods are **quanteda** methods for **spacyr** objects. This does not require us to import **spacyr**. (just like defining a `corpus.Corpus()` for **tm** objects does not require us to import anything from **tm**). 

Disadvantage: It means we need to maintain functions in two packages instead of one for the same set of functions. (I'm also assuming that this can be made to work without any loss of functionality.) 

Advantage: The one advantage is that for the user is just one thing really, that they can install **quanteda** without a second-order dependency installation of **reticulate** and **rappdirs**. Both of those packages are maintained by RStudio, and are likely to be maintained pretty well.

Is this really worth it?

koheiw:
Before going into details, my question is why we need `quanteda::spacy_tokenize.corpus` when this works perfectly? 
```r
corp <- quanteda::data_char_ukimmig2010
lis <- spacyr::spacy_tokenize(corp)
#> Found 'spacy_condaenv'. spacyr will use this environment
#> successfully initialized (spaCy Version: 2.1.7, language model: en)
#> (python options: type = "condaenv", value = "spacy_condaenv")
toks <- quanteda::as.tokens(lis)
print(toks, 2)
#> Tokens consisting of 9 documents.
#> BNP :
#>  [1] "IMMIGRATION"  ":"            "AN"           "UNPARALLELED" "CRISIS"      
#>  [6] "WHICH"        "ONLY"         "THE"          "BNP"          "CAN"         
#> [11] "SOLVE"        "."           
#> [ ... and 3,333 more ]
#> 
#> Coalition :
#>  [1] "IMMIGRATION" "."           "The"         "Government"  "believes"   
#>  [6] "that"        "immigration" "has"         "enriched"    "our"        
#> [11] "culture"     "and"        
#> [ ... and 256 more ]
#> 
#> [ reached max_ndoc ... 7 more documents ]
```
As you said, methods for `spacyr_parsed` objects are harmless as they are just `data.frame` (it should support the "token exchange format"). I understand that **spacyr** users want to save the output objects and use them only with **quanteda**.

kbenoit:
I am pretty sure that only works because we defined a method for it in **quanteda**, and despite the package not being attached, R is somehow finding that function.

I loaded the 1.5.2 version of a corpus, and then uninstalled **quanteda** entirely. Here's what happens:

```r
> packageVersion("quanteda")
[1] ‘1.5.2’
> search()
 [1] ".GlobalEnv"        "package:spacyr"    "tools:rstudio"     "package:stats"     "package:graphics" 
 [6] "package:grDevices" "package:utils"     "package:datasets"  "package:methods"   "Autoloads"        
[11] "package:base"     
> 
> data("data_corpus_irishbudget2010", package = "quanteda")
> save(data_corpus_irishbudget2010, file = "~/tmp/data_corpus_irishbudget2010.rda")
> remove.packages("quanteda")
```

then try to run `spacy_tokenize()` after clearing the workspace and restarting:

```
Restarting R session...

> load("~/tmp/data_corpus_irishbudget2010.rda")
> library(spacyr)
> spacy_tokenize(data_corpus_irishbudget2010)
Error in UseMethod("spacy_tokenize") : 
  no applicable method for 'spacy_tokenize' applied to an object of class "c('corpus', 'list')"
```

Thinking that we might do better with the v2 corpus class being a special `character`, I loaded the v2 object and tried the same, but no joy:

```r
Restarting R session...

> load("~/Dropbox (Personal)/GitHub/quanteda/quanteda/data/data_corpus_irishbudget2010.rda")
> library(spacyr)
> spacy_tokenize(data_corpus_irishbudget2010)
Error in UseMethod("spacy_tokenize") : 
  no applicable method for 'spacy_tokenize' applied to an object of class "corpus"
```

So: I think that this only works when quanteda is installed. Or, it works if we implement my solution of above of defining the **spacyr** methods for corpus objects in **spacyr**.

kbenoit:
But this works, for the v2 corpus when I added "character" to the class label list, without **quanteda** installed:
```
> class(data_corpus_irishbudget2010) <- c("corpus", "character")
> library(spacyr)
> stoks <- spacy_tokenize(data_corpus_irishbudget2010)
```

any reason for us not to let the `class(corp) <- c("corpus", "character")` ?

koheiw:
My post above was on `no-spacyr`. It should works without `quanteda::spacy_tokenize.corpus()` because the new `corpus` is a `character`.  If our new object does not invoke this method, we need to fix something.
https://github.com/quanteda/spacyr/blob/a0900d90e2924791169befeb2c61dc5148c3f5d7/R/spacy_tokenize.R#L54-L64

kbenoit:
Ah but you didn't test it on a corpus, you tested it on a character: `quanteda::data_char_ukimmig2010` not a corpus. Because we stripped the `character` from the class attribute of a corpus, this fails (and here I've built the `no-spacyr` branch.
```r
> corp <- quanteda::data_corpus_inaugural
> lis <- spacyr::spacy_tokenize(corp)
Error in UseMethod("spacy_tokenize") : 
  no applicable method for 'spacy_tokenize' applied to an object of class "corpus"
```

Can we let a corpus keep its `character` label, as I suggest above, without causing havoc?

koheiw:
I thought the R's underlying class trigger methods, but I was probably wrong. I think it is a good idea to set `c("corpus", "character")` in these object builders.

https://github.com/quanteda/quanteda/blob/36c1ee584e663e155f4e9a27e42645585aeaa44c/R/object-builder.R#L137

https://github.com/quanteda/quanteda/blob/36c1ee584e663e155f4e9a27e42645585aeaa44c/R/object-builder.R#L147

kbenoit:
S3 is pretty dumb.

kbenoit:
Make sure the order is `corpus` first - the `union()` function you have above puts the existing class first.
```r
> txt <- "one"
> union(class(txt), "corpus")
[1] "character" "corpus"  
```

koheiw:
Issue PR #1887. Let's open an new PR for removing `spacy_tokenize.corpus`, because this PR removes to many files.

kbenoit:
I just put them back, except the .corpus methods for the **spacyr** generics. If the other PR works then this will all work and we can merge both, and **quanteda** no longer needs to import **spacyr**.

kbenoit:
The attributes mess things up, but it's easy to fix that by adding an `as.character()` inside the character methods in **spacyr**. I'll do that now.
``` r
library("spacyr")
corp <- quanteda::corpus(c(
  doc1 = "This is Sparta!",
  doc2 = "This is the 2nd document."
))
spacy_tokenize(corp)
## Found 'spacy_condaenv'. spacyr will use this environment
## successfully initialized (spaCy Version: 2.1.8, language model: en)
## (python options: type = "condaenv", value = "spacy_condaenv")
## Error in `[[<-.data.frame`(`*tmp*`, "docname_", value = character(0)): replacement has 0 rows, data has 2
spacy_tokenize(as.character(corp))
## $doc1
## [1] "This"   "is"     "Sparta" "!"     
## 
## $doc2
## [1] "This"     "is"       "the"      "2nd"      "document" "."

spacy_parse(corp)
## Error in `[[<-.data.frame`(`*tmp*`, "docname_", value = character(0)): replacement has 0 rows, data has 2
spacy_parse(as.character(corp))
##    doc_id sentence_id token_id    token    lemma   pos    entity
## 1    doc1           1        1     This     this   DET          
## 2    doc1           1        2       is       be  VERB          
## 3    doc1           1        3   Sparta   Sparta PROPN     GPE_B
## 4    doc1           1        4        !        ! PUNCT          
## 5    doc2           1        1     This     this   DET          
## 6    doc2           1        2       is       be  VERB          
## 7    doc2           1        3      the      the   DET          
## 8    doc2           1        4      2nd      2nd   ADJ ORDINAL_B
## 9    doc2           1        5 document document  NOUN          
## 10   doc2           1        6        .        . PUNCT
```

Issue #1893, opened 2020-02-26
Update and re-organize data objects


kbenoit:
With #1892 solved, we should re-do the corpus objects, but in a separate PR since this could also break some tests.

**quanteda.textmodels** also now has some of the data objects used primarily in textmodels (see https://github.com/quanteda/quanteda.textmodels/pull/18), namely
- `data_corpus_dailnoconf1991`
- `data_corpus_irishbudget2010`
- `data_dfm_lbgexample`

We should remove these from **quanteda** and remove them from examples. (We can keep them for tests by including `Suggests: quanteda.textmodels`.

I would also favour adding a non-political corpus that could contain, for instance, sentiment. Maybe `data_corpus_movies`?

PR #1898, opened 2020-02-27
Remove data_corpus_irishbudget2010, data_corpus_dailnoconf1991

kbenoit:
- replaces their use in examples with data_corpus_inaugural
- gets the objects from **quanteda.textmodels** in the tests, where used, or replaces them with other data objects
- adds a documentation entry noting that these have been moved, with aliases to the old object names.  This entry now points to the **quanteda.textmodels** package.

Keeps `data_dfm_lbgexample` since this is used a lot in `textstat_*()` examples.

Solves #1893, at least on the **quanteda** package side.  (They still need updating on the **quanteda.textmodels** side.)


kbenoit:
I wonder if we should merge this to a 2.1 branch for now, or just drop the bomb now and get people used to these being in the other package. I will try a revdep check before merging this to master. It's unlikely we break anything but better to be sure.

Issue #1901, opened 2020-02-27
R dev 4.0 breaks on our use of data.frames, needs fixing for CRAN


kbenoit:
Got this today from CRAN:

> Please see the problems shown on
> <https://cran.r-project.org/web/checks/check_results_quanteda.html>.
> 
> Specifically, see the problems shown for the r-devel Debian checks.
> 
> These can be reproduced by checking with --as-cran using a very current
> r-devel (r77865 or later), which makes data.frame() and read.table() use
> a stringsAsFactors = FALSE default, which is planned to become the new
> default for the upcoming R 4.0.0.
> 
> Please see
> <https://developer.r-project.org/Blog/public/2020/02/16/stringsasfactors/index.html>
> for more information about this change.
> 
> Can you please fix your package to work with both the old and new
> default?  In principle, this can easily be achieved by adding
> stringsAsFactors = TRUE to the relevant calls to data.frame() or
> read.table() [or other read.* function calling read.table()], but please
> only do this if the sort order used in the string to factor conversion
> really does not matter (see the blog post about the locale dependence of
> the conversion).  Otherwise, please change to create the factors with
> explicitly given levels.
> 
> The new problems may be from code in a package you depend on: in this
> case, please let me know, and get in touch with the maintainer of that
> package.
> 
> Please correct before 2020-03-20 to safely retain your package on CRAN.

kbenoit:
I just installed the very latest dev version of R in a docker container running Debian, and could not reproduce the error.
```r
> sessionInfo()
R version 3.6.3 Patched (2020-02-29 r77878)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Debian GNU/Linux bullseye/sid
> test_file("tests/testthat/test-convert.R")
✔ |  OK F W S | Context
✔ |  60       | test convert function and shortcuts [0.9 s]

══ Results ═════════════════════════════════════════════════════════════════════
Duration: 0.9 s

OK:       60
Failed:   0
Warnings: 0
Skipped:  0
> test_file("tests/testthat/test-tokens_segment.R")
✔ |  OK F W S | Context
✔ |  32       | test tokens_segment [0.2 s]

══ Results ═════════════════════════════════════════════════════════════════════
Duration: 0.2 s

OK:       32
Failed:   0
Warnings: 0
Skipped:  0
```

kbenoit:
Same under the exact version reporting the error:
```r
[...]

PR #1904, opened 2020-03-02
Fix test errors for r-devel

kbenoit:
These are caused by a change to the default in data.frame in the next major R release, in which `stringsAsFactors = FALSE` by default.

This fixes the tests by forcing factors, but also cleans up something else that was happening with `dfm2lsa()`. I think this was **lsa**'s problem, not ours. In any case, I fixed it by using the `lsa::as.textmatrix()` converter instead of re-engineering it. We could avoid the conversion to dense matrix and back by re-engineering the conversion ourselves, but I will leave this for future work.

Solves #1901.

### DIFF
--- a/R/convert.R
+++ b/R/convert.R
@@ -249,9 +249,9 @@ dfm2austin <- function(x) {
 dfm2tm <- function(x, weighting = tm::weightTf) {
     attrs <- attributes(x)
     if (!requireNamespace("tm", quietly = TRUE))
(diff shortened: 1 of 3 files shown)


kbenoit:
Failing codecov because it's really impossible to test using CI what it's like when the packages are not installed (short of uninstalling them in the test code). We could put this at the end of the test file, or just live with it.

Issue #1017, opened 2017-10-12
Make ngram wrapper for patterns?


koheiw:
It seems that users face problems when they work with ngrams: https://stackoverflow.com/questions/46685498/remove-ngrams-with-leading-and-trailing-stopwords

Then, how about making a `ngram` wrapper similar to `phrase` that basically does this for users:
```r
pattern = c(paste0("^", stopwords("english"), "_"),  paste0("_", stopwords("english"), "$"))
```
We have something similar as an internal function for `dfm_select` only, but it can be exported so that users can user on tokens too. 
```r
## convert patterns (remove and select) to ngram regular expressions
make_ngram_pattern <- function(features, valuetype, concatenator) {
    if (valuetype == "glob") {
        features <- stri_replace_all_regex(features, "\\*", ".*")
        features <- stri_replace_all_regex(features, "\\?", ".{1}")
    }
    features <- paste0("(\\b|(\\w+", concatenator, ")+)", 
                       features, "(\\b|(", concatenator, "\\w+)+)")
    features
}
```

kbenoit:
I'm thinking a way to implement this would be through some modifier to `pattern()`, where we have a "mask" of logicals corresponding to each ngram. So in the above example, we would have

```r
dfm_remove(x, pattern = phrase(stopwords("english"), mask = c(TRUE, FALSE, TRUE)))
```

or something like that. Since we store `concatenator` in the attributes of `x`, we do not need the user to supply the character used to demarcate the individual "grams" of the ngram tokens or features.

This would also be very useful in filtering collocations.

koheiw:
I like the mask idea, but we need to generalize it a bit more to allow selection of ngrams and collocations with different length. I will also think about it.

PR #1906, opened 2020-03-03
Increase codecov

kbenoit:
I decided to fill in some of the gaps in our code coverage. Not surprisingly, this revealed a few inconsistencies, which I fixed. Example: the verbose report for `dfm.tokens(x, stem = TRUE)`, which unlike the same option for `dfm.dfm()`, did not take the default dictionary from quanteda_options.

I also removed (or commented out) some functions that were untested, because they are no longer used.

### DIFF
--- a/R/dfm.R
+++ b/R/dfm.R
@@ -123,7 +123,6 @@ dfm <- function(x,
 
     dfm_env$START_TIME <- proc.time()
     object_class <- class(x)[1]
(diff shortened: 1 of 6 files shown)


kbenoit:
Well, it doesn't really solve #1017, since the function I commented out was never actually called by the pre-existing code. I'm happy without it though so we can close #1017 if you agree.

PR #1913, opened 2020-03-12
Obtaining very large values from quanta

Ahmedmohammed88:
Dear Professor Kenneth,
I hope you are fine.
As per your discussion with my supervisor via email, I used quanteda to obtain the fog index for some statements in the annual reports of the FTSE 350 such as the chairman statement, CEO … etc. The thing which is surprising is that I obtained very large values of the readability measure (the fog index). Most values are greater than 100 with an average of 400. Usually, the average should be around 17-18. What would be the reason?
Thank you very much,
I appreciate your help and support.


kbenoit:
@Ahmedmohammed88 Thanks for pointing this out. Can you file this an issue? This is a pull request (and not at all clear from what source).

We have an issue template an instructions. Can you provide an example and perhaps a link to a "correct" computation of FOG for the same example for which `textstat_readability(x, measure = "FOG")` produces the wrong answer?

PR #1924, opened 2020-04-09
Add dictionary_edit() + related fns

kbenoit:
Adds `dictionary_edit()` for interactive editing of dictionaries.

Also exports supporting functions `list_edit()` and `char_edit()`, to solve https://github.com/quanteda/stopwords/issues/14.

**Note to @koheiw** 
I encountered an inconsistency here in that subsetting a dictionary obliterated its user metadata. (This is not how subsetting works for corpus or tokens objects.) I fixed this in https://github.com/quanteda/quanteda/pull/1924/files#diff-0b36c445bf82a97452e429faaf1b445dR415-R417 but you might prefer to rewrite it in the framework of the object-builder.R functions (which I admit I find hard to follow).


kbenoit:
Note: I will add tests once the additions are approved.

koheiw:
It errors with an old dictionary object. You should upgrade the input dictionary using `as.dictionary.dictionary2`. 
```r
> out <- dictionary_edit(newsmap::data_dictionary_newsmap_en)
 Error in `meta<-`(`*tmp*`, value = x_meta) : value must be a named list 
```
Also you should use `rebuild_dictionary2()` or `build_dictionary2()`. We are no longer allowed to DIY objects since v2.0.

koheiw:
I wonder why plain text is highlighted like R code. Is there any way to fix this?

Screenshot_20200412_042629

koheiw:
Edit window has problem with Unicode characters on Windows
```r
> out <- char_edit("あいうえお")
> out
[1] "<U+3042><U+3044><U+3046><U+3048><U+304A>"
> Encoding(out)
[1] "unknown"
```
Annotation 2020-04-12 044221

koheiw:
On Linux, the length of the vector changes from 1 to 2, probably because of the final line.
```
> out <- char_edit("abcd")
> length(out)
[1] 2

> out <- char_edit("あいうえお")
> Encoding(out)
[1] "unknown" "unknown"
> length(out)
[1] 2
```

kbenoit:
Works nicely now, except:

- there is a last blank line in the editor when `char_edit()` is called. There is also one when  `dictionary_edit()` is called but `read_yaml()` removes that. A simple solution then would be to convert a vector to a list, and then edit it as unnamed yaml (with the `-` in front of each element). Or, to remove the newline from the end of the file before `edit()` imports it. Or, just ignore this problem.

- I am not using the build/rebuild functions in `dictionary_edit()`, @koheiw can you change that code in the "allowed" way? 😀

koheiw:
build/rebuild must be used, because this the only way to guarantee that objects have all the necessary attributes/slots without S4. `meta()` does not work because we do not allow users to set system meta. attri(x, "meta") works but it does not go thorough necessary checks.

kbenoit:
fine with me, can you make the edit?

R/dictionary_edit.R
```diff
@@ -0,0 +1,70 @@
...
+#' @examples
+#' \dontrun{
+#' my_stopwords <- stopwords("en", source = "snowball") %>%
+#'     char_edit()
+#' }
+char_edit <- function(x, sep = "\n") {
+    if (!is.vector(x) || !is.character(x))
+        stop("x must be a character vector")
+    tfile <- tempfile(fileext = ".txt")
+    fwrite(as.list(x), file = tfile, sep = sep, col.names = FALSE, quote = FALSE)
```
koheiw, line 61: If I change this to `writeLines(x, con = tfile)`, no empty element is inserted. Please try.

koheiw:
```r
> out <- char_edit("abcd")
> length(out)
[1] 1
```

kbenoit:
I did some more exploring. The problem is not in the function to write the vector, it's in the `edit()` function, which automatically adds a `\n` to the end of the file when one did not exist.

``` r
txt <- LETTERS[1:3]
writeLines(txt, con = "~/tmp/text_writeLines.txt")
system("wc -l ~/tmp/text_writeLines.txt")
```
Only three lines, but then
```r
new_txt <- edit(file = "/Users/kbenoit/tmp/text_writeLines.txt")
```
produces
Screenshot 2020-04-13 09 06 01
which after the return, has a fourth element. (But the function currently removes this.)

It's an issue with the editor though, since other editors like calling `vi()` or `emacs()` - which I have to do from a terminal R, not RStudio - does _not_ add the end-of-file newline.

Just something we have to live with I guess.

koheiw:
I thought it is a problem of edit() initially, but it is correct to have an empty line in the end and `readLines()` even complains without it. The problem was in `fwrite()` that appends LF in the last line. The last line looked empty but it wasn't earlier.

koheiw:
I think you are right. `edit()`also has the problem. We probably should not use the function to edit texts.
 
Screenshot_20200413_124702

kbenoit:
But it still reads back ok.

That could be platform-dependent, though, as in `stringi::stri_write_lines()`, there is this argument:
```r
sep = ifelse(.Platform$OS.type == "windows", "\r\n", "\n")
```

We could pass through the editor argument from `edit()` - it works ok in other editors.

koheiw:
Seems that some elements will be removed...

Windows

Annotation 2020-04-14 123946

koheiw:
I think it is fine to add `char_edit()` as an experimental function, but we need to ask the RStuido to fix the bug, or create a RStudio Addin.

PR #1925, opened 2020-04-10
Add char_select() functions

kbenoit:
Adds:
- `char_select()`
```r
char_select(x,
  pattern,
  selection = c("keep", "remove"),
  valuetype = c("glob", "fixed", "regex"),
  case_insensitive = TRUE
)
```
- `char_remove()`
- `char_keep()`

Could be a nice addition to include the additional arguments from `tokens_select()`, namely:
```r
  padding = FALSE,
  window = 0,
  min_nchar = NULL,
  max_nchar = NULL,
  startpos = 1L,
  endpos = -1L
```
but not essential imho


kbenoit:
Interesting that this now works for removing documents from a corpus object (but not individual elements). Could be useful for cleaning up segmented parts such as short fragments of text that show up as documents, pdf footers, things that are not really sentences after sentence segmentation, etc.

``` r
library("quanteda")
## Package version: 2.0.2
## 
## 
##     View
corp <- corpus(letters[1:5],
  meta = list(test = "Example corpus.")
)
temp <- char_remove(corp, c("b", "b"))
meta(temp)
## $test
## [1] "Example corpus."
```

kbenoit:
That character string transformation differences is a different issue, but worth thinking about. We do some regularization of corpus inputs in `corpus()` and one of those is to make all hyphen-like characters into a simple hyphen, but that was before we started using Unicode categories for hyphens.

PR #1928, opened 2020-04-15
Don't require tests in packages only in Suggests

### LINKED ISSUE #1927: Skip tests with Suggests only packages
Some tests don't execute when they depend on packages that are not in Imports, but only in Suggests. We skip some of these but not all. We should wrap them in conditionals or add `skip_if_not_installed()` to the test blocks.

```
* checking tests ... [127s/184s] ERROR
 Running ‘spelling.R’
 Running ‘testthat.R’ [127s/182s]
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:

 ══ testthat results  ═══════════════════════════════════════════════════════════
 [ OK: 2356 | SKIPPED: 24 | WARNINGS: 0 | FAILED: 9 ]
 1. Error: (unknown) (@test-corpus.R#3)
 2. Error: (unknown) (@test-dfm.R#3)
 3. Error: docnames are alwyas unique (@test-docnames.R#73)
 4. Error: (unknown) (@test-docvars.R#3)
 5. Error: fcm print works as expected (@test-fcm.R#211)
 6. Error: textplot_network error when fcm is too large (@test-textplot_network.R#31)
 7. Error: tokens printing works (@test-tokens.R#885)
 8. Error: generate_groups works for tokens objects (@test-tokens_group.R#75)
 9. Error: generate_groups works for corpus objects (@test-tokens_group.R#100)

 Error: testthat unit tests failed
 Execution halted
* checking for unstated dependencies in vignettes ... OK
```

kbenoit:
When we moved the textmodel functions to the new package **quanteda.textmodels**, we did not skip the tests that required functions from that package. This PR fixes that. Since most of the tests affected only needed the `data_corpus_irishbudget2010` object that has been now moved to the other package, I put it in a the `tests/data` directory for use in the tests only. (Another alternative would be to change the tests to use a different object.)

The vignette however still needs **quanteda.textmodels** to be installed. @eddelbuettel does that cause revdepcheck problems or can we leave that in the vignette? Thanks!

**Added 2020-04-18**

I changed all of the tests to use a subset of the inaugural corpus. Continuing to use the data_corpus_irishbudget2010 corpus in tests was just an expedient, after we moved that object to **quanteda.textmodels**.  There was nothing special about using that corpus, so I changed all the tests to use only the built-in corpus object(s). So it's now removed from the `tests/data` directory.

Also made its use in the `quickstart.Rmd` vignette conditional on it being installed (as we did before with **stm**).

Closes #1927


eddelbuettel:
I tend to skip vignettes as the rev.dep already takes me up to 18h.  

I also "folded" for a few packages and grudgingly installed the Suggests.  I can similarly install `quanteda.textmodels`.  I just sent you that email because I thought you'd be receptive to the "engineering" aspect of it (as per my reading of _Writing R Extensions_; some very vocal people take a different view :man_shrugging: ).  So good to see the PR :)

eddelbuettel:
Thumbs up -- just tested this 2.0.2 candidate. Ended up installing `quanteda.textmodels` to satisfy `R CMD build`.  The new tarball passes as one would hope -- nice.

(And others still fail.  E.g. for `textrecipes` I installed the missing Depends `modeldata` yet it still bombed. Only once Suggests `textfeatures` was installed did it pass. That is meh.)

PR #1935, opened 2020-04-22
Replace NNBSP with regular space

### LINKED ISSUE #1934: Whitespace error for stemming when constructing a dfm
I'm tokenizing data from Tweets and run into an error removing whitespace, where there is presumably a special character that looks like a space that isn't being removed. In the first case, the "space" is treated as if it is a normal character and thus doesn't get split. In the second case, I removed the "space" and put a normal space and it separates as expected. Hopefully this will come through with a copy and paste on GitHub. Although the two commands look identical, the strings are not.

```
> tokens("in @usaidffp")
tokens from 1 document.
text1 :
[1] "in @usaidffp"

> tokens("in @usaidffp")
tokens from 1 document.
text1 :
[1] "in"        "@usaidffp"

> "in @usaidffp" == "in @usaidffp"
[1] FALSE
```

kbenoit:
Replaces NNBSP with a regular space when a corpus is constructed.

Fixes #1934.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,9 @@
 * Added new methods `char_select()`, `char_keep()`, and `char_remove()` for easy manipulation of character vectors.
 * Added `dictionary_edit()` for easy, interactive editing of dictionaries, plus the functions `char_edit()` and `list_edit()` for editing character and list of character objects.
 
+## Bug fixes
+
+* Fixed a problem caused by Narrow No-Break Space (NNBSP) not being recognized as an intra-token space. (#1934)
 
 # quanteda 2.0.1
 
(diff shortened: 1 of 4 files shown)


kbenoit:
That's more comprehensive, but it's also more invasive. **stringi** handles all of the other space variants fine, and while it should also handle the NNBSP, for some reason does not. (I filed an issue at https://github.com/gagolews/stringi/issues/377 however). There might be a reason for some users to have other space variants - do we really want to sanitize it so thoroughly that a corpus cannot contain those? The answer could well be, Yes! but that's a substantive change. My principle on this fix is, just address the immediate problem, but note the bigger issue for a possibly more comprehensive fix later.

The addition to the vectorized replacement as a fixed replacement is the most efficient method method too, since it involves just adding an additional replacement to the (single) fast fixed replacement function call.

kbenoit:
By the way we definitely do not want to 
```r
stringi::stri_replace_all_regex(x, "[\\p{White_Space}]", " ")
```

since that obliterates newlines!

koheiw:
We refrain from invasive changes in texts but worth considering raising warnings for non-breaking spaces.

PR #1493, opened 2018-11-17
Update cheatsheet

stefan-mueller:
- dfm_compress() occured twice in the previous version, I remove one of them
- change "quanteda.data" to "quantda.corpora"
- change date of last update to 11/18 
- save file as PDF

@kbenoit: If you are happy with the changes, I can also send the updated version to RStudio.

PR #1947, opened 2020-05-11
Update cheatsheet

stefan-mueller:
Addresses #1493 

Changes: 
- Revise extensions packages (add **quanteda.textmodels**)
- Update output of `summay(data_corpus_inaugural, n = 2)`
- Update dfm output of `head(x, n = 2, nf = 4)` 
- Specify `pattern` when in `kwic()`
- Adjust font in `dfm_weight()`
- Add note that `textmodel_()` functions require the **quanteda.textmodels** package
- Add `textmodel_svm()`
- Use @kbenoit's and @stefan-mueller's quanteda.org email addresses
- Update date


stefan-mueller:
Thanks, @kbenoit. I will incorporate your suggested changes.
Regarding listing one option or more than one options: other cheat sheets (for instance the [**stringr** cheat sheet](https://github.com/rstudio/cheatsheets/raw/master/strings.pdf)) use a combination of both. I would suggest we present more than one option when the output or logic are fundamentally different. For instance, I am thinking about adding: `tokens_select(x, c("powerful", "text"), selection = ("keep", "remove")` instead of only specifying `selection = "keep"`. 

Yet, since our package documentation is quite extensive and comprehensible anyway, we should not make it too complex and stick to one option for most functions. The aim of the cheat sheet is to provide an overview of the functions. I would recommend we stick to the current structure, I will check whether other functions should present more than one option, and mainly limit the edits to the issues you identified above. Do you agree?

kbenoit:
How about:
* If it's generic, we show options
* If specific - in the sense of supplying specific values as in `tokens_select(x, c("powerful", "text"), ...` then we list just one option.

stefan-mueller:
I have made the requested changes. Please have a look at the updated PDF file and let me know what you think, @kbenoit.

Issue #1314, opened 2018-04-13
Weighted dictionary solution


kbenoit:
Following the discussion in #1263 and #1271, and discussions, here is the proposal for recording and using dictionary weights.

### Add a new `dictionaryweights()` constructor function

This will have the signature and functionality as `dictionary()`:
```r
dictionaryweights(x, file = NULL, format = NULL, separator = " ",
  tolower = TRUE, encoding = "auto")
```

It will read a file or take a list of named numeric weights, where the list element name corresponds to the dictionary key, and the numeric element name corresponds to the dictionary value. Example:
```r
dictionaryweights(list(sentiment1 = c(zealots = -.8, zzz = -1.2), sentiment2 = c(zzz = 5)))
```
If a file, it would work on Wordstat formatted dictionaries only (since only they take weights, among the formats we currently support).

We would also create a coercion function `as.dictionaryweights()` for data.frame objects such as these:
```r
> head(tidytext::sentiments[tidytext::sentiments$lexicon == "AFINN", ], 20)
# A tibble: 20 x 4
   word       sentiment lexicon score
   <chr>      <chr>     <chr>   <int>
 1 abandon    NA        AFINN      -2
 2 abandoned  NA        AFINN      -2
 3 abandons   NA        AFINN      -2
 4 abducted   NA        AFINN      -2
 5 abduction  NA        AFINN      -2
 6 abductions NA        AFINN      -2
 7 abhor      NA        AFINN      -3
 8 abhorred   NA        AFINN      -3
 9 abhorrent  NA        AFINN      -3
10 abhors     NA        AFINN      -3
11 abilities  NA        AFINN       2
12 ability    NA        AFINN       2
13 aboard     NA        AFINN       1
14 absentee   NA        AFINN      -1
15 absentees  NA        AFINN      -1
16 absolve    NA        AFINN       2
17 absolved   NA        AFINN       2
18 absolves   NA        AFINN       2
19 absolving  NA        AFINN       2
20 absorbed   NA        AFINN       1
> lexicon::hash_sentiment_vadar
              x    y
   1:       aas  2.5
   2:      aayf  2.7
   3:   abandon -1.9
   4: abandoned -2.0
   5: abandoner -1.9
  ---               
7232:      ywia  2.5
7233:    zealot -1.9
7234:   zealots -0.8
7235:   zealous  0.5
7236:       zzz -1.2
```

This would have the signature:
```r
[...]

PR #1948, opened 2020-05-11
Add textstat_sentiment()

kbenoit:
This is far from complete, but I'll make it a PR so we can avail of the review capabilities for discussion.

The idea here is:
* to provide a standardised sentiment statistic that applies one of several built-in functions, or lets the user supply their own function, for maximum flexibility and extensibility;
* right now, that this works with required fields of "positive" and "negative", and optionally "neutral" and "other", although those may not be used in the function to compute sentiment;
* to require the user to have applied a dictionary before this function, to maximise flexibility with respect to how the user wants to process the tokens and/or apply the dictionary;
* work for both tokens and dfm objects.

It uses the name "sentiment" to make it very clear what this function is for, when used by 75% of our users; the other 25% can use "pos" and "neg" as "right" and "left" or whatever other bi-polar scales they need.

Let's get the design right, and then think about ways to compute sentiment from continuously valued or weighted scales (rather than just from category counts).


R/textstat_sentiment.R
```diff
@@ -0,0 +1,136 @@
...
+}
+
+#' @export
+textstat_sentiment.dfm <- function(x, pos = "positive", neg = "negative",
+                                   neut = NULL, fun = sent_logit, ...) {
+    if (!all(c(pos, neg, neut) %in% featnames(x)))
+        stop("pos, neg, and neut must match features in x")
+
+    # get the categories that are not sentiment
+    x_other <- dfm_remove(x, c(pos, neg, neut))
```
koheiw, line 67: There is `nomatch` argument in `dfm_lookup()`. I don't remember why we have it, but it might be useful here (if not we should making it useful).

kbenoit:
Totally agree. This is designed so that this the lookup has to be done before invoking this function. The encouraged use here involves first applying `tokens_lookup()` (although supplying a dfm is also possible). 

See the examples in `?textstat_sentiment`.

R/textstat_sentiment.R
```diff
@@ -0,0 +1,136 @@
...
+#' @return a data.frame of sentiment scores
+#' @export
+#' @references  Lowe, W., Benoit, K. R., Mikhaylov, S., & Laver, M. (2011).
+#'   Scaling Policy Preferences from Coded Political Texts. _Legislative Studies
+#'   Quarterly_, 36(1), 123–155.
+#'   <http://doi.org/10.1111/j.1939-9162.2010.00006.x>
+#' @examples
+#' toks <- tail(data_corpus_inaugural, n = 5) %>%
+#'     tokens(remove_punct = TRUE, remove_numbers = TRUE) %>%
+#'     tokens_lookup(dictionary = data_dictionary_LSD2015, nomatch = "_other_")
```
kbenoit, line 28: here's the `nomatch` argument

R/textstat_sentiment.R
kbenoit: Already used it - see the comment I just added. The user has to use the lookup function before calling `textstat_sentiment()`. In this design (which is just one possible design), the user just supplies the list of types (or features)  to be associated with the positive and negative categories. These types were created from dictionary keys by `tokens_lookup()`. If the user used `nomatch = "something"` then that will be counted by a `fun` that needs an "other" count.

R/textstat_sentiment.R
koheiw: Why don't you include `tokens_lookup()` this in the new function, so that input objects will be only `toks` and `data_dictionary_xxx`?

kbenoit: Actually I started it that way. But we would still need the fields to identify the pos, neg, neutral keys or features.

If the signature were:
```r
textstat_sentiment(x, 
                   dictionary = list(pos = "positive", neg = "negative", neut = NULL),
                   fun = sent_logit)
```
then that would pick up types or features as before, with the option of overwriting the value field, but also work on texts to which dictionaries had not yet been applied.

koheiw:
I had similar problem in my package **LSS**:
```r
require(LSS)
> data_dictionary_sentiment
Dictionary object with 2 key entries.
- [positive]:
  - good, nice, excellent, positive, fortunate, correct, superior
- [negative]:
  - bad, nasty, poor, negative, unfortunate, wrong, inferior
> as.seedwords(data_dictionary_sentiment)
       good        nice   excellent    positive   fortunate     correct    superior         bad       nasty 
          1           1           1           1           1           1           1          -1          -1 
       poor    negative unfortunate       wrong    inferior 
         -1          -1          -1          -1          -1 
```
The functions is just `as.seedwords(x, upper = 1, lower = 2)`.

The fundamental problem here is that dictionaries do not have information about their polarity, but we save it in the object meta field.

R/textstat_sentiment.R
kbenoit: @koheiw Should I implement a version of this and we can see how it looks?

koheiw: How about adding `polarity = c(1, -1)` to dictionaries (as in my comment below), so that sentient analysis can be performed only with tokens and a dictionary?

koheiw: It is just to add a `polarity` or `weight` argument to `dictionary()` and saved in object meta. Users can also save that to using `meta()` in user meta.

kbenoit: We can but we also need an argument to override this, since some dictionaries will have multiple possible poles. (LIWC, ANEW, etc). So it still belongs in the argument, but if omitted, we can take it from the dictionary setting.

koheiw: I see. Then, let's make the argument values similar to `c("pos" = 1, "neg" = -1)`. Not only for  consistency but for the flexibility like `c("very_pos" = 2, "pos" = 1, "neg" = -1, "very_neg" = -2)`. The name of the elements should match the dictionary keys of course. The argument default to `NULL`.

kbenoit: I think we should leave any weights to the function side, for complete flexibility. See my notes below on the rewrite I just pushed.

kbenoit:
### Notes on latest revision and redesign

1. Now works in one step:

```r
> textstat_sentiment(tail(data_corpus_inaugural), dictionary = data_dictionary_LSD2015)
        doc_id sentiment
1 1997-Clinton 0.7169984
2    2001-Bush 0.8581440
3    2005-Bush 0.8853072
4   2009-Obama 0.4787824
5   2013-Obama 0.7872986
6   2017-Trump 0.9635760
```

2. Works on character, corpus, tokens, and dfm objects, although we should discourage users from inputting dfm objects since this will make phrase matching impossible.

3. Requires a new `polarity()` setting in the dictionary's object metadata, set by a new function `polarity()`. This allows for positive, negative, and neutral categories.

4. Computation of sentiment is always in terms of some combination (and usually, subset) of `pos`, `neg`, `neut`, and `other`, supplied by the argument `fun`. There are three functions now, but the example in `?textstat_sentiment` shows how to create a user supplied function.

### Things to do

- [ ] Add tests
- [ ] Add error checking for `fun`
- [ ] Think of how we will be able to extend this to allow functions for compiling dictionary keys whose values have continuous sentiment scores, such as when we adapt dictionary objects to allow values, for handling (e.g.) the AFINN dictionary:
```r
> tidytext::get_sentiments(lexicon = c("afinn"))
# A tibble: 2,477 x 2
   word       value
   <chr>      <dbl>
 1 abandon       -2
 2 abandoned     -2
 3 abandons      -2
 4 abducted      -2
 5 abduction     -2
 6 abductions    -2
 7 abhor         -3
 8 abhorred      -3
 9 abhorrent     -3
10 abhors        -3
# … with 2,467 more rows
```
- [ ] @koheiw you might want to look at how I operate on the bare-metal meta slots in `polarity<-()`, since I find the new object creation/rebuilding/meta stuff complicated to follow...

R/textstat_sentiment.R
kbenoit: See ?`sentiment-functions`

koheiw:
I think it is much streamlined with the `polarity()`function, but I still think polarity should be a named numeric vector, because it is more flexible. Taking AFINN as an example, we can make a dictionary with using the discreet sentiment scores as keys:

```r
> afinn <- tidytext::get_sentiments(lexicon = c("afinn"))
> lis <- split(afinn$word, afinn$value)
> dict <- dictionary(lis)
> dict
Dictionary object with 11 key entries.
- [-5]:
  - bastard, bastards, bitch, bitches, cock, cocksucker, cocksuckers, cunt, motherfucker, motherfucking, niggas, nigger, prick, slut, son-of-a-bitch, twat
- [-4]:
  - ass, assfucking, asshole, bullshit, catastrophic, damn, damned, damnit, dick, dickhead, fraud, frauds, fraudster, fraudsters, fraudulence, fraudulent, fuck, fucked, fucker, fuckers [ ... and 23 more ]
- [-3]:
  - abhor, abhorred, abhorrent, abhors, abuse, abused, abuses, abusive, acrimonious, agonise, agonised, agonises, agonising, agonize, agonized, agonizes, agonizing, anger, angers, angry [ ... and 244 more ]
- [-2]:
  - abandon, abandoned, abandons, abducted, abduction, abductions, accident, accidental, accidentally, accidents, accusation, accusations, accuse, accused, accuses, accusing, ache, aching, admonish, admonished [ ... and 946 more ]
- [-1]:
  - absentee, absentees, admit, admits, admitted, affected, afflicted, affronted, alas, alert, ambivalent, anti, apologise, apologised, apologises, apologising, apologize, apologized, apologizes, apologizing [ ... and 289 more ]
- [0]:
  - some kind
[ reached max_nkey ... 5 more keys ]
```
Then assign polarity scores with a named numeric vector:
```r
sent <- structure(as.numeric(names(dict)), names = names(dict))
polarity(dict) <- sent # dose not work currently
```
Internally, it is extremely fast and easy to compute document's polarity scores with a numeric vector:
```r
> dfmat_dict <- dfm_lookup(dfmat, dict)
> dfmat_dict %*% sent / rowSums(dfmat)
5 x 1 Matrix of class "dgeMatrix"
                 [,1]
2001-Bush  0.05537099
2005-Bush  0.06228374
2009-Obama 0.02454444
2013-Obama 0.03884333
2017-Trump 0.07289157
```
This also makes the tricky `fun` argument unnecessary, because polarity scores works as an equation. Users might want to choose either how to normalize scores, so substitute `fun` with something like `denominator = c("all", "matched", "none")`.
```r
# all
dfmat_dict %*% sent / rowSums(dfmat)
# matched
dfmat_dict %*% sent / rowSums(dfmat_dict)
# none
dfmat_dict %*% sent
```
I have a few suggestions on handling of meta field, but I can takeover that part.

kbenoit:
OK, then two questions about this approach. But first let me outline what I am thinking distinguishes two different types of dictionaries, that need different handling.

#### 1. Polarity dictionaries.
 
These are dictionaries like LSD2015 (or LIWC or MFD) that have keys whose values have identical associations with a polarity of sentiment, which we can think of as positive, negative, or (possibly) neutral. For dictionaries such as LSD2015, there are two positive and two negative categories. For dictionaries such as the LIWC, with many categories, we would need to select the poles for a particular application.

A sentiment scoring consists of some formula built on `p`, `n`, `u` (for neutral), and `N` (total counts). There are only 3-4 common ways to do this, so I thought we could create packages of these but also allow the user to extend them. I admit this could be simplified in the current PR, perhaps with just those single letters, to work a bit like the distance and similarity formula descriptions in **proxy** in terms of `a`, `b`, etc.

Polarity dictionaries do not need to be any different from our existing dictionaries, except for an added setting that identifies the keys associated with the specific poles. This allows existing dictionaries to be used for polarity-based sentiment computation with no breaking changes to any existing use. It also allows for a single dictionary object to contain many different keys, for instance LIWC or MFD that contain multiple psychological or affect categories that the user can select from for a particular sentiment computation.

2. Valence dictionaries

For some approaches as in ANEW, NRC, AFINN, etc we will have words that have different continuously valued scores along a continuum. Unlike most polarity dictionaries, the same words will have multiple values, with membership in multiple keys (but this is not at all necessary). ANEW for instance is one dictionary, but with three keys (pleasure, arousal, dominance). 
```r
head(anew)
##        Word Wdnum ValMn ValSD AroMn AroSD DomMn DomSD
## 1 abduction   621  2.76  2.06  5.53  2.43  3.49  2.38
## 2      able  1041  6.74  2.00  4.30  2.17  6.83  2.04
## 3  abortion   622  3.50  2.30  5.39  2.80  4.59  2.54
## 4    absent  1042  3.69  1.72  4.73  1.76  4.35  1.87
## 5    absurd   623  4.26  1.82  4.36  2.20  4.73  1.72
## 6 abundance   624  6.59  2.01  5.51  2.63  5.80  2.16
```

Right away we see that we need a different structure from the polarity approach, since a key does not have the same "polarity" for all of its values, and since the same value will have different valence scores depending on which key it's a member of.  This is why I consider this to require a special additional layer of information in a "dictionary3" type extended object. (See https://github.com/quanteda/quanteda/issues/1314#issuecomment-629776052 for ideas.)

But what about "hybrid" models such as your example above with AFINN where a single list of words have a discrete score, and you treated the discrete score as a "key" and split the word membership across those keys? In my view this should be treated as a valence dictionary, not a polarity dictionary.

Note that it's not a given that we would score a new text by the weighted mean of the valence scores of its words, although that's clearly the most obvious choice. VADER includes "valence shifters" where a negation can shift the score of its target word by arbitrary factors such as 1.8, and this is implemented in the **sentimentr** package. It would be nice to have a design that allows for such flexibility as well. (Why should the other packages have all the fun??)

A valence dictionary should also allow us to use the dictionary for more classic dictionary functions. So if we want to use a dictionary as a `pattern` argument, we should able to use the same approach that works now, for instance `tokens_select(toks, pattern = afinn["arousal"])`, which ignores the valence scores but takes into account the values that are members of that key. 

So in terms of my **fundamental principles** in guiding the extensions:
1. We should not redefine what keys and values mean just to accommodate polarities or valences. 
2. We should make ease of use a key feature of `textstat_sentiment()` (as the PR achieves) but also allow power users to customise their sentiment functions.

#### Problems with the "polarity/valence-as-key" approach

Two issues I see here are:
* Word memberships in keys as we use them now overlap and have different values according to their key, as in the AFINN example above. We would be violating the concept of a "dictionary" to say that one object can only refer to one key at a time, and that the key values would be a highly specific set of values such as 2.76, 2.00, 2.30, etc that might just contain one or two values each. 
* The aggregation function of scoring words by their numeric "weights" assumes a very specific form of sentiment aggregation. How would we implement the logit scale then? I agree we could implement different normalizations via other options, and we could require a user to log the counts first and then apply the -1, +1 to subtract the logged polarities, in effect giving them log(p/n), but then we've required about as much manual work as could be done now using `tokens_lookup() %>% dfm()` and then manipulating the dfm.

Both issues violate the fundamental principles I'd prefer to stick to.

koheiw:
You want to add this function for beginners who do not know how to apply dictionary analysis. If so, it does not need to support continuous polarity scores and non-linear transformation (e.g. logit). Users who need those should write a function themselves. If you want write a code for them or do something more, the **dictionarytools** package is the best place.

PR #1954, opened 2020-05-16
Add interface to set hash table sizes

koheiw:
Add interface to change max_load_factor. It is a setting of the hash table that affects performance of C++ functions. 
```r
quanteda:::qatd_cpp_set_load_factor("pattern", 0.025)
quanteda:::qatd_cpp_get_load_factor()
```
We will not export these functions, but they allow us to find optimal settings on different systems with different data.

### DIFF
--- a/src/RcppExports.cpp
+++ b/src/RcppExports.cpp
@@ -304,6 +304,27 @@ BEGIN_RCPP
     return rcpp_result_gen;
 END_RCPP
 }
+// qatd_cpp_set_load_factor
+void qatd_cpp_set_load_factor(std::string type, float value);
+RcppExport SEXP _quanteda_qatd_cpp_set_load_factor(SEXP typeSEXP, SEXP valueSEXP) {
+BEGIN_RCPP
+    Rcpp::RNGScope rcpp_rngScope_gen;
+    Rcpp::traits::input_parameter< std::string >::type type(typeSEXP);
+    Rcpp::traits::input_parameter< float >::type value(valueSEXP);
+    qatd_cpp_set_load_factor(type, value);
+    return R_NilValue;
+END_RCPP
+}
+// qatd_cpp_get_load_factor
+List qatd_cpp_get_load_factor();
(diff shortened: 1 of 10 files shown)


koheiw:
Smaller max_load_factor increases the speed but users more memory
```r
> corp <- quanteda.corpora::download("data_corpus_guardian")
> quanteda_options(threads = 8)
> 
> system.time({
+   quanteda:::qatd_cpp_set_load_factor("pattern", 0.1)
+   tokens(corp)
+ })
   user  system elapsed 
   9.78    0.11    7.40 
> 
> system.time({
+   quanteda:::qatd_cpp_set_load_factor("pattern", 0.05)
+   tokens(corp)
+ })
   user  system elapsed 
   9.76    0.17    7.22 
> 
> system.time({
+   quanteda:::qatd_cpp_set_load_factor("pattern", 0.01)
+   tokens(corp)
+ })
   user  system elapsed 
   8.92    0.18    6.70 
>
```

src/RcppExports.cpp
```diff
@@ -304,6 +304,27 @@ BEGIN_RCPP
     return rcpp_result_gen;
 END_RCPP
 }
+// qatd_cpp_set_load_factor
```
kbenoit, line 307: Can you provide some (internal) function documentation?

koheiw:
There is a significant difference between this branch and master on my system. The performance gain is from twice faster `tokens_select()`, but I am not sure why the gap is so large... please take a benchmark on you system.

This branch
```
+ toks <- tokens(head(corp, 10000)),
+ times = 10
+ )
Unit: seconds
                              expr      min      lq     mean   median       uq      max neval
 toks <- tokens(head(corp, 10000)) 7.864123 8.25525 9.015221 8.463831 9.798426 11.52866    10
```
master
```
+ toks <- tokens(head(corp, 10000)),
+ times = 10
+ )
Unit: seconds
 toks <- tokens(head(corp, 10000)) 9.213969 11.93565 12.47803 12.84023 13.71376 13.92283    10
```

PR #1955, opened 2020-05-17
Implement cached summary

koheiw:
I have finished writing the draft code for the new `summary` function with caching. We need to discuss decided the detail sometime soon, but please play with it.

We need to decided

- What columns a summary should have?
- Whether we need `print.summary` or not?

Currently 
```
           doc_id number punct symbol any n_token n_type      noise n_sent is_dup
1 1789-Washington      1   108      0 109    1537   9360 0.07091737     23  FALSE
2 1793-Washington      0    12      0  12     147   9360 0.08163265      4  FALSE
3      1797-Adams      0   263      0 263    2577   9360 0.10205666     37  FALSE
4  1801-Jefferson      0   203      0 203    1923   9360 0.10556422     41  FALSE
5  1805-Jefferson      0   216      0 216    2380   9360 0.09075630     45  FALSE
6    1809-Madison      0    88      0  88    1261   9360 0.06978588     21  FALSE
```


koheiw:
You can see that the cache is passed to the downstream object, but updated when `summary()` is run on a new or modified objects.
https://github.com/quanteda/quanteda/blob/72af0042d4021ed1e9ba114a87f005f1510366d1/tests/testthat/test-summary.R#L95-L122

This will be a breaking change so one option is adding this as `textstat_summary()` and deprecate `summary()`.

kbenoit:
`textstat_summary()` is a great idea, and very consistent with our idea of textstat functions. We could cache the information after running either `summary()` or `textstat_summary()`, with the two differing as follows:

- `textstat_summary()` would produce summary statistics on the texts, but not output docvars. It could trigger the caching of this information into the object, for future use. This could work on character, corpus, tokens, and dfm objects.

- `summary()` would continue temporarily to operate as it does now, but we would deprecate it or at least make it more minimal in output, and strongly discourage (with removal soon) its use to return the invisible data.frame of the summary data, which should be the function of `textstat_summary()`. We could either strip `summary()` to a minimal set of reporting about an object, or move to deprecate it entirely. It might be nice to retain it in a simpler fashion (and add it to other objects), since it's such a fundamental and widely used function in R. We could also have it trigger the caching of the full set of information to be reported in `textstat_summary()`, even though `summary()` itself would not output all of that information.

koheiw:
As a transition strategy, we can call `textstat_summary()` from `summary()` but aim to make `summary()` much simpler. A simplified `summary()` would be just to show `ndoc()` and `ntype()` or `nfeat()` without any heavy operations.

koheiw:
With the `toks <- tokens(data_corpus_inaugural)` corpus, the new function fully execute tokens counts as you can see `min = 118.7311` as because the value for `remove` is different every time.

```
+     textstat_summary(toks, remove = sample(stopwords(), 10))
+ )
Unit: milliseconds
                                                     expr      min       lq     mean   median       uq      max
 textstat_summary(toks, remove = sample(stopwords(), 10)) 118.7311 138.9327 168.4645 151.3495 166.4806 953.3552
 neval
   100
> 
```
However, it is only once when the value is the same as you can see `max = 163.7101`.
```
+     textstat_summary(toks, remove = stopwords()[1:10])
+ )
Unit: milliseconds
                                               expr      min      lq     mean   median      uq      max neval
 textstat_summary(toks, remove = stopwords()[1:10]) 14.21151 15.2399 18.58585 16.47446 18.6645 163.7101   100

```

koheiw:
We might want to add deprecation message to `summary()`.

kbenoit:
Impressive. I'm not sure I fully understood the `remove` point about subsequent calls though, but I look forward to your explanation.

A few reactions:

- Shouldn't we also cache this when `summary()` is executed?

- RIP `summary.character()` ??

- What if we updated the cache rather than clearing it, when modifying the object? I assume this has to recompute the stats:
```r
toks <- tokens("One one two three gr8!")
textstat_summary(toks, cache = TRUE)
tokens_remove(toks, "one") %>%
    textstat_summary(cache = TRUE)
```

- We can discuss the appropriate summary fields and their names. (I'd like to include emoji for instance.)

- In `summarize_texts_extended()` I faced the same issues and implemented the counts in terms of our `tokens()` functions using `remove_*` definitions. This results in slightly different counts, since by our definition, "gr8" should not be counted as a "number". This might be a better approach, for consistency.
```r
> quanteda:::summarize_texts_extended(toks) %>%
+     head(7)
$total_tokens
[1] 6

$all_tokens
text1 
    6 

$total_punctuation
[1] 1

$total_symbols
[1] 0

$total_numbers
[1] 0

$total_words
[1] 5

$total_stopwords
[1] 0
```

- Should the `...` allow dfm options? I can see that the correct numbers as (re)computed, but why allow the function to describe a different object, by allowing modifications in the summary function?
```r
> textstat_summary(toks, remove = "one")
  document n_token n_type n_sent number punct symbol any noise is_dup
1    text1       4      4     NA      1     1      0   2   0.5  FALSE
> textstat_summary(toks)
  document n_token n_type n_sent number punct symbol any     noise is_dup
1    text1       6      5     NA      1     1      0   2 0.3333333  FALSE
```

koheiw:
``` r
require(quanteda)
#> Package version: 2.0.2
#> Parallel computing: 2 of 6 threads used.
#> 
#> 
#>     View
txt <- c("£ € 👏 Rock on❗ 💪️🎸",
         "Hi #qi #quanteda https://quanteda.io")
toks <- tokens(txt)
textstat_summary(toks)
#>       document duplicated n_sent n_token n_type punct numbers symbols url
#> text1    text1      FALSE     NA       8      8     0       0       6   0
#> text2    text2      FALSE     NA       4      4     0       0       0   1
#>       hashtag emoji
#> text1       0     4
#> text2       2     0
```

<sup>Created on 2020-05-21 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

kbenoit:
On `remove_numbers`, do you mean in the `tokens()` signature? Everything is plural except `remove_url`, so for consistency, we would pluralise that. (I'm not counting `remove_punct` since thats not only an abbreviation so does not count, but also a mass noun.

Plus shorter versions resolve to longer formals, but not in reverse.
```r
> tokens("1 two 3", remove_number = TRUE)
text1 :
[1] "two"
```

kbenoit:
So on the `textstat_summary()` columns, two choices to be made:

1.  Naming scheme. Should it be `n_item`? or `item`? 
2. Plural v singular. Should `item` be plural or singular? Right now we have both (`symbols`, `hashtag`).

I suggest using `n_item` and making it shortened and singular. This is closest to `ntype()`, `ndoc()`, etc. So: `n_sent` `n_tok` `n_type` `n_punct`, `n_number`, `n_ symbol`, `n_url`, `n_tag`, `n_emoji`.

Do we also want `n_char`?

(Here I changed `hashtag` to `tag` since it covers both @usernames and #hashtags.)

On `duplicated`:  I'd be in favour of dropping this, since it's trivial to compute using `duplicated()`
```r
> corp <- corpus(c("one", "one", "two 2", "two 2"),
+                docvars = data.frame(num = c(1, 2, 3, 3)))
> textstat_summary(corp)
      document duplicated n_sent n_token n_type punct numbers symbols url hashtag emoji
text1    text1      FALSE      1       1      1     0       0       0   0       0     0
text2    text2       TRUE      1       1      1     0       0       0   0       0     0
text3    text3      FALSE      1       2      2     0       1       0   0       0     0
text4    text4       TRUE      1       2      2     0       1       0   0       0     0
> duplicated(corp)
[1] FALSE  TRUE FALSE  TRUE
> duplicated(tokens(corp))
[1] FALSE  TRUE FALSE  TRUE
```

I accept that this also applies to `n_char`, so we could drop that too, although it's a bit trickier to compute "manually" for tokens, and will not count separators already removed.
```r
> stringi::stri_length(corp)
[1] 3 3 5 5

# but:
> tokens(corp) %>%
+     sapply(function(y) sum(stringi::stri_length(y)))
text1 text2 text3 text4 
    3     3     4     4 

# only way to keep the separators is to go back to the source
> tokens(corp, remove_separators = FALSE) %>%
+     sapply(function(y) sum(stringi::stri_length(y)))
text1 text2 text3 text4 
    3     3     5     5 
``` 

Of course we could add, for corpus objects, `n_separator`...

koheiw:
How about this? I remove `n_` from all because they are redundant. I fancy `duplicated` as it is useful and does not take much time even for tokens. `char` is only for a corpus. `tag` is a bit tricky because it need to be taken from `quanteda_options()` (potential inconsistency).

``` r
require(quanteda)
#> Package version: 2.0.2
#> Parallel computing: 2 of 6 threads used.
#> 
#> 
#>     View
txt <- c("£ € 👏 Rock on❗ 💪️🎸",
         "Hi @qi #quanteda https://quanteda.io")

corp <- corpus(txt)
textstat_summary(corp, cache = FALSE)
#>   document duplicated sent char token type punct numbers symbols url tag emoji
#> 1    text1      FALSE    1   18     8    8     0       0       6   0   0     4
#> 2    text2      FALSE    1   36     4    4     0       0       0   1   2     0

toks <- tokens(txt)
textstat_summary(toks, cache = FALSE)
#>   document duplicated sent char token type punct numbers symbols url tag emoji
#> 1    text1      FALSE   NA   NA     8    8     0       0       6   0   0     4
#> 2    text2      FALSE   NA   NA     4    4     0       0       0   1   2     0

dfmt <- dfm(txt)
textstat_summary(dfmt, cache = FALSE)
#>   document duplicated sent char token type punct numbers symbols url tag emoji
#> 1    text1         NA   NA   NA     8    8     0       0       6   0   0     4
#> 2    text2         NA   NA   NA     4    4     0       0       0   1   2     0
```

<sup>Created on 2020-05-21 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

kbenoit:
I like keeping tags since that is a good flag for whether social media handling is needed.

I still favor not including `duplicated`, then all results are counts. It's trivial to get using `duplicated()` and is not really a summary of the text, but rather an attribute of texts in the context of the corpus/tokens. There should not be a difference between:
```r
textstat_summary(corp)[1]
textstat_summary(corp[1])
```

if we drop the `n_`, then we should standardize plurals. I vote for:
sents chars tokens types puncts numbers symbols urls tags emojis

or n_* and making it plural.

I guess for summaries of a weighted dfm, we just sum the weighted values and let the user beware?

koheiw:
OK, let's drop `duplicated` and make all end with `s`. Please review and merge the new PR #1957. I need it for this branch.

Normalized dfm should be fine because `dfm_lookup` takes the sum of the proportions. `Log-transformed dfm is interesting, because it is a interaction (multiplication) of words in the same category.

PR #1957, opened 2020-05-21
Update preserve_special()

koheiw:
Update `preserve_special()` for more accurate counting of social media user names in `textstat_summary()`.
 - Integrate protection of website and email address
 - Use pattern_username option only for usernames

### DIFF
--- a/R/quanteda_options.R
+++ b/R/quanteda_options.R
@@ -155,6 +155,6 @@ get_options_default <- function(){
                  base_compname = "comp",
                  language_stemmer = "english",
                  pattern_hashtag = "#\\w+#?",
-                 pattern_username = "@")
+                 pattern_username = "@[a-zA-Z0-9]+")
     return(opts)
 }
--- a/R/quanteda_options.R
+++ b/R/quanteda_options.R
@@ -62,15 +62,13 @@ quanteda_options <- function(..., reset = FALSE, initialize = FALSE) {
         args <- args[[1]]
     
     # initialize automatically it not yet done so
-    if (is.null(options('quanteda_initialized')) || !"package:quanteda" %in% search())
+    if (is.null(getOption('quanteda_initialized')) || !"package:quanteda" %in% search())
         quanteda_initialize()
         
     if (initialize) {
-        quanteda_initialize()
-        return(invisible(TRUE))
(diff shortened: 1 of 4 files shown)


R/quanteda_options.R
```diff
@@ -155,6 +155,6 @@ get_options_default <- function(){
                  base_compname = "comp",
                  language_stemmer = "english",
                  pattern_hashtag = "#\\w+#?",
-                 pattern_username = "@")
+                 pattern_username = "@[a-zA-Z0-9]+")
```
kbenoit, line 158: That's not quite precise enough. 
``` r
patt1957 <- "@[a-zA-Z0-9]+"
pattbetter <- "^@[a-zA-Z0-9_]+$"

txt <- c("kbenoit@lse.ac.uk", "@@invalid", "@kenbenoit", "@4ever", "@ken_ben")

stringi::stri_detect_regex(txt, patt1957)
## [1] TRUE TRUE TRUE TRUE TRUE
stringi::stri_detect_regex(txt, pattbetter)
## [1] FALSE FALSE  TRUE  TRUE  TRUE
```
Twitter rules are here https://help.twitter.com/en/managing-your-account/twitter-username-rules and allow just 15 characters, so we could even say:
```r
"^@[a-zA-Z0-9_]{1,15}$"
```
if we wanted to stick to Twitter.

There is an accepted answer [on SO](https://stackoverflow.com/questions/2304632/regex-for-twitter-username/6351873) that is more complicated, but I think it's based on when Twitter did not permit the first character of the username to be a number. (That's permitted now.)

koheiw: This is not for tokens, not need of `^` and `$` but added `_`.

koheiw: Internet address will be captured by `address <- "(https?:\\/\\/(www\\.)?|@)[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-z]{2,4}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)"` before the username matching in `preserve_special()`.

kbenoit: OK, as long as it works (wherever it's used) correctly for the test cases above.

PR #1958, opened 2020-05-21
Add textstat_sentiment() with unified valence approach

kbenoit:
OK Sensei, hozwabout this approach.

* Everything is valence. What I called "polarity" earlier is all treated as a special case of valence, where all dictionary values within a key have the same valence scores. To emulate the "polarity" approach, these need to be assigned -1 and 1 for negative and positive, which means that if we feed it smoothed, logged counts, we can get the equivalent of log(p/n) (which is as you know my favourite sentiment scale for polar categories).
* Computing sentiment uses your approach to "flipping" the valence scores and making them into a dictionary, and then computing the dfm where the "features" are the valence scores. Then we can use `%*%` for the combination to the document level.
* It's all part of meta$object$valence - so normal items are unaffected
* We do all of the robustness not through overloading indexing operators etc, but simply by naming each valence score element, and matching those. Any value not valenced will not be counted.

Let me know what you think, and I can start rewriting the tests and making this as robust as possible.

Note: Replaces #1948

### Demonstration

``` r
library("quanteda")
## Package version: 2.0.2

# simple example
txt <- c(
  d1 = "Great, I really like positive statements, not bad!",
  d2 = "This crap movie sucks bad, I'm not happy about it!"
)
toks <- tokens(txt)

dict <- dictionary(list(
  positive = c("great", "really like", "positive"),
  neg_positive = c("not happy"),
  negative = c("crap", "sucks", "bad"),
  neg_negative = c("not bad")
))

tokens_lookup(toks, dict, nomatch = "other") %>%
  dfm()
## Document-feature matrix of: 2 documents, 5 features (30.0% sparse).
##     features
## docs positive neg_positive negative neg_negative other
##   d1        3            0        1            1     5
##   d2        0            1        3            0     7

# METHOD 1
valence(dict) <- c(positive = 1, negative = -1)
# internally:
valence(dict)
## $positive
##       great really like    positive 
##           1           1           1 
## 
## $negative
##  crap sucks   bad 
##    -1    -1    -1

textstat_sentiment(toks, dict)
##   doc_id sentiment
## 1     d1       0.5
## 2     d2      -1.0
valence(dict) <- c(positive = 2, neg_negative = 1, negative = -2, neg_positive = -1)
textstat_sentiment(toks, dict)
##   doc_id sentiment
## 1     d1      1.00
## 2     d2     -1.75

# METHOD 2
valence(dict) <- list(positive = c(1, 2, 1.5), negative = c(-2, -2, 1))
[...]

PR #1962, opened 2020-05-24
Add logsmooth, needed for sentiment

kbenoit:
Adds new scheme for weighting, which is log(x + smoothing), needed for the logit scaling of sentiment (in #1958).

### DIFF
--- a/R/dfm_weight.R
+++ b/R/dfm_weight.R
@@ -121,7 +121,7 @@ dfm_weight.dfm <- function(
             .Deprecated(msg = 'scheme = "tfidf" is deprecated; use dfm_tfidf(x) instead')
             return(dfm_tfidf(x, base = base))
         } else if (scheme == "logsmooth") {
-            return(dfm_smooth(x, smoothing) %>% log(base = base) %>% as.dfm())
+            return(as.dfm(log(dfm_smooth(x, smoothing), base = base)))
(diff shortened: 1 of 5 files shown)


R/dfm_weight.R
```diff
@@ -112,6 +120,8 @@ dfm_weight.dfm <- function(
         } else if (scheme == "tfidf") {
             .Deprecated(msg = 'scheme = "tfidf" is deprecated; use dfm_tfidf(x) instead')
             return(dfm_tfidf(x, base = base))
+        } else if (scheme == "logsmooth") {
+            return(dfm_smooth(x, smoothing) %>% log(base = base) %>% as.dfm())
```
koheiw, line 124: Let's not induce unnecessary dependency on `%>%`. **quanteda** should not fail because of a bug convenience packages.

kbenoit: We re-export (from **magrittr**) it already, and it's such an integral part of many packages that it's unlikely to change.

But I'm indifferent. Is this better?
```r
return(as.dfm(log(dfm_smooth(x, smoothing), base = base)))
```

koheiw: Exporting magrittr is fine (I use it myself outside of functions), because we don't use `%>%` when it is broken.

Issue #1965, opened 2020-05-28
Consider parallelizing tokenization


koheiw:
There is a package called **future.apply** which provides parallelized apply-type functions. It seems that we can parallelize tokenization with `future_lapply()`.
```r
require(quanteda)
require(future.apply)
plan(multiprocess)

> corp <- readRDS("/home/kohei/Dropbox/Brexit/Data/data_corpus_guardian_2016.RDS")
> #corp <- quanteda.corpora::download("data_corpus_guardian")
> txt <- texts(corp)
> length(txt)
[1] 60045
> 
> chunks <- split(txt, rep_len(1:4, length(txt)))
> microbenchmark::microbenchmark(
+   do.call(c, future_lapply(chunks, tokens, remove_separator = FALSE)),
+   quanteda::tokens(txt, remove_separator = FALSE),
+   times = 1
+ )
Unit: seconds
                                                                expr       min        lq      mean    median
 do.call(c, future_lapply(chunks, tokens, remove_separator = FALSE))  68.73448  68.73448  68.73448  68.73448
                     quanteda::tokens(txt, remove_separator = FALSE) 105.10601 105.10601 105.10601 105.10601
        uq       max neval
  68.73448  68.73448     1
 105.10601 105.10601     1
```

kbenoit:
I saw a presentation on this remarkable package at the RStudio::conf in January, was thinking exactly the same!

koheiw:
I experimented with parallelization in different ways. My initial idea was call `string_split_boundaries()` in parallel but it was slower, probably because of the large object size (list of character vectors). So, serialization should be done before returning the tokens. This means we have to reassign token IDs very quickly to combine multiple objects. `tokens.c()` is not terrible slow, but it can be done more efficiently.

kbenoit:
I found the same using other parallelisation methods (from R): the vectorized `stri_split_boundaries()` was still fastest.

Still we batch the tokens now, maybe that could be executed in parallel?

koheiw:
Yes, this part will be the target of parallelization.

https://github.com/quanteda/quanteda/blob/70ceece7f93901e60e7cd67fe88ff97d17306e68/R/tokens.R#L274-L286

but serialization depends on `attr(x[[i - 1]], "types")`, so need to make recompiliation very fast, especially 

https://github.com/quanteda/quanteda/blob/70ceece7f93901e60e7cd67fe88ff97d17306e68/R/tokens-methods-base.R#L184-L185

[...]

Issue #1926, opened 2020-04-12
Should we really normalize hyphens


koheiw:
In #1925, we noticed that normalization of hyphens in the corpus constructor is causing inconsistency. 

https://github.com/quanteda/quanteda/blob/5a8b661944d4917006a23131bd850c3bee4be32c/R/corpus.R#L153-L156

If corpus is a an object to keep original texts, we probably should not do it in the corpus constructor but in token constructor.

PR #1966, opened 2020-06-05
Optimize tokenization loop for very large corpora

koheiw:
This is for restructuring of tokens internals to increase the efficiency with very large corpora.

- Concentrate slow functions within a `lapply()` loop (which can be replaced by `future_lapply()`) (#1965)
- Change how tokenization progress is displayed
- Move character normalization (e.g. hyphens) to tokenizer
- Fix handling of the `verbose` option in `tokens.tokens()` (#1926)

The internal design is tested in `tokens_parallel()` in [a separate branch](https://github.com/quanteda/quanteda/blob/dev-tokens_parallel/R/tokens_parallel.R) and worked with a corpus of 30 million texts (20GB) (takes only 30 min in serial process).

Remaining questions are

- How to print the progress when `verbose = TRUE`?
- When should be switch to `future_lapply()`?

### DIFF
--- a/R/tokenizers.R
+++ b/R/tokenizers.R
@@ -120,9 +120,10 @@ restore_special <- function(x, special, recompile = TRUE) {
             )
         }
     }
-    if (!identical(type, attr(x, "types")) && recompile) {
+    if (!identical(type, attr(x, "types"))) {
         attr(x, "types") <- type
-        x <- tokens_recompile(x)
+        if (recompile)
+            x <- tokens_recompile(x)
     }
     return(x)
 }
@@ -164,7 +165,7 @@ preserve_special1 <- function(x, split_hyphens = TRUE, split_tags = TRUE, verbos
 }
 
 # re-substitute the replacement hyphens and tags
-restore_special1 <- function(x, split_hyphens = TRUE, split_tags = TRUE, recopile = TRUE, 
+restore_special1 <- function(x, split_hyphens = TRUE, split_tags = TRUE, recompile = TRUE, 
                              verbose = FALSE) {
     type <- attr(x, "types")
     if (!split_hyphens)
@@ -172,9 +173,10 @@ restore_special1 <- function(x, split_hyphens = TRUE, split_tags = TRUE, recopil
     if (!split_tags)
         type <- stri_replace_all_fixed(type, c("_ht_", "_as_"), c("#", "@"),
                                        vectorize_all = FALSE)
-    if (!identical(type, attr(x, "types")) && recopile) {
+    if (!identical(type, attr(x, "types"))) {
         attr(x, "types") <- type
-        x <- tokens_recompile(x)
+        if (recompile)
+            x <- tokens_recompile(x)
     }
     return(x)
 }
--- a/R/tokenizers.R
+++ b/R/tokenizers.R
@@ -229,3 +229,21 @@ tokenize_fasterword <- function(x, ...) {
 tokenize_fastestword <- function(x, ...) {
     stri_split_regex(x, " ")
 }
+
+
+normalize_characters <- function(x) {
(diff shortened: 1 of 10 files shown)


koheiw:
Speed is more or less the same for the medium size Guardian corpus
```r
 > length(txt) 
[1] 204061

# this branch
+ tokens(txt, verbose = FALSE), times = 5
+ )
Unit: seconds
 tokens(txt, verbose = FALSE) 313.8735 315.4183 317.3557 317.2339 319.5595 320.6934     5

Restarting R session...


# master
+ tokens(txt, verbose = FALSE), times = 5
+ )
Unit: seconds
 tokens(txt, verbose = FALSE) 302.9805 304.0984 310.1614 305.5729 315.5114 322.6437     5
 

```

R/tokens.R
```diff
@@ -264,42 +264,44 @@ tokens.corpus <- function(x,
...
-        if (verbose) catm(" ...tokenizing", i, "of", length(x), "blocks\n")
-        temp <- tokenizer_fn(x[[i]], split_hyphens = split_hyphens, verbose = verbose)
-        if (verbose) catm(" ...serializing tokens, ")
-        if (i == 1) {
-            x[[i]] <- serialize_tokens(temp)
-        } else {
-            x[[i]] <- serialize_tokens(temp, attr(x[[i - 1]], "types"))
+    x <- texts(x)
+    x <- split(x, factor(ceiling(seq_along(x) / 10000)))
+    x <- lapply(x, function(y) {
```
koheiw, line 273: This can be `future_lapply`.

R/tokenizers.R
```diff
@@ -38,7 +38,7 @@ NULL
 #' @export
 tokenize_word <- function(x, split_hyphens = FALSE, verbose = quanteda_options("verbose")) {
     
-    if (verbose) catm(" ...segmenting tokens\n")
+    if (verbose) catm(" ...segmenting texts\n")
```
kbenoit, line 41: perhaps: `" ...segmenting words\n"` ?

R/tokens.R
```diff
@@ -264,42 +264,44 @@ tokens.corpus <- function(x,
...
-    for (i in seq_along(x)) {
-        if (verbose) catm(" ...tokenizing", i, "of", length(x), "blocks\n")
-        temp <- tokenizer_fn(x[[i]], split_hyphens = split_hyphens, verbose = verbose)
-        if (verbose) catm(" ...serializing tokens, ")
-        if (i == 1) {
-            x[[i]] <- serialize_tokens(temp)
-        } else {
-            x[[i]] <- serialize_tokens(temp, attr(x[[i - 1]], "types"))
+    x <- texts(x)
+    x <- split(x, factor(ceiling(seq_along(x) / 10000)))
```
kbenoit, line 272: Doesn't have to be done here, but at some point the fixed 10000 might best to move to an option.

kbenoit:
Oh and on the other qn: Let's merge this first and work on the `future_lapply()` next. That way we isolate the parallelism issues and can test that throughly on different platforms.

R/tokenizers.R
koheiw: Change to `segmenting into words`.

R/tokens.R
koheiw: Added a quanteda option `block_size`

koheiw:
I hide  process ID from the verbose message for now because it is useful only in parallel tokenization. "print the verbose stuff on just the first pass" is possible only in serial tokenization. We can work on these in parallelization branch.

R/tokens.R
kbenoit: 👍  but how about we call it `tokens_block_size` - it's only for the tokens construction, right?

Issue #1976, opened 2020-06-30
textstat_readability() returns nothing for an empty document


kbenoit:
The question has arisen in #1969 of how to handle `NA` documents. What about empty string documents? Right now, `textstat_readability()` returns nothing if the document was `""`. This is not true for other `textstat` returns. We should make them consistent.

``` r
library("quanteda")
## Package version: 2.1.0

txt <- c(d1 = "The cat in the hat", d2 = "", d3 = "Once upon a time.")
corp <- corpus(txt)
toks <- tokens(corp)
dfmat <- dfm(toks)

textstat_readability(txt)
##   document  Flesch
## 1       d1 117.160
## 2       d3  97.025
textstat_readability(corp)
##   document  Flesch
## 1       d1 117.160
## 2       d3  97.025

textstat_dist(dfmat) %>%
  as.data.frame()
##   document1 document2 euclidean
## 1        d3        d1  3.464102
## 2        d1        d3  3.464102

textstat_entropy(dfmat)
##   document  entropy
## 1       d1 1.921928
## 2       d2 0.000000
## 3       d3 2.321928

textstat_lexdiv(toks)
##   document TTR
## 1       d1 0.8
## 2       d2 NaN
## 3       d3 1.0
textstat_lexdiv(dfmat)
##   document TTR
## 1       d1 0.8
## 2       d2 NaN
## 3       d3 1.0
```

**Suggestion**

Return NA for `textstat_readability()` if the document is zero.

PR #1977, opened 2020-06-30
Fix 1976

kbenoit:
Don't drop documents from the results of `textstat_readability()`, `textstat_lexdiv()`, and `nscrabble()` when documents are empty.

Also fixes a bug in `print.corpus()` caused by empty documents.

Solves #1976

### DIFF
--- a/R/textstat_lexdiv.R
+++ b/R/textstat_lexdiv.R
@@ -364,6 +364,7 @@ compute_lexdiv_dfm_stats <- function(x, measure = NULL, log.base = 10) {
     result <- data.frame(document = docnames(x), stringsAsFactors = FALSE)
     if (length(measure))
         result <- cbind(result, as.data.frame(temp[, measure, with = FALSE]))
+    result[is.na(result)] <- NA
     class(result) <- c("lexdiv", "textstat", "data.frame")
     return(result)
 }
@@ -388,6 +389,7 @@ compute_lexdiv_tokens_stats <- function(x, measure = c("MATTR", "MSTTR"),
 
     # reorder output as originally supplied
     result <- result[, c("document", measure), drop = FALSE]
+    result[is.na(result)] <- NA
     class(result) <- c("lexdiv", "textstat", "data.frame")
     return(result)
 }
--- a/R/textstat_readability.R
+++ b/R/textstat_readability.R
@@ -526,9 +526,11 @@ textstat_readability.corpus <- function(x,
 
     x <- texts(x)
     if (!is.null(min_sentence_length) || !is.null(max_sentence_length)) {
-        x <- char_trim(x, "sentences",
-                       min_ntoken = min_sentence_length,
-                       max_ntoken = max_sentence_length)
+        temp <- char_trim(x, "sentences",
+                          min_ntoken = min_sentence_length,
+                          max_ntoken = max_sentence_length)
+        x[names(temp)] <- temp
+        x[!names(x) %in% names(temp)] <- ""
     }
 
     # get sentence lengths - BEFORE lower-casing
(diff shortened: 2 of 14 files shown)


R/textstat_readability.R
```diff
@@ -526,9 +526,11 @@ textstat_readability.corpus <- function(x,
 
     x <- texts(x)
     if (!is.null(min_sentence_length) || !is.null(max_sentence_length)) {
-        x <- char_trim(x, "sentences",
-                       min_ntoken = min_sentence_length,
-                       max_ntoken = max_sentence_length)
+        temp <- char_trim(x, "sentences",
+                          min_ntoken = min_sentence_length,
+                          max_ntoken = max_sentence_length)
+        x[names(temp)] <- temp
```
koheiw, line 532: This could be done in a more robust way in corpus using factor levels of docid etc.

kbenoit: Yes it could, and the root of the problem is deeper, but here it's just a fix to prevent documents from disappearing. The root problem comes when we reshape to sentences, which obliterates empty documents, and then they cannot be reshaped to documents. I thought we (you 😀) could fix this later.

``` r
library("quanteda")
## Package version: 2.1.0
## Parallel computing: 2 of 8 threads used.
## 
## 
##     View

corp <- corpus(c(
  d1 = "The cat in the hat at green ham and eggs.",
  d2 = "",
  d3 = "Once upon a time."
))
corp
## Corpus consisting of 3 documents.
## d1 :
## "The cat in the hat at green ham and eggs."
## 
## d2 :
## ""
## 
## d3 :
## "Once upon a time."

corp %>%
  corpus_reshape(to = "documents")
## Corpus consisting of 2 documents.
## d1 :
## "The cat in the hat at green ham and eggs."
## 
## d3 :
## "Once upon a time."
```

R/textstat_lexdiv.R
```diff
@@ -388,6 +389,7 @@ compute_lexdiv_tokens_stats <- function(x, measure = c("MATTR", "MSTTR"),
 
     # reorder output as originally supplied
     result <- result[, c("document", measure), drop = FALSE]
+    result[is.na(result)] <- NA
```
koheiw, line 392: Doesn't `is.na(result)` trigger conversion of data.frame to matrix? If so, it can be more efficient.

kbenoit: no, it just replaces the values, does not drop the data.frame class.

R/textstat_readability.R
koheiw: That was my guess. If you postpone solving problems, they just become bigger.

You could use `segment_texts()` which is the internal for `corpus_reshape()`:
```r
segment_texts <- function(x, pattern = NULL, valuetype = "regex",
                          case_insensitive = TRUE,
                          extract_pattern = FALSE, pattern_position = "after",
                          omit_empty = TRUE, what = "other", ...) {
```
Interestingly, `omit_empty = FALSE` here...

https://github.com/quanteda/quanteda/blob/c43c37b29e581674f0bd0ed503d02b75e84fda7f/R/corpus_reshape.R#L68-L69

kbenoit: Actually the problem is in https://github.com/quanteda/quanteda/pull/1977/files#diff-745460f071607a11357036687422228dR50, where the `corpus_subset()` removes documents - here, sentences - that don't meet the conditions. Adding the code in `textstat_readability()` as I have ensures that no document is ever dropped from the result. We can open a separate issue for what to do with operations that result in empty (`""`) or `NA` documents, but I think ensuring that all documents are returned here - meaning https://github.com/quanteda/quanteda/pull/1977/files#diff-745460f071607a11357036687422228dR50 -  in is still appropriate.

koheiw: The links do not seem to be working. Can you link to the code in the master?

kbenoit: OK will do. The first link was 
https://github.com/quanteda/quanteda/blob/c43c37b29e581674f0bd0ed503d02b75e84fda7f/R/corpus_trim.R#L50
where the subset operation removes the "document". So a one-sentence document that is shorter than the minimum readability tokens is simply removed. The extra code in `textstat_readability()` ensures this will never happen.

But I will also open a larger issue about what happens to documents during reshape operations etc. when those documents are empty.

The second link was just referring to the same code at the top of this comment, i.e.
```r
    x[names(temp)] <- temp
```

Issue #1979, opened 2020-07-03
Make stringi's locale option accessible 


koheiw:
After seeing an [SO post](https://stackoverflow.com/questions/62691994/quantedas-corpus-reshape-function-how-not-to-break-sentences-after-abbreviatio), I am pressed to expose `locale` in `stringi::stri_opts_brkiter()`. The best place would be `quanteda_options(tokens_locale = "")`. 

```
> stri_split_boundaries("Mr. Jones and Mrs. Brown are very happy.
+ So am I, Prof. Smith.", type="sentence", locale="en_US@ss=standard") # ICU >= 56 only
[[1]]
[1] "Mr. Jones and Mrs. Brown are very happy.\n" "So am I, Prof. Smith."                     
```

kbenoit:
OK, that's really interesting. I didn't know there were locale options that could be set to affect sentences. We should include the "ss=standard" in the tokenizer for what = "sentence" by default.

```r
> stri_split_boundaries("Mr. Jones and Mrs. Brown are very happy.
+ So am I, Prof. Smith.", type="sentence", locale="en_GB")
[[1]]
[1] "Mr. "                    "Jones and Mrs. "         "Brown are very happy.\n" "So am I, Prof. "        
[5] "Smith."                 

> stri_split_boundaries("Mr. Jones and Mrs. Brown are very happy.
+ So am I, Prof. Smith.", type="sentence", locale="en_GB@ss=standard")
[[1]]
[1] "Mr. Jones and Mrs. Brown are very happy.\n" "So am I, Prof. Smith."  
```

koheiw:
These are the hard-coded patterns in our tokenizer. It seems that many of the patters are handled properly by recent ICU locale. "Dr" is an exception but I think it don't think it is a big problem. I am adding `tokens_locale = "en_US@ss=standard"` to the options.
```
c("Mr", "Mrs", "Ms", "Dr", "Jr", "Prof", "Ph.D", "M", "MM", "St", "etc")
```
```r
> stri_split_boundaries("He is Mr. Jones.", type = "sentence", locale = "en_US@ss=standard")
[[1]]
[1] "He is Mr. Jones."

> stri_split_boundaries("She is Ms. Jones.", type = "sentence", locale = "en_US@ss=standard")
[[1]]
[1] "She is Ms. Jones."

> stri_split_boundaries("Jones Jr. is him.", type = "sentence", locale = "en_US@ss=standard")
[[1]]
[...]

Issue #1978, opened 2020-07-02
corpus_reshape and trim operations drop documents


kbenoit:
When a document is reshaped to sentences, and then some operation removes all of the original document's sentences - such as in `corpus_trim()` - then reshaping to the document level removes that document. This seems like an unanticipated side effect, so if we want this to be policy, we should decide on that and document it rather than having it be something of a surprise.

I noticed this when answering [this SO question](https://stackoverflow.com/questions/62628221/r-applying-quantedas-textstat-readability-function-producing-error-in-set/62654993#62654993) affecting `textstat_readability()`. Since that function trims short sentences rather than score them, it was removing an entire document that consisted only of one short sentence. (This can happen a lot in Tweets.) We fixed that in https://github.com/quanteda/quanteda/pull/1977 but made a note to revisit the underlying issue for a more careful and more fundamental fix.

``` r
library("quanteda")
## Package version: 2.1.0
## Parallel computing: 2 of 8 threads used.
## See https://quanteda.io for tutorials and examples.
## 
## Attaching package: 'quanteda'
## The following object is masked from 'package:utils':
## 
##     View

txt <- c(
  d1 = "The cat in the hat ate green ham and eggs.",
  d2 = "",
  d3 = "Once upon a time.",
  d4 = NA
)

corp <- corpus(txt)
## Warning: NA is replaced by empty string

txt
##                                           d1 
## "The cat in the hat ate green ham and eggs." 
##                                           d2 
##                                           "" 
##                                           d3 
##                          "Once upon a time." 
##                                           d4 
##                                           NA
corp
## Corpus consisting of 4 documents.
## d1 :
## "The cat in the hat ate green ham and eggs."
## 
## d2 :
## ""
## 
## d3 :
## "Once upon a time."
## 
## d4 :
## ""

corp %>%
  corpus_reshape(to = "sentences") %>%
  corpus_reshape(to = "documents")
## Corpus consisting of 2 documents.
## d1 :
## "The cat in the hat ate green ham and eggs."
## 
## d3 :
## "Once upon a time."

char_trim(txt, what = "documents", min_ntoken = 0)
## Warning: NA is replaced by empty string
##                                           d1 
## "The cat in the hat ate green ham and eggs." 
##                                           d2 
##                                           "" 
##                                           d3 
##                          "Once upon a time." 
##                                           d4 
##                                           ""
char_trim(txt, what = "documents", min_ntoken = 1)
## Warning: NA is replaced by empty string
[...]

PR #1980, opened 2020-07-03
Fix corpus_reshape()

koheiw:
Address multiple issues related to `corpus_reshape()`
- Stop removing empty documents (#1978)
- Expose stringi's `locale` option (#1979)
- Allow further reshaping segmented corpus

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: quanteda
-Version: 2.1.0
+Version: 2.1.0.9000
 Title: Quantitative Analysis of Textual Data
 Description: A fast, flexible, and comprehensive framework for 
     quantitative text analysis in R.  Provides functionality for corpus management,
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,11 @@
+# quanteda > 2.0.0
+
+## Changes
+
+## Bug fixes and stability enhancements
+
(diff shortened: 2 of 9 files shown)


koheiw:
Please also check this solves the underlying problem of #1976.

kbenoit:
It doesn't solve #1976 because of the `corpus_subset()` operation in `corpus_trim()`, but we do not want that to keep empty documents. So I think this is okay - but means we still need the fix I applied to `textstat_readability()`, which should never drop documents.

``` r
library("quanteda")
## Package version: 2.1.0

corp <- corpus(c(
  d1 = "The cat in the hat at green ham and eggs.",
  d2 = "",
  d3 = "Once upon a time."
))

corpus_trim(corp, min_ntoken = 0)
## Corpus consisting of 3 documents.
## d1 :
## "The cat in the hat at green ham and eggs."
## 
## d2 :
## ""
## 
## d3 :
## "Once upon a time."
corpus_trim(corp, min_ntoken = 1)
## Corpus consisting of 2 documents.
## d1 :
## "The cat in the hat at green ham and eggs."
## 
## d3 :
## "Once upon a time."
corpus_trim(corp, min_ntoken = 6)
## Corpus consisting of 1 document.
## d1 :
## "The cat in the hat at green ham and eggs."
```

PR #1986, opened 2020-07-17
Fix dictionary printing problem when max_nkey = 0,  max_nval = 0

kbenoit:
Fixes this problem:

```r
> print(data_dictionary_LSD2015, 0, 0)
Dictionary object with 4 key entries.
NULL
```

In the fix, it does not show the `NULL` which is returned by the early return of the function in `print_dictionary()`.


kbenoit:
I checked the other `print()` methods, none have this issue, so we're ok there.

Issue #1991, opened 2020-08-17
Unit test crash on Solaris 


koheiw:
There has been a mysterious but persistent errors on tests on Solaris ODS on CRAN and RHub. This does not happen on any other major OS but "Segmentation Fault" is worrying.

```r
* checking tests ...
  Running ‘spelling.R’
  Running ‘testthat.R’Segmentation Fault

 ERROR
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:
  34: withOneRestart(expr, restarts[[1L]])
  35: withRestarts(testthat_abort_reporter = function() NULL, force(code))
  36: with_reporter(reporter = reporter, start_end_reporter = start_end_reporter,     {        reporter$start_file(basename(path))        lister$start_file(basename(path))        source_file(path, new.env(parent = env), chdir = TRUE,             wrap = wrap)        reporter$.end_context()        reporter$end_file()    })
  37: FUN(X[[i]], ...)
  38: lapply(paths, test_file, env = env, reporter = current_reporter,     start_end_reporter = FALSE, load_helpers = FALSE, wrap = wrap)
  39: force(code)
  40: doWithOneRestart(return(expr), restart)
  41: withOneRestart(expr, restarts[[1L]])
  42: withRestarts(testthat_abort_reporter = function() NULL, force(code))
  43: with_reporter(reporter = current_reporter, results <- lapply(paths,     test_file, env = env, reporter = current_reporter, start_end_reporter = FALSE,     load_helpers = FALSE, wrap = wrap))
  44: test_files(paths, reporter = reporter, env = env, stop_on_failure = stop_on_failure,     stop_on_warning = stop_on_warning, wrap = wrap)
  45: test_dir(path = test_path, reporter = reporter, env = env, filter = filter,     ..., stop_on_failure = stop_on_failure, stop_on_warning = stop_on_warning,     wrap = wrap)
  46: test_package_dir(package = package, test_path = test_path, filter = filter,     reporter = reporter, ..., stop_on_failure = stop_on_failure,     stop_on_warning = stop_on_warning, wrap = wrap)
  47: test_check("quanteda.core")
  An irrecoverable exception occurred. R is aborting now ...
```

koheiw:
[...]

PR #1992, opened 2020-08-17
Activate caching on Solaris

koheiw:
See #1991


koheiw:
As far as I can tell based on my multiple tests on Rhub, it will not cause these errors, but please check yourself too.

```
  ── 2. Error: adding summary info works (@test-meta.R#126) ─────────────────────
  error in evaluating the argument 'x' in selecting a method for function 'as.list': error in evaluating the argument 'x' in selecting a method for function 'which': Start of codes indicating failure. (U_ILLEGAL_ARGUMENT_ERROR)
  Backtrace:
    1. quanteda:::add_summary_metadata(corp)
   19. base::.handleSimpleError(...)
   20. base:::h(simpleError(msg, call))
  
  ── 3. Error: adding extended summary information works (@test-meta.R#181) ─────
  error in evaluating the argument 'x' in selecting a method for function 'as.list': error in evaluating the argument 'x' in selecting a method for function 'which': Start of codes indicating failure. (U_ILLEGAL_ARGUMENT_ERROR)
  Backtrace:
    1. quanteda:::add_summary_metadata(corp, extended = TRUE)
   19. base::.handleSimpleError(...)
   20. base:::h(simpleError(msg, call))

```

We might not he only victim of a bug in some of the C++ libraries on Solaris. I think stringi is another victim rathen than the offender. In this case, we will continue to see errors in `test-nfunctions.R`.

PR #1994, opened 2020-08-19
Move CI to GitHub Actions

kbenoit:
- Sets up Actions to run R CMD check via the rcmdcheck package on the three major OSs (linux, macOS and Windows) on the release version of R and R-devel.

- Removes Travis and AppVeyor for CI


kbenoit:
@koheiw we're almost there in terms of being ready to mothball Appveyor and Travis and move to GHA, but take a look at the Ubuntu results. This looks to me like the locale setting affecting **stringi** - and something similar could be affecting the Solaris tests for the nfunctions. Is there a way to set this more stably?

koheiw:
It could be related to the ICU version. I don't have the latest because I am still on Ubuntu 18.

> 2020-08-19T10:28:36.8357601Z libicu-dev is already the newest version (66.1-2ubuntu2).

kbenoit:
I think it's exactly that. Can we put a conditional in the test to check this and skip the test if not? How can we check the ICU version from R? I remember you showed me this once before but I can't find it now. (anyway you're the expert on ICU.)

koheiw:
If the test is breaking on recent ICU, worth investigating. You might remember that the ICU boundary detection changed its handling of a sequence of the white space, and broke many tests while ago.

Also, we should print the installed version of ICU on start up to make users aware of different outputs.

PR #2003, opened 2020-09-26
Fix threads option

koheiw:
I noticed that the number of threads cannot be more than 2 (default value) because of the [change in the latest RcppParallel](https://cran.rstudio.com/web/packages/RcppParallel/NEWS). Therefore,

- Set default threads equal to defaultNumThreads()
- Stop calling setThreadOptions() when the value is the same as default
- Warn if users call threads option more than once and ignore


kbenoit:
This is all good **except**:

We cannot set `quanteda_options()$threads` to `defaultNumThreads()` because this will almost always be higher than 2. From https://cran.r-project.org/web/packages/policies.html#Source-packages:

> If running a package uses multiple threads/cores it must never use more than two simultaneously: the check farm is a shared resource and will typically be running many checks simultaneously.

You might remember, when preparing the 1.0 release, we had a CRAN submission rejected because we used to set half the threads as default, or defaultNumThreads - 1, but whatever it was it was > 2.

So we should put in the guidelines that it's better to set this in `.Rprofile`.

tests/testthat.R
```diff
@@ -5,8 +5,8 @@ library(quanteda)
 
 Sys.setenv("_R_CHECK_LENGTH_1_CONDITION_" = TRUE)
 
-ops <- quanteda_options()
 quanteda_options(reset = TRUE)
+#quanteda_options(threads = 2)
```
koheiw, line 9: I remember the problem, but it seems that other RcppParallel dependent packages are not doing anything to limit the number of threads. But we can set the number before running the tests if we really need.

kbenoit: The problem is that the examples run before the tests. So we'd need to do this before anything at all started to run, including vignettes.

koheiw: I think examples are fine, because they don't require a lot of computing power, but vignettes might not. How about including `quanteda_options(threads = 2)` in the hidden code chunk in every Rmd file?

koheiw: Also posted a quetsion https://github.com/RcppCore/RcppParallel/issues/110

koheiw:
Based on https://github.com/Rdatatable/data.table/issues/3300, I think it is best to respect `OMP_THREAD_LIMIT` which is set to 2 on CRAN.

koheiw:
Done. Please review and merge.

kbenoit (commented):
What does this set the threads to then, on:
- CRAN? (needs to be 2)
- user's computer ?

koheiw:
On CRAN it will be two based on the environment variable Open MP. On users' machines, it will be all the cores as it does not set any limit.

kbenoit (approved):
Using all of the user's cores is a wee bit greedy but hey.

koheiw:
It is the default behavior of RcppParallel (and probably Open MP), so it should be fine.

Issue #1959, opened 2020-05-22
Simplify summary()


koheiw:
We have  `textstat_summary()` now so it is the time to make the `summary.corpus()` as simple as `base::summary()`. We should also lead `add_summary_metadata()` to retire.

kbenoit:
I propose that each core object class have a summary method, that gives an overview of the object but not details of either the documents or the document variables.

Something like this:
```r
> summary(data_corpus_inaugural)
Corpus consisting of 58 documents and four docvars.

Title: US presidential inaugural address speeches
Docvars: Year<int>, President<char>, FirstNames<char>, Party<char>
Tokens: 148,770  Types: 46,684  Median document length: 2,315
```

To get a summary of the document variables, users would need to do this:
```r
docvars(data_corpus_inaugural) %>%
    summary()
```
or with **quanteda.tidy**:
```r
> library("quanteda.tidy")
> glimpse(data_corpus_inaugural)
Rows: 58
Columns: 6
$ doc_id    <chr> "1789-Washington", "1793-Washington", "1797-Adams", "1801-Jefferson", "1805-Jefferson…
$ text      <chr> "Fellow-Cit…", "Fellow cit…", "When it wa…", "Friends an…", "Proceeding…", "Unwilling…
$ Year      <int> 1789, 1793, 1797, 1801, 1805, 1809, 1813, 1817, 1821, 1825, 1829, 1833, 1837, 1841, 1…
$ President <chr> "Washington", "Washington", "Adams", "Jefferson", "Jefferson", "Madison", "Madison", …
$ FirstName <chr> "George", "George", "John", "Thomas", "Thomas", "James", "James", "James", "James", "…
$ Party     <fct> none, none, Federalist, Democratic-Republican, Democratic-Republican, Democratic-Repu…
```
(That works for `quanteda.tidy::glimpse()` because **quanteda.tidy** is all about docvars.)

kbenoit:
ps `add_summary_metadata()` is internal only but needed by some other applications, so we need to keep it.

koheiw:
How about this?
```r
> summary(quanteda:::get_docvars(data_corpus_inaugural, user = TRUE, system = TRUE))
   docname_                     docid_       segid_       Year     
 Length:58          1789-Washington: 1   Min.   :1   Min.   :1789  
 Class :character   1793-Washington: 1   1st Qu.:1   1st Qu.:1846  
 Mode  :character   1797-Adams     : 1   Median :1   Median :1903  
[...]

PR #2007, opened 2020-11-20
Separate textstat

kbenoit:
This is part of a multi-step process to modularise the package. 

- Create a branch called `modularise` from the master. This will be the branch that will work with the modularised packages, and from which we successively split off from the main branch.

- Implement the modularisation.
    - [**quanteda.textmodels**](https://github.com/quanteda/quanteda.textmodels): This is already split and already on CRAN. It might have its textplot function(s) moved, but is more or less fine. 
    - [**quanteda.textstats**](https://github.com/quanteda/quanteda.textstats): This is the package containing all textstat functions. We need to split it and then do a full set of revdep checks on the pared down **quanteda**, to see how much we have broken, and then fix these
    - **quanteda.textplots**: Once the textstats split is working, we can then perform the next split, moving the textplot functions to this new package. This will be slightly different in that it will Import all three other packages (**quanteda**, **quanteda.textmodels**, and **quanteda.textstats**) since it operates on objects from all three. But we should think carefully about this since it could be that we would like these functions to work on non-quanteda objects too.

- Submit these to CRAN as part of a careful update strategy. The big problem will be finding and fixing the use of `quanteda::movedfunction()` (e.g. `quanteda::textstat_collocations()` for instance, since it will have moved to another namespace.

- Fix the pkgdown documentation and vignettes. It would be nice to have all of these in one place, as now, but also make it clear that they are divided among different packages. See https://github.com/r-lib/pkgdown/issues/1251.

- Add more packages to the quanteda family, such as **quanteda.sentiment**, **quanteda.newsmap**, etc.

**Note:** @koheiw got this branch by merging the `modularise` branch from [**quanteda.core**](https://github.com/koheiw/quanteda.core/pull/5). Because it was a copy from **quanteda** I was able to merge it and then tidy it up. I restored any functions not part of the textstat separation - we should only separate one thing at a time. This included the textplot functions but also some others that had vanished for reasons unknown - `bootstrap_dfm()`, and other things we had recently added like `dictionary_edit()`, `char_select()`, etc.


R/char_select.R
```diff
@@ -44,7 +44,7 @@ char_select.character <- function(x, pattern, selection = c("keep", "remove"),
     
     if (is.dictionary(pattern))
         pattern <- unique(unlist(as.list(pattern), use.names = FALSE))
-    if (is.collocations(pattern))
+    if ("collocations" %in% class(pattern))
```
koheiw, line 47: I think it is better to have object checking functions in the core package at least as internals.

R/summary.R
```diff
@@ -25,7 +25,7 @@ summary.corpus <- function(object, n = 100, tolower = FALSE, showmeta = TRUE, ..
     ndoc_all <- ndoc(object)
     object <- head(object, n)
     ndoc_show <- ndoc(object)
-    result <- textstat_summary(object, cache = FALSE, tolower = tolower, ...)
+    result <- summarize(as.corpus(object), cache = FALSE, tolower = tolower, ...)
```
koheiw, line 28: Let's make summary to be like 
```
> summary(attr(data_corpus_inaugural, "docvars"))
   docname_                     docid_       segid_       Year     
 Length:58          1789-Washington: 1   Min.   :1   Min.   :1789  
 Class :character   1793-Washington: 1   1st Qu.:1   1st Qu.:1846  
 Mode  :character   1797-Adams     : 1   Median :1   Median :1903  
                    1801-Jefferson : 1   Mean   :1   Mean   :1903  
                    1805-Jefferson : 1   3rd Qu.:1   3rd Qu.:1960  
                    1809-Madison   : 1   Max.   :1   Max.   :2017  
                    (Other)        :52                             
  President          FirstName                           Party   
 Length:58          Length:58          Democratic           :21  
 Class :character   Class :character   Democratic-Republican: 7  
 Mode  :character   Mode  :character   Federalist           : 1  
                                       none                 : 2  
                                       Republican           :24  
                                       Whig                 : 3  
```

R/summary.R
```diff
@@ -59,4 +59,94 @@ print.summary.corpus <- function(x, ...) {
...
+        result$chars <- nchar(x)
+        result$sents <- ntoken(tokens(x, what = "sentence"))
+    }
+
+    if (cache)
+        set_cache(x, "summary", result, ...)
+    return(result)
+}
+
+get_cache <- function(x, field, ...) {
```
koheiw, line 114: Better to move them to `cache.R`

tests/testthat/test-tokens_compound.R
```diff
@@ -129,42 +129,6 @@ test_that("tokens_compound works as expected with nested and overlapping tokens"
     )
 })
 
-test_that("tokens_compound works as expected with collocations", {
```
koheiw, line 132: Better not delete. We can test `tokens_compound()` with a save collocation object.

tests/testthat/test-tokens_select.R
```diff
@@ -218,43 +218,43 @@ test_that("tokens_select works as expected for bigrams selected on char, list of
...
-    )
-    expect_equal(
-        as.list(tokens_select(toks_bi, coll_tri)),
-        list(d1 = character(), d2 = character())
-    )
-    expect_silent(
-        tokens_select(toks_bi, coll_tri)
-    )
-})
+# test_that("tokens_select works correctly with collocations objects", {
```
koheiw, line 221: We should test using saved collocation objects.

koheiw (changes requested):
Looks good overall. Only two major issues as commented on the code:

- We should keep the tests of `tokens_select()` and `tokens_compound()` with collocations using saved objects.
- We should really simplify the output of `summary().
-

R/char_select.R
kbenoit: Agreed. Added back in c66f617.

R/summary.R
kbenoit: The key for the separation is how to wean it from `textstat_summary()`. How we design the summary is better discussed and resolved via #1959.

R/summary.R
kbenoit: Done.

tests/testthat/test-tokens_compound.R
kbenoit: Agreed. Added back in 9e9aa94. I put them into a separate test file called `test-collocations.R` to make this easily distinct. (This had been in **quanteda.textstats** but I moved it out.)

tests/testthat/test-tokens_select.R
kbenoit: Added back in 9e9aa94.

R/summary.R
koheiw: Duplication is not separation: there are `quanteda::summarize()` and `quanteda.textstats::summarize()`. I am fine if this is only transitional.

koheiw (approved):

R/summary.R
kbenoit: Agreed. We will eliminate the duplication by redesigning `summary.corpus()`.

Issue #2008, opened 2020-11-21
kwic crashes with vector window


kbenoit:
This crashes R:

```r
library("quanteda")
kwic("This is a sample text.", "sample", window = c(1, 2))
```

koheiw:
It is because the C++ function expects single value. I did not checking the value because a vector just cause error on Linux and Windows. Isn't it the same on Mac?
```
> kwic(toks, "people", window = c(1,2))
 Error in qatd_cpp_kwic(x, attrs[["types"]], ids, seq_along(ids), window,  : 
  Expecting a single value: [extent=2]. 
```

kbenoit:
macOS Terminal R: 

```
R version 4.0.3 (2020-10-10) -- "Bunny-Wunnies Freak Out"
Copyright (C) 2020 The R Foundation for Statistical Computing
Platform: x86_64-apple-darwin17.0 (64-bit)

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

  Natural language support but running in an English locale

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

> library("quanteda")
Package version: 2.9.9000
Parallel computing: 12 of 12 threads used.
See https://quanteda.io for tutorials and examples.

Attaching package: ‘quanteda’

The following object is masked from ‘package:utils’:

    View

> kwic("This is a sample text.", "sample", window = c(1, 2))

 *** caught segfault ***
address 0x21, cause 'memory not mapped'

Traceback:
 1: qatd_cpp_kwic(x, attrs[["types"]], ids, seq_along(ids), window,     separator)
 2: kwic.tokens(tokens(x, what = "word", ...), pattern, window, valuetype,     separator, case_insensitive)
 3: kwic(tokens(x, what = "word", ...), pattern, window, valuetype,     separator, case_insensitive)
 4: kwic.corpus(corpus(x), pattern, window, valuetype, separator,     case_insensitive, ...)
 5: kwic(corpus(x), pattern, window, valuetype, separator, case_insensitive,     ...)
 6: kwic.character("This is a sample text.", "sample", window = c(1,     2))
 7: kwic("This is a sample text.", "sample", window = c(1, 2))

[...]

PR #2010, opened 2020-11-24
Address #2008

koheiw:
Address #2008

### DIFF
--- a/R/kwic.R
+++ b/R/kwic.R
@@ -103,7 +103,8 @@ kwic.tokens <- function(x, pattern, window = 5,
         names(pattern) <- pattern
 
     valuetype <- match.arg(valuetype)
-    window <- as.integer(window)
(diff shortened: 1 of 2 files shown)


R/kwic.R
```diff
@@ -103,7 +103,8 @@ kwic.tokens <- function(x, pattern, window = 5,
         names(pattern) <- pattern
 
     valuetype <- match.arg(valuetype)
-    window <- as.integer(window)
+    window <- as.integer(window[1])
```
kbenoit, line 106: Shouldn't we issue some warning, such as 

```r
'window' has length > 1 and only the first element will be used
```

kbenoit (approved):

kbenoit:
Once merged into `master`, we should merge this into the `modularise` branch as well, before deleting the branch.

PR #2013, opened 2020-11-28
Update recompile

koheiw:
Salvage the code from quanteda.core that has been archived. 

- Treat empty tokens as padding (L15) 
- Several improvements in coding


koheiw:
It is not good time to

PR #2014, opened 2020-11-29
Separate the textplot functions

kbenoit:
Removes the `textplot_()` functions, as these are being split into **quanteda.textplots**.


koheiw:
I think all the network related functions should be in the textplot package as they are used only in `textplot_nextwork()` and messy with many dependencies. It actually needs reengineering in near future.

kbenoit:
Agreed on both their appropriate location in the textplots package, done b73bb91.

Also agree that the network plot functions need an overhaul. And the wordcloud functions.

koheiw (approved):

PR #2019, opened 2020-12-08
Print Unicode and ICU versions

koheiw:
to make trouble shooting easier and make users aware of the differences. It also helps to address https://github.com/quanteda/quanteda.textstats/issues/24.

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -1,7 +1,9 @@
 .onAttach <- function(...) {
     # startup message
     #packageStartupMessage("Initializing package: ", sQuote('quanteda'))
-    packageStartupMessage("Package version: ", as.character(utils::packageVersion("quanteda")))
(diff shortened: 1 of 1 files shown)


kbenoit (approved):
Good change. I wonder if we should check a user preference that could turn on silent messages by default, if they wish.

koheiw:
Generally speaking, we should stop using `cat()` for messaging, because `suppressMessages()` does not work.

```
> suppressMessages(cat("aaaa"))
aaaa
> suppressMessages(message("aaaa"))
> 
```

kbenoit:
Totally right on `cat()` - that's why we wrote `catm()` which wraps `message()`. We should replace `cat()` with that if we have forgotten to use it anywhere.

koheiw:
Create an issue #2020

Issue #2023, opened 2020-12-14
Inconsistency in how size affect dfm_sample and corpus_sample


koheiw:
As I add input value checks, I noticed that the default value for `size` are `dfm_sample(x, size = ndoc(x))` but `corpus_sample(x, size = NULL)`. The easy solution seemed setting the default to `ndoc(x)` in both, but it was not because of sampling by group, that it is only possible in `corpus_sample()`. i.e. `size` is passed to `quanteda:::sample_bygroup()` that interprets it as `size` within each group, so the result will become multiple times of larger than `ndoc()`. 

https://github.com/quanteda/quanteda/blob/65b31c9473c3290af7a3ef2593c011fbe1dffbf9/R/corpus_sample.R#L46-L53

We need to add group to `dfm_sample()` and make their behaviors and default values for `size` the same. This is also related to `bootstrap_dfm()`.

https://github.com/quanteda/quanteda/blob/65b31c9473c3290af7a3ef2593c011fbe1dffbf9/R/bootstrap_dfm.R#L78-L80

kbenoit:
Excellent point. We need a way for the default to be the size of each group, or a single value, or a vector of length equal to the group sizes.

We probably want the default to be the vector `as.integer(table(groups))` but that's a bit ugly for an argument default.

koheiw:
OK, let's set `size` to be `as.integer(table(groups))` when NULL (default).

kbenoit:
That's a bit non-literal - since `NULL` actually means a vector of specific values - but we can explain this clearly in the documentation. We could also argue that it's appropriate since `groups = NULL` by default anyway, so this could not be the default when `groups = NULL`.

So OK but let's make sure we explain it very clearly.

PR #2025, opened 2020-12-15
Issue 2023

koheiw:
Address #2023


R/dfm_sample.R
```diff
@@ -5,11 +5,9 @@
 #' @param size a positive number, the number of documents or features to select.
 #'   The default is the number of documents or the number of features, for
 #'   `margin = "documents"` and `margin = "features"` respectively.
-#' @param replace logical; should sampling be with replacement?
-#' @param prob a vector of probability weights for obtaining the elements of the
-#'   vector being sampled.
 #' @param margin dimension (of a [dfm]) to sample: can be `documents` or
 #'   `features`
+#' @inheritParams corpus_sample
```
kbenoit, line 10: The addition of `by` is welcome but works very differently for `margin = "features"`, so we might need to think this through.

kbenoit (changes requested):
Looks good but we need to think through how the sampling works with the vectors and with `by` for feature dimensions of the dfm.

R/dfm_sample.R
koheiw: `by` has no effect for feature. I also think we don't need sampling for `features` as it is really easy: `dfmt[,sample(nfeat(dfmt), 1000)]`

kbenoit: OK, I think we could simplify this then and make it more consistent by defining the `*_sample()` as operating on documents only. Let's deprecate margin with a warning and remove it from the next release. I suspect it gets very little use.

kbenoit:
We should also make `tokens_sample()` consistent with the other two sample functions.

R/dfm_sample.R
koheiw: 👍

kbenoit (approved):

PR #2027, opened 2020-12-17
Check values

koheiw:
Address problems similar to #2008 globally. Let's see how much test coverage it will change...


kbenoit:
Nice work - and a lot of work! A few comments:
- This is all necessary only because of S3 and the fact that it checks nothing with regard to argument types. ☹️ 
- For some ranges, should we differentiate between "0 and Inf" when it requires integers versus continuous values?
- For some ranges, the max is actually fixed by conditions of the inputs, for instance, `dfm_weight()` has a max of the number of features when using rank types.

kbenoit:
Also it's worth exporting the utility functions so that the other quanteda packages can use them, plus writing something in the Wiki for how to validate inputs.

kbenoit (approved):
Looks great. 

Remember to close the issues that this solves, after merging.

Issue #2020, opened 2020-12-09
Improve messaging 


koheiw:
- [x] Replace `cat()` with `message()` (or shortcut to the function) to allow users to suppress messages using `suppressMessages()` (#2019)
- [x] Consolidate `unused_dots()` and `check_dots()` (#1877).
- [x] Centralize value checks (#2009) 
- [x] Rename `friendly_class_undefined_message()` to `check_class()`
- [x] Centralize common error messages using https://github.com/quanteda/quanteda/blob/24c78db958ca500c5a96ac9fb5a262a4b343c97d/R/utils.R#L228-L242

kbenoit:
Agree with all of these. But you will soon learn the joys of `revdep_check()`, since half of those functions are now defining functions in **quanteda.textstats** and **quanteda.textplots**. 😁 

I'm not sure we need to define all error messages in `message_error()`, just the ones that are reused.

koheiw:
I do not promote `message_error()` too hard, but it is usually good to separate logic and view (c.f. [MVC](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller)).

PR #2029, opened 2020-12-18
Check class

koheiw:
As part of #2020, 

- consolidate `check_dots()` and `unused_dots()`
- rename `friendly_class_undefined_message()` to `check_class()`

There are changed in those functions to make them more useful.


koheiw:
`convert()` raises annoying warnings because of the default value of `omit_empty`.

```
> convert(dfm("a b c"), "data.frame")
  doc_id a b c
1  text1 1 1 1
Warning message:
In convert.dfm(dfm("a b c"), "data.frame") :
  omit_empty not used for 'to = "data.frame"'
```

R/validator.R
```diff
@@ -117,3 +117,40 @@ check_length <- function(x, min_len, max_len, arg) {
...
+#' quanteda:::check_class("tokens", "dfm_select")
+#' }
+check_class <- function(class, method) {
+    class_valid <- rownames(attr(utils::methods(method), "info"))
+    class_valid <- stringi::stri_replace_first_fixed(class_valid, paste0(method, "."), "")
+    class_valid <- class_valid[class_valid != "default"]
+    if (!length(intersect(class, class_valid)))
+        stop(method, "() only works on ", paste(class_valid, collapse = ", "), " objects.", call. = FALSE)
+}
+friendly_class_undefined_message <- check_class # for compatibility
```
kbenoit, line 138: :+1:

but I will update the extension packages too to use the new functions

kbenoit (approved):
More great cleanup work!

Let’s add guidelines for the use of these functions by (quanteda) package developers to the Wiki.

I also can see in the individual changes list that now, codecov alerts lack of coverage for specific lines (which is great). There are few not covered, if you want to target them and add tests. 

Otherwise :+1: +1: +1:

kbenoit:
On the `convert()` warning why not raise a separate issue?

There are many misunderstandings about this that appear on StackOverflow (for instance), but we cannot convert empty documents to many of the topic model formats. 
E.g. https://stackoverflow.com/a/65241434/4158274

koheiw:
Opened separate issues #2030 and 2031.

PR #2033, opened 2020-12-21
Allow topfeatures() by group for weighted dfms

### LINKED ISSUE #2032: topfeatures(x, group = something) fails with weighted dfms
Discovered in https://stackoverflow.com/questions/59148580/whats-the-correct-way-to-extract-tf-idf-topfeatures-by-document/59152416?noredirect=1#comment115615693_59152416:

``` r
library("quanteda")
## Package version: 2.1.2

dfmat <- dfm(c("a b b b c c d", "a a b d d"))

topfeatures(dfmat, n = 1, groups = c("group1", "group2"))
## $group1
## b 
## 3 
## 
## $group2
## a 
## 2

topfeatures(dfm_tfidf(dfmat), n = 1, groups = c("group1", "group2"))
## Error: will not group a weighted dfm; use force = TRUE to override
```

Solution is to use `force = TRUE` by default.

kbenoit:
Fixes #2032 

But because topfeatures sums by group, and that may not be appropriate when they are weighted, it could be worth considering whether we want to allow this.


koheiw:
It is surprising that you are developing the function that I am proposing to remove. I don't think `topfeatures()` as a simple inspection function needs `group`.

kbenoit:
Even if we decide to phase it out, we will not want to do this suddenly, so might as well fix it in the meantime.

koheiw (approved):

PR #2034, opened 2020-12-24
Dev message

koheiw:
For #2020, add a messaging function. If this looks good, I will replace all the `cat()`, `catm()` and `warning()` with `info()`.

```r
> dfmt <- data_dfm_lbgexample
> quanteda:::info("dfm has %d %s and %d %s",
+      list(ndoc(dfmt), c("document", "documents"),
+           nfeat(dfmt), c("feature", "features")),
+      list(1, ndoc(dfmt) > 1,
+           1, nfeat(dfmt) > 1), warn = FALSE)
# dfm has 6 documents and 37 features

> dfmt <- data_dfm_lbgexample[1,]
> quanteda:::info("dfm has %d %s and %d %s",
+      list(ndoc(dfmt), c("document", "documents"),
+           nfeat(dfmt), c("feature", "features")),
+      list(1, ndoc(dfmt) > 1,
+           1, nfeat(dfmt) > 1), warn = FALSE)
# dfm has 1 document and 37 features
```
We can raise error i this function if it is useful.


kbenoit (commented):
These are from a developer's standpoint, since this is a PR aimed at developers: us, but presumably also others who would contribute or extend the package(s).

### General comments

Two general comments. 

First, to me at it seems overly complicated, especially for replacing `catm()` and `warning()`. (In fact, I always thought `catm()` was unnecessary - why not just use `message()` directly? Adam introduced this I think mainly as a quick way to replace `cat()`, which we'd incorrectly used years ago, with `message()`.) I have always preferred using the print, cat, or message with format() to produce messages, since these are more straightforward to me than sprintf(), which was included mainly as a gesture to C programmers.

I had to spend time examining the code and running some tests to figure out how it operates, which will be the same for anyone wanting to use this function as a developer. (`indices` is supposed to be a list of integers but takes a single-element logical in the example? Huh?) The base R functions, by contrast, will be familiar already to all developers.

It's also a choice of whether a specific function for a specific purpose is better than a single function with switches to make it perform different functionalities. I prefer the first, especially when these are already part of R.

Case in point: When there is an option such as `call.` (in `warning()`), I prefer the option to turn this on or off, not hardwired to `FALSE` as in the current function. In most cases, it's `TRUE` (the default), and I think we turned it off mostly when we have nested functions that would not be meaningful to the user. In most cases though it's useful to have the top-level called function referenced in the warning. So that suggests we'd need another argument, `call.` added to the general-purpose `info()` machinery. Ok, we can take out `call.` and let it be passed via `...` with the other arguments, but then I still need to know how `warning()` or `message()` works to control it by `...`. 

That's why I am a bit surprised by your approach here since it seems to make functionality more complicated rather than simplifying things as you often urge in other areas. Yes, it centralises our function calls, but it also complicates how the general one-call function must be used, and complication risks errors and misuse.

Here's an example of what I mean:

```r
> warning("Argument not used.")
Warning message:
Argument not used. 
> quanteda:::info("Argument not used.", values = list(), warn = TRUE)
Warning message:
Argument not used. 
```

I appreciate the intention of the PR but really, I strongly prefer the first. And I cannot decide whether to print the calling function.

Also for the `values` argument, for use with `warning()`, a very small minority of the cases I could find needed to print any values.

Second, it's hard to evaluate the PR without knowing exactly where it's intended to be used. If we think of this as a `catm()` replacement, then it's probably for messaging with `verbose = TRUE`, right? Some examples of the usefulness of the values and its formatting would be welcome.

In fact I'd see this as a very important part of the documentation for this function, since we'd be encouraging its use.

It should not replace `cat()` since that function is more appropriate than `message()` in print methods. I would propose we leave the print methods as they are. It's not always the most parsimonious code but it's readily and directly understandable. To make this one function work for `cat()` behaviour as well as `message()` and `warning()` would really make it complicated.


### Suggestions

1. I would not suggest using this to replace `warning()`. I prefer searching for "warning" in code if I want to find it, and prefer having full direct access to its arguments. And `warning()` almost never needs the `values()` argument as we have used it, so let's keep it simple.

2. For print methods, we should continue to use `cat()`. If we need functions to get the formatting right for adding the "," to thousands places, we can write a function for that. Same for the plurals. (Which needs plural for 0 as well as > 1 by the way.) Making the user achieve this behaviour through a combination of `values` and `indices` seems like a cure worse than the illness when it comes to `cat()`.

3. That leaves us with `message()` and replacing `catm()`. Here, I can see some benefits, but then we are talking about essentially upgrading `catm()`. So... why not literally just upgrade `catm()`? You can add `values = list(), 

    I could see this being used as a replacement for `catm()`, but see the next point. If you keep the arguments in `info()`, you could make it 

```r
catm(..., values = list(), indices = NULL, sep = " ", appendLF = FALSE)
```

### Recommendation

However, I have what I think is a better approach.

1. We go in the other direction, replacing `catm()` with `message()`. This means we are using the base R functions `cat()`, `catm()`, and `message()` where and as they are appropriate. No quanteda-specific wrappers that a developer needs to learn to get going; no assigning internal functions from extension packages to internal functions in the core **quanteda** package; etc.

2. We turn your code into a wrapper around a paste/sprintf function that can be used, if and only if needed, to standardise the printing of any part of an output that includes values. (When these are not needed, we would not need the function.) This could:
     - handle plurals correctly
     - handle number printing correctly, e.g. "features" in 
    ```r
    > print(dfm(data_corpus_inaugural), 0, 0)
    Document-feature matrix of: 58 documents, 9,360 features ...
    ```
    - Allow referencing your inputs as in the existing `info()`.  

    It would work this way:
   ```r
    message(info("tokens has %s", 
                   c("sentences", "paragraphs", "documents"), indices = 2))
    ```
    where `info()` returns character. When no `values` are used, `info()` can be used but is unnecessary since it just returns the input character.

    That way, we provide a utility to centralise our formatting - good! - but keep the three specific base R functions for their intended uses (in my view, also good).

kbenoit:
An added thought: Why did I write so much detail on what seems like a simple PR?

It's important to understand the thinking behind even small design decisions, since they shape bigger decisions later. My thinking here is to use existing functions when possible (only three discussed above), and to create new functions only when they serve specific purposes for our needs - like standardising the output formatting through a wrapper that returns `character()`.

This is similar to our motivation to use base R functions instead of importing tidyverse functions, for instance, even when the tidyverse functions are really convenient.

koheiw:
If am fine with the idea to just return character, because the main function of `info()` is conditional `sprintf()` to simplify message generation.

koheiw:
I am also happy to change the name to the function, but it should be short.

kbenoit:
OK thanks I feel more comfortable with this.

How about `msg()` for the function?

Should we also replace `catm()` with `message()`?

Doesn't have to be in this PR necessarily, but what about a way to format the numbers with a locale-specific thousands separator and decimal places? and something to get the plurals right, so that 0 and >1 are plural, but 1 is singular?

kbenoit (approved):
Looks good, see my latest comments.

The real issue will be where and when we use it of course but as a consistent and centralised message text generator this looks good.

koheiw:
`msg()` and `fmt()` are nicely short. As for large numbers, `sprintf()` does not support segmenting large numbers, so either pass `format(1000000, big.mark = ",")` as string to `%s` or pass the result to `prettyNum()` automatically. As far as the messages are English, we should use "," for big numbers.

```r
> prettyNum("dfm has 1000000000 features and 0.000001% dense", ",")
[1] "dfm has 1,000,000,000 features and 0.000001% dense"
```

Issue #2031, opened 2020-12-20
fcm.dfm() might be wrong


koheiw:
When I was trying to increase test coverage, I noticed something suspicious. Can anyone explain why the occurance of the-the is 0 in `fcmat1` and 2 in `fcmat2`?
``` r
require(quanteda)
#> Loading required package: quanteda
#> Package version: 2.9.9000
#> Unicode version: 13.0
#> ICU version: 66.1
#> Parallel computing: 6 of 6 threads used.
#> See https://quanteda.io for tutorials and examples.
#> 
#> Attaching package: 'quanteda'
#> The following object is masked from 'package:utils':
#> 
#>     View
txt <- c("The quick brown fox jumped over the lazy dog.",
         "The dog jumped and ate the fox.")
toks <- tokens(char_tolower(txt), remove_punct = TRUE)
dfmat <- dfm(toks)
fcmat1 <- fcm(dfm_weight(dfmat, scheme = "boolean"))
fcmat1
#> Feature co-occurrence matrix of: 10 by 10 features.
#>         features
#> features the quick brown fox jumped over lazy dog and ate
#>   the      0     1     1   2      2    1    1   2   1   1
#>   quick    0     0     1   1      1    1    1   1   0   0
#>   brown    0     0     0   1      1    1    1   1   0   0
#>   fox      0     0     0   0      2    1    1   2   1   1
#>   jumped   0     0     0   0      0    1    1   2   1   1
#>   over     0     0     0   0      0    0    1   1   0   0
#>   lazy     0     0     0   0      0    0    0   1   0   0
#>   dog      0     0     0   0      0    0    0   0   1   1
#>   and      0     0     0   0      0    0    0   0   0   1
#>   ate      0     0     0   0      0    0    0   0   0   0

fcmat2 <- fcm(dfmat, count = "boolean")
fcmat2
#> Feature co-occurrence matrix of: 10 by 10 features.
#>         features
#> features the quick brown fox jumped over lazy dog and ate
#>   the      2     1     1   2      2    1    1   2   1   1
#>   quick    0     0     1   1      1    1    1   1   0   0
#>   brown    0     0     0   1      1    1    1   1   0   0
#>   fox      0     0     0   0      2    1    1   2   1   1
#>   jumped   0     0     0   0      0    1    1   2   1   1
#>   over     0     0     0   0      0    0    1   1   0   0
#>   lazy     0     0     0   0      0    0    0   1   0   0
[...]

PR #2035, opened 2020-12-26
Upgrade FCM

koheiw:
In relation to #2031 

-  Simplify construction of FCM
-  Modernize the FCM object and methods (i.e. move S4 slots to object meta)


R/fcm.R
```diff
@@ -141,57 +140,50 @@ fcm.dfm <- function(x, context = c("document", "window"),
...
-    # discard the lower diagonal if tri == TRUE
-    if (tri) result <- triu(result)
-
-    # create a new feature context matrix
-    result <- new("fcm", as(result, "dgCMatrix"), count = count,
-                  context = context, margin = margin, weights = 1,
-                  tri = tri)
-    set_fcm_slots(result) <- attributes(x)
-    set_fcm_dimnames(result) <- list(rownames(result), colnames(result))
+    result <- build_fcm(
```
kbenoit, line 176: I see how this is consistent now with the other constructors, but I feel they could use more documentation. For instance here you are passing the attributes that were formerly slots, to become object metadata. The use of `...` however is not very clear in the documentation.

The more we document, the more I - and you, in six months! - will have to scratch our heads trying to figure out how these build, rebuild, and upgrade functions work.

Last week for instance I was trying to figure out how quickly to construct a tokens object from an externally passed list of integers and the vocabulary ("types") object. I could call the `build_tokens()` but it also needed a very specific kind of docvars with the `docname_`, `docid_` columns etc. I figured it out only through reading the code and a bit of trial and error.

Last week I

kbenoit (approved):
Makes sense, this is more consistent with how we have simplified the slots for dfm and used object metadata for what were formerly a variable set of slots. This is easier to maintain and extend.

R/meta.R
```diff
@@ -308,6 +310,24 @@ make_meta_dfm <- function(inherit = NULL, ...) {
     update_meta(default, inherit, ...)
 }
 
+#' @rdname make_meta
+make_meta_fcm <- function(inherit = NULL, ...) {
+    if (is.null(inherit))
+        inherit <- list()
+    default <- list(
+        "unit" = "documents",
```
koheiw, line 318: What is your thought on object meta inherited from DFM or tokens? Currently only `unit` and `concatenator`?

kbenoit: Sounds right. Maybe `what` as well?

R/object-builder.R
```diff
@@ -243,7 +243,7 @@ upgrade_fcm <- function(x) {
     if (!is_pre2(x)) return(x)
     attrs <- attributes(x)
     build_fcm(
-        x, colnames(x),
+        x, rownames(x), colnames(x),
```
kbenoit, line 246: Does it make sense to have a non-symmetric fcm? I can't think of how it does.

koheiw: Non-square FCM can exists, so just to avoid error
```
> fcmt[1,1:10]
Feature co-occurrence matrix of: 1 by 10 feature.
        features
features fellow - citizens  of the senate and house representatives :
  fellow      3 6       16 224 361      3 146     6               6 4
```

R/meta.R
koheiw: Added

tests/testthat/test-fcm.R
```diff
@@ -40,6 +40,16 @@ test_that("fcm works with character and tokens in the same way", {
                       round(as.matrix(fcmt_toks), 2))
 })
 
+test_that("fcm works with dfm and tokens in the same way", {
+    skip("needs to be investigated")
```
koheiw, line 44: Diagonal elements do not match. Is this how it should?

R/fcm.R
koheiw: Improved documentation.

tests/testthat/test-fcm.R
kbenoit: That's a good question. The difference is not between how `fcm()` operates on a dfm versus a tokens object, but rather how `context = "document"` operates versus `context = "window"`. With the window counting, each token is counted individually to see what other tokens co-occur within its window. 

``` r
library("quanteda")
## Package version: 2.9.9000

toks <- tokens(c(d1 = "b a b c"))

fcm(toks)
## Feature co-occurrence matrix of: 3 by 3 features.
##         features
## features b a c
##        b 1 2 2
##        a 0 0 1
##        c 0 0 0
fcm(dfm(toks))
## Feature co-occurrence matrix of: 3 by 3 features.
##         features
## features b a c
##        b 1 2 2
##        a 0 0 1
##        c 0 0 0

fcm(toks, context = "window", window = 2)
## Feature co-occurrence matrix of: 3 by 3 features.
##         features
## features b a c
##        b 2 2 1
##        a 0 0 1
##        c 0 0 0
```

I think that the behaviour is correct, since the context is equivalent to splitting up a tokens stream into the equivalent of individual "documents", one document per context window. So in that case, "b" co-occurs with "b" in this simple example twice, first in the "NULL NULL **b** a b" window and second in the "b a **b** c NULL" window. If you add the argument `ordered = TRUE` then it gets counted only once, and the diagonals will match that from the context = "document".

kbenoit: So the test that is skipped should be changed to either:
* changing the context of the tokens fcm:
    ```r
    `fcmt_toks <- fcm(toks, context = "document")`
    ```
    or

* just testing the equality of the diagonals, from 
    ```r
    fcmt_toks <- fcm(toks, context = "window", window = 1000, ordered = TRUE)
    ```

koheiw: This is the output. The sum of upper and lower triangles equals to `fcm_dfm`, but diagonal can be 2 here, because this "directed" in network analysis terminology, so a₁ a₂ is different from a₂ a₁. 
```r
> fcm(toks, context = "window", window = 1000, ordered = TRUE)
Feature co-occurrence matrix of: 6 by 6 features.
        features
features b a c e f g
       b 1 1 2 1 0 0
       a 3 1 4 3 1 1
       c 1 0 0 2 1 1
       e 0 0 0 0 1 1
       f 0 0 0 0 0 1
       g 0 0 0 0 0 0
```

kbenoit:
@koheiw are we done with this PR? It's approved so might as well merge, unless we are waiting on agreement to some question. I think we are not.

tests/testthat/test-fcm.R
koheiw: Let's solve this puzzle.

kbenoit: Consider just the first document, for simplicity.
```r
txt <- "b a b c"
```

* For `context = "document"`, the count is 1 for (b, b) since they co-occur just once, in one document.

    The window context works diffferently, since we are crawling through the text and considering each token and the co-occurrences within that token's window. 

* For `context = "window", ordered = TRUE`, the first b co-occurs with the second, but because the count is forward only, there the second b does not co-occur with the first. So:
    - for an ordered co-occurrence of "b a b c", we have (b, a), (b, b), (b, c), (a, b), (a, c), (b, c).
    - if this were "z a b c", we would have (z, a), (z, b), (z, c), (a, b), (a, c), (b, c) - the same total co-occurrences, just substituting. There is still no (b, z) that would have provided a second co-occurrence had the z been b as above.

    ```r
    > fcm(tokens("b a b c"), context = "window", ordered = TRUE)
    Feature co-occurrence matrix of: 3 by 3 features.
            features
    features b a c
           b 1 1 2
           a 1 0 1
           c 0 0 0
    ```

- for `context = "window", ordered = FALSE`, the idea is that the first b co-occurs with the second, and the second with the first, so the diagonal co-occurrence is 2. So for "b a b c", we count the co-occurrence of b with b as 2.
    - b1: a, b2, c
    - a: b1, b2, c
    - b2: b1, a, c
    - c: b1, a,  b2

    Collapsing b1 and b2, we see that the co-occurrence for b should be 2, 2, 2 as it is below.
    ```r
    > fcm(tokens("b a b c"), context = "window", ordered = FALSE)
    Feature co-occurrence matrix of: 3 by 3 features.
            features
    features b a c
           b 2 2 2
           a 0 0 1
           c 0 0 0
    ```

Mystery solved?

koheiw: I thought the ordered output records co-occurrences in forward (upper-triangle) and backward (lower-triangle). But current behavior should be correct if it is to record the frequency of ngrams. The current `fcm()` matches `tokens_ngrams()`:

```r
require(quanteda)
txt <- c("b a b c", "a a c b e", "a c e f g")
toks <- tokens(txt)

fcmt_toks <- fcm(toks, context = "window", window = 1000, ordered = TRUE)
sum(fcmt_toks) # 26

(toks_ng <- tokens_ngrams(toks, skip = 0:100))
sum(ntoken(toks_ng)) #26
```

This is more about documentation than the code...

```
#' @param ordered if `TRUE` the number of times that a term appears before
#'   or after the target feature are counted separately. Only makes sense for
```

kbenoit: It's not forward + backward, but rather the (i, j) cell address records how many forward co-occurrences of j occur for the target i.  That's why `tri` is disabled when `ordered = TRUE`. How about we change the param explanation to:

```
@param ordered if `TRUE`, count only the forward co-occurrences for each target token, 
  so that the `i, j` cell of the co-occurrence matrix is the number of times that token `j` 
  co-occurs with the target token `i` within the window. Only makes sense for 
  `context = "window"`, and when `ordered = TRUE`, the argument `tri` has no effect.
```

koheiw: I think mentioning bigram helps to understand. How about this?
```
@param ordered if `TRUE`, count only the forward co-occurrences for each target token for bigram models, 
  so that the `i, j` cell of FCM is the number of times that token `j` 
  occurs after the target token `i` within the window. Only makes sense for 
  `context = "window"`, and when `ordered = TRUE`, the argument `tri` has no effect.
```

PR #2036, opened 2020-12-29
Dev print dfm

koheiw:
Modify `msg()` and use in `dfm_pring()` for #2020


tests/testthat/test-dfm.R
```diff
@@ -803,7 +803,8 @@ test_that("test null dfm is handled properly", {
     expect_equal(rbind(mx, mx), mx)
     expect_equal(cbind(mx, mx), mx)
 
-    expect_output(print(mx), "Document-feature matrix of: 0 documents, 0 features.")
+    expect_output(print(mx), 
+                  "Document-feature matrix of: 0 document, 0 feature (0.00% sparse) and 0 docvar.", fixed = TRUE)
```
kbenoit, line 807: 0 of anything needs to be plural (documents, features, or docvars).

tests/testthat/test-message.R
```diff
@@ -8,6 +8,13 @@ test_that("msg works", {
         "there are 3 documents"
     )
     
+    # NULL indices
+    expect_equal(
+        quanteda:::msg("there are %d %s", list(3, "documents"),
+                       indecies = list(NULL, NULL)),
```
kbenoit, line 14: third argument should be `indices`. Let's add the `check_dots()` to all functions that take `...`. `msg()` does not call that...

kbenoit (changes requested):
Good consolidation of this function. But
- make 0 counts plural
- it uncovered an omission in `msg()`

tests/testthat/test-message.R
koheiw: Let's not use `check_dots()` in internal functions.

kbenoit (approved):
Sorry, I hadn't seen the changes, should have approved this 2 weeks ago!

PR #2040, opened 2021-01-25
Update inaugural

kbenoit:
Add Biden 2021 address, and update the scripts for updating this corpus.


kbenoit:
Looks like a config error with the macOS checks related to missing fortran, strange that this was not happening earlier. Probably some result of updating macOS on the GitHub system.

PR #2041, opened 2021-01-30
Stop omit_empty warnings

kbenoit:
These happened even when the argument was not supplied, because `check_logical()` created a value even when this value was missing.

### DIFF
--- a/man/pattern2list.Rd
+++ b/man/pattern2list.Rd
@@ -25,7 +25,8 @@ exact matching. See \link{valuetype} for details.}
 \item{case_insensitive}{logical; if \code{TRUE}, ignore case when matching a
 \code{pattern} or \link{dictionary} values}
 
-\item{concatenator}{concatenator that join multi-word expression in tokens object}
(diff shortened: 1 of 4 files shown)


man/pattern2list.Rd
```diff
@@ -42,17 +43,18 @@ Function to assign multiple slots to a S4 object
 \code{\link[=pattern2id]{pattern2id()}}
 }
 \keyword{Convert}
+\keyword{Used}
+\keyword{[kwic].}
+\keyword{[tokens_compound],}
+\keyword{[tokens_select],}
```
kbenoit, line 49: This is causing one of the R-CMD-CHECK tests to fail on CI. But it should not be generating these keywords at all. Something is wrong with the Rd here and I think it has to do with the undocumented arguments. I'm not sure why the CRAN tests are not showing that several parameters are undocumented, but I think if you document them, it will fix this, and then the last test will pass.

koheiw (approved):

PR #2043, opened 2021-02-01
Issue 2042

koheiw:
Address #2041, but I did not separate `pattern2id()` to minimize impact on existing functions and  make `object2id()` a one-stop converter.


R/object2fixed.R
```diff
@@ -0,0 +1,93 @@
...
+    
+    types <- check_character(types, min_len = 0, max_len = Inf, strict = TRUE)
+    valuetype <- match.arg(valuetype)
+    case_insensitive <- check_logical(case_insensitive)
+    concatenator <- check_character(concatenator)
+    levels <- check_integer(levels, min = 1, max_len = Inf)
+    remove_unigram <- check_logical(remove_unigram)
+    keep_nomatch <- check_logical(keep_nomatch)
+    
+    if (is.collocations(x)) {
```
kbenoit, line 32: Do you think this will be easier to maintain than using method dispatch for specific object classes of `x`?

R/utils.R
```diff
@@ -55,63 +55,6 @@ NULL
...
-#' Convert various input as pattern to a vector used in tokens_select,
-#' tokens_compound and kwic.
-#' @inheritParams pattern
-#' @inheritParams valuetype
-#' @param concatenator concatenator that join multi-word expression in tokens object
-#' @param levels only used when pattern is a dictionary
-#' @param remove_unigram ignore single-word patterns if `TRUE`
-#' @seealso [pattern2id()]
-#' @keywords internal
-pattern2list <- function(x, types, valuetype, case_insensitive,
```
kbenoit, line 67: Can you create a temporary internal function that maps to the old one? We need this to work for the first version at least:
https://github.com/quanteda/quanteda.textstats/blob/4a7ee5fdf5534b814585379125ceadf87b79c20c/tests/testthat/test-utils.R#L10

kbenoit (changes requested):
Looks generally good, but please resolve the conflicts from merging #2041 and see my comments.

R/utils.R
kbenoit: Added in 1ab1816

kbenoit (approved):

kbenoit:
@koheiw I thought this was ready to go, but in 1ab1816 I added a mapping to `pattern2id()`, but now see that that function has its own file. Was this because of a merge from master? Please check this, verify or resolve, then merge. Thanks!

Issue #1988, opened 2020-08-02
Add drop_docid = FALSE to *_subset() 


koheiw:
There should be a way to keep all the original docid even when documents are subset. It is very useful when `dfm_group(fill = TRUE)` is applied afterwards. This might solve #1978 as well.

Currently,
```r
> require(quanteda)
> dfmt <- dfm(data_corpus_inaugural)
> dfmt2 <- dfm_subset(dfmt, Party == "Democratic")
> 
> levels(docid(dfmt))
 [1] "1789-Washington" "1793-Washington" "1797-Adams"      "1801-Jefferson"  "1805-Jefferson"  "1809-Madison"   
 [7] "1813-Madison"    "1817-Monroe"     "1821-Monroe"     "1825-Adams"      "1829-Jackson"    "1833-Jackson"   
[13] "1837-VanBuren"   "1841-Harrison"   "1845-Polk"       "1849-Taylor"     "1853-Pierce"     "1857-Buchanan"  
[19] "1861-Lincoln"    "1865-Lincoln"    "1869-Grant"      "1873-Grant"      "1877-Hayes"      "1881-Garfield"  
[25] "1885-Cleveland"  "1889-Harrison"   "1893-Cleveland"  "1897-McKinley"   "1901-McKinley"   "1905-Roosevelt" 
[31] "1909-Taft"       "1913-Wilson"     "1917-Wilson"     "1921-Harding"    "1925-Coolidge"   "1929-Hoover"    
[37] "1933-Roosevelt"  "1937-Roosevelt"  "1941-Roosevelt"  "1945-Roosevelt"  "1949-Truman"     "1953-Eisenhower"
[43] "1957-Eisenhower" "1961-Kennedy"    "1965-Johnson"    "1969-Nixon"      "1973-Nixon"      "1977-Carter"    
[49] "1981-Reagan"     "1985-Reagan"     "1989-Bush"       "1993-Clinton"    "1997-Clinton"    "2001-Bush"      
[55] "2005-Bush"       "2009-Obama"      "2013-Obama"      "2017-Trump"     
> levels(docid(dfmt2))
 [1] "1829-Jackson"   "1833-Jackson"   "1837-VanBuren"  "1853-Pierce"    "1857-Buchanan"  "1885-Cleveland"
 [7] "1893-Cleveland" "1913-Wilson"    "1917-Wilson"    "1933-Roosevelt" "1937-Roosevelt" "1941-Roosevelt"
[...]

PR #2044, opened 2021-02-01
Issue 1988

koheiw:
Address #1988 (and #1978 next). The new argument is `drop_docid` but we could simply name that `drop` as in the `[` generics.


kbenoit (commented):
Looks good, but two questions/suggestions.

First, I think of `docid()` and the quantity that it returns as internal-only. I think we will confuse users if we appear to be suggesting that they use it. It makes more sense to document this only for ourselves and for (other) developers. So to describe this in the simple way we have in the documentation could confuse users.

Second, would we break many or any existing behaviours if this were `FALSE` by default, or even if it were just automatically `FALSE`, with no option? If we want to recompile them, we could do so through corpus.corpus() or tokens.tokens() etc, to rebuild it. Does this cause any harm not to be reindexed?

koheiw:
I prefer to set `drop_docid = FALSE` but it is `TRUE` only to avoid changing current behavior. The change would affect users who use `dfm_group(fill = TRUE)` a lot. 

In many cases `docnames()` is enough, but `docid()` is necessary to get (non-unique) original names of documents, especially when texts are reshaped or created with `corpus(unique_docnames = FALSE)`. The name `docid_field` should be `docnames_field` to avoid confusion.

kbenoit:
OK, let's use the `drop_docid = TRUE` to avoid breaking current behaviour, but document this clearly in `dfm_group()` help page.

You make a very good point about `docid_field`. But I thought that reason we did not use `docname_field` is that if we imported a set of sentences spread across a smaller set of documents, then these would be `docid` values, and docnames would be made unique, in the same way as `corpus_reshape()` does it. (All assuming that `unique_docnames = FALSE` of course.)

kbenoit (approved):

Issue #2038, opened 2021-01-14
A function to change windows of quanteda functions


hungnguyen167:
Dear developers,

I am wondering whether there is a better way to change the window argument of a quanteda function with the same pattern without having to go through the whole process again, which sometimes takes a lot of time. For example:
```{r}
## Get kwic() of a random corpus 
kwic_1 <- kwic(corpus_example, pattern = "[\\p{Lu}][\\p{Ll}]+:", window = 10).

## Now imagine I would like to use kwic() on the same corpus, using the same pattern, just with a different window. The new function would be
window_change(kwic_1, new_window = 20)
```
I believe it would be much more resource-efficient if one could develop a function like this, especially for heavy tasks like performing kwic(), tokens_keep() or tokens_lookup() on hundreds of thousands of documents. Since kwic() or tokens_keep() store information about the positions of matched phrases/keywords, it would be easier to tell R to get back to these specific positions and simply expand/contract the window than having to do the whole tokenization/kwic process again. I can't speak for everyone but I'm sure having a function like this would help me speed up the research process tremendously.

I would love to hear your thoughts on this. Thank you in advance!

Sincerely,
Hung.

koheiw:
Thanks for the suggestion. I am planing to [redesigning kwic() functions](https://github.com/quanteda/quanteda/issues/1840). It is a good idea to move `window` to a printing function.

kbenoit:
Here's an easy way: we define a `kwic.kwic()` function that can take a `window` value <= the input `kwic`, and it just compresses the pre and post by trimming it.

Issue #1840, opened 2020-01-14
How to predict in advance a kwic search that will take a very long time?


kbenoit:
What would be reasonable limits on what we allow a user to ask for the pattern matching functions? It appears to be an issue mainly in the number of patterns. Below, the number of patterns in `stopwords("en")` is 175, so 160/8.5 = 18.8 is not bad, but it's till 2.5 minutes of frozen console.

Maybe we could issue a warning and ask to proceed if the user tries to fit more than a vector of 10 patterns? We could set this limit and the prompt with warning as global options.

``` r
library(quanteda)
## Package version: 1.9.9007

data_corpus_guardian <- as.corpus(quanteda.corpora::download("data_corpus_guardian"))

# ndoc = 6000
system.time(
  kw1 <- kwic(data_corpus_guardian, "the")
)
##    user  system elapsed 
##   8.473   0.445   8.615

# compare with longer documents
data_corpus_guardiansents <- corpus_reshape(data_corpus_guardian, to = "sentences")
system.time(
  kw3 <- kwic(data_corpus_guardiansents, "the")
)
##    user  system elapsed 
##  12.248   0.388  11.803

# with shorter documents
data_corpus_guardianonebig <-
  corpus(texts(data_corpus_guardian, groups = rep(1, ndoc(data_corpus_guardian))))
system.time(
  kw4 <- kwic(data_corpus_guardianonebig, "the")
)
##    user  system elapsed 
##   6.233   0.377   6.653

# not really ok
system.time(
  kw2 <- kwic(data_corpus_guardian, stopwords("en"))
)
##    user  system elapsed 
## 160.170   5.907 125.154
```

koheiw:
It takes a lot of time to construct data.frame actually. When it is too large for human inspection, isn't it nice to add something like pagination?

kbenoit:
The view of the data.frame using DT already does this, e.g. `View.kwic()`. Did you have something else in mind?

koheiw:
DT has pagination, but it is still a large kwic data.frame with all rows. My idea is to `qatd_cpp_kwic()` to return from i to i + n rows of the data.frame. Here `len` is the total number of rows in the data.frame. The following section (from L101) is data.frame construction, which takes looooong time.

https://github.com/quanteda/quanteda/blob/fca739904c6c6144b8a95903cf231937170bbd26/src/kwic_mt.cpp#L86-L99

We can even split below and above L101 into different functions that pass intermediate kwic objects.

[...]

PR #2045, opened 2021-02-02
Issue 1840

koheiw:
Separated text search (`kwic`) and concordance view generation `print.kwic()` to address #2038 and #1840. This patch needs final adjustment, but difficult part is already done.


R/corpus.R
```diff
@@ -234,54 +230,6 @@ corpus.data.frame <- function(x, docid_field = "doc_id", text_field = "text",
...
-#' @examples
-#' # from a kwic
-#' kw <- kwic(data_char_sampletext, "econom*", separator = "",
-#'            remove_separators = FALSE) # keep original separators
-#' summary(corpus(kw))
-#' summary(corpus(kw, split_context = FALSE))
-#' texts(corpus(kw, split_context = FALSE))
-#'
-#' @export
-corpus.kwic <- function(x, split_context = TRUE, extract_keyword = TRUE, meta = list(), ...) {
```
kbenoit, line 254: Why delete this method for `corpus()`? A lot of users have found this useful for selecting contexts to create a new corpus. In fact I thought you used it for this purpose!

i.e.
https://stackoverflow.com/questions/37445449/how-do-i-transform-a-kwic-from-quanteda-package-to-a-corpus
https://stackoverflow.com/questions/52883002/selecting-sentences-that-contain-specific-words/52883578#52883578

koheiw: There are two reasons. 
1. Users should use `tokens_select(window)` to extract contexts. 
2. User can still use `corpus.data.frame()` with `as.data.frame(kw)`.

kbenoit: Perhaps deprecating it first? The corpus.kwic() has some options that those other methods don't, such as split_context and extract_keyword.

koheiw: split_context and extract_keyword can be done with data.frame manually. Demands for `corpus.kwic` is another case for separating the core functions to quanteda.core from quanteda. In this way, quanteda can keep miscellaneous short-cut functions for beginners.

kbenoit: Removing this method is not necessary to this PR and should not be part of the PR.

And I think removing it is a bad idea, for several reasons. 

1. Removing it will break existing code unnecessarily.

2. It's not complicating anything else through its existence - nothing else relies on it.

3. Lots of useful actions can be done "manually" using sequences of lower-level steps, but why remove a convenient function like corpus.kwic() that makes it simple and robust? This is super convenient, especially if combined with a way to create a corpus of sentences selected on specific terms. e.g.

```r
> kw <- kwic(data_char_sampletext, "econom*", separator = "", 
+            remove_separators = FALSE)
> 
> corpus(kw, split_context = FALSE)
Corpus consisting of 5 documents and 1 docvar.
text1.L313 :
" the Irish economy in pursuit "

text1.L390 :
" the domestic economy? As we"

text1.L516 :
" the domestic economy show the "

text1.L941 :
" dislocates the economy. Otherwise those"

text1.L976 :
" the domestic economy, stimulating demand"
```

koheiw: Why don't you separate low-level functions to quanteda.core? I have no time or motivation to maintain high-level functions that I do not use (or recommend others to use). If you don't like the name (quanteda.core), I can come up with something very different.

kbenoit: Maybe we should name it quanteda.functionskoheidislikes 😄

koheiw: Then I will abandon quanteda and make a new package.

kbenoit: My preference would be to make corpus.kwic internal and keep it working with a deprecation warning. It's not doing much harm and the short typical window means converting tokens to a text is not too damaging to performance, despite the philosophical offence.

kbenoit:
Overall my sense is that if the kwic object now includes the full set of tokens as an attribute. Isn't this pretty wasteful in terms of space?

There are ways to minimise the break with previous behaviours, by letting the user specify `separator` and `window` in `kwic()`, and passing these through via attributes to `print.kwic()` and `as.data.frame.kwic()` as defaults.

`print.kwic()` needs to be documented.

`View.kwic()` is no longer functioning correctly.

And see my comment above about `corpus.kwic()`.

kbenoit:
Thinking about this more, the PR essentially builds a token position index for the search terms, and includes the entire token set. This is like a single-use indexing, but means that with two kwic objects - common for instance in our tutorials and examples where these are plotted together - each kwic object is greater than the original tokens object. This is just one of the features of the functional, pass-by-value approach of R and quanteda. In another language, we might attach searched terms to the tokens object, and make these available, but that's very different from our existing approach in the package.

This could be solved if we returned to an idea we discussed early on, which is that a corpus could be processed to include tokens and indexes to all types in the document. Then processing a kwic would be fast since it would just find the indexed tokens and extract their context. It would be slow, but only once, when the corpus is created (or modified) or optionally, processed following creation of the existing type of corpus we have. And "creating" a tokens object would then not involve the tokenisation.

koheiw:
The tokens object is a subset of original:
```r
attr(result, "tokens") <- x[unique(result$docname)]
```
As far as I know, R does not make copy unless changes are made, so attaching the original `x` would actually save RAM space.

kbenoit:
Looks like it does make a full copy, see below.

Is there a performance gain by copying the full tokens for documents with a match?

``` r
library("quanteda")
## Package version: 2.9.9000
## Unicode version: 10.0
## ICU version: 61.1
## 
## 
##     View
library("pryr")
## Registered S3 method overwritten by 'pryr':
##   method      from
##   print.bytes Rcpp

object_size(toks)
## 1.32 MB

kw <- kwic(toks, "a")
object_size(kw)
## 1.37 MB
object_size(attr(kw, "tokens"))
## 1.32 MB

object_size(toks, kw)
## 2.06 MB
```

<sup>Created on 2021-02-10 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

koheiw:
If full tokens is attached to kwic, the attribute only point to the original. If tokens is sliced, new object is created. Unless kwic object is saved in RDS, the former approach is more efficient.

```r
> require(pryr)
> toks <- tokens(data_corpus_inaugural)
> kw <- kwic(toks, "a")
> 
> address(toks)
[1] "0x5605cfe02630"
> attr(kw, "tokens") <- toks
> toks_out <- attr(kw, "tokens")
> address(toks_out)
[1] "0x5605cfe02630"
> 
> attr(kw, "tokens") <- toks[1:10]
> toks_out2 <- attr(kw, "tokens")
> address(toks_out2)
[1] "0x5605d1711988"
```

kbenoit:
OK, looks like this is done, so here are my comments on the current PR.

Overall, this looks fine, I can see the usefulness, although it's unclear how much of a performance difference this creates. To the extent that `as.data.frame()` on a kwic object returns what used to be the main output object, I think this is fine, and it gives us more flexibility to redefine the kwic structure further if we want, since we are no longer required to maintain this as a data.frame object, but can produce that through coercion. This is a good move and consistent with our approach elsewhere.

However this breaks too much, including our own code, much user code, and one of our packages. We can virtually eliminate this incompatibility however through some small changes.

1. Add back `window` and `separator` to `kwic()`. These won't do anything in the object but can be passed to `print.kwic()` and `as.data.frame.kwic()` as defaults, with the print methods being able to override them, so we get the best of this PR's new approach but compatibility with and the convenience of the former approach. This means we don't have to change the examples either. (The `print()` looks unwieldy and reminds me too much of Python.)

2. This broke `textplots::textplot_xray()`, but I've already managed to fix this in a way that works for the pre-v3 kwic objects as well as the v3 kwic objects. But as a note, we always need to think about what we are breaking in our own quanteda-verse packages when we change the core package. 

3. I'd like to fix and undelete `corpus.kwic()`, which I don't see as doing any harm. If we decide to sentence it to death then it should at least be deprecated and made internal before removing it.

koheiw:
I am OK with restoring corpus.kwic() for deprecation and restore the arguments. I would be nice to add `max_nrow` to `print.kwic()`. I will leave these with you. have done my part on this branch.

kbenoit:
OK updates done, with some small improvements and stronger tests. `kwic()` now works exactly as before, from the user's standpoint, but with the new machinery. (See the added tests for details.) `window` and `separator` can be specified upon creation but can be changed in either `print.kwic()` or `as.data.frame.kwic()`.

I'm happy to add `max_nrow` to kwic. This would make it work as do our other print methods, but those by default are pretty 
conservative. We'd probably want to decide what is a suitable default, since most people are going to want to see the whole thing. I suggest we merge this PR and make this a new issue.

R/kwic.R
```diff
@@ -135,39 +135,84 @@ is.kwic <- function(x) {
...
+    else
+        separator <- check_character(separator)
+    
+    attrs <- attributes(x)
+    x$pre <- rep("", nrow(x))
+    x$keyword <- rep("", nrow(x))
+    x$post <- rep("", nrow(x))
+    if (nrow(x)) {
+        toks <- attrs$tokens[x$docname]
+        lis_pre <- as.list(tokens_select(toks, startpos = pmax(x$from - window, 1), endpos = x$from - 1))
```
kbenoit, line 159: I discovered a bug when processing phrases that our tests did not detect, because they only compared parts of the result to the expectation. Fixed in a4741b1.

```r
> toks <- tokens(c(doc1 = "a a a b c d d d", doc2 = "a b c d e"))
> kwic(toks, phrase("b c"), window = 2)
Keyword-in-context with 2 matches.                                
 [doc1, 4:5] a a b | b c | c d d
 [doc2, 2:3]   a b | b c | c d e
```

kbenoit (approved):

R/kwic.R
```diff
@@ -135,39 +135,84 @@ is.kwic <- function(x) {
...
-        cat("kwic object with 0 rows")
-    } else if (!is.kwic(x)) {
-        NextMethod()
-    } else {
+print.kwic <- function(x, window = NULL, separator = NULL, ...) {
+    
+    attrs <- attributes(x)
+    x <- as.data.frame(x, window = window, separator = separator)
+    
+    cat(sprintf("Keyword-in-context with %d match%s.", nrow(x), ifelse(nrow(x) != 1, "es", "")))
```
koheiw, line 187: Don't you want to use `quanteda:::msg()`?

R/kwic.R
```diff
@@ -83,44 +82,45 @@ kwic.corpus <- function(x, pattern, window = 5,
...
-#' kwic(toks, c("is", "a"), valuetype = "fixed")
-#' kwic(toks, phrase(c("is", "a", "is it")), valuetype = "fixed")
+#' kwic(txt, "is", valuetype = "regex")
+#' kwic(txt, "or", window = 2)
 #'
 #' corp <- corpus(txt)
-#' kwic(corp, c("is", "a"), valuetype = "fixed", separator = "", remove_separators = FALSE)
+#' print(kwic(corp, "is"), separator = "")
 #' @export
 kwic.tokens <- function(x, pattern, window = 5,
```
koheiw, line 96: Shall we deprecate window and separator to make clear that `kwic()` is only for locating patters?

R/kwic.R
kbenoit: Sure, that's fine, but you wrote that line! I almost never use `sprintf()`.

R/kwic.R
kbenoit: I would say no, simply because this is the way that users have gotten used to using kwic since it was first written. It's also more natural to put this into the function that creates the kwic. We can make it clear that this can be changed using `as.data.frame()` or `print()`, but most people using quanteda probably don't want to type `print()` to see output, they just want to see default output. My experience with `kwic()` in particular is that this is a function most used by our most qualitative and interactive users.

When I use `kwic()` or teach this, I also prefer to set it at the time we call it. And this is not like what we've discussed about removing the convenience options from `dfm()` or removing `dfm.character()`, since those are a form of bad practice, skipping a step that the user should be taking more direct responsibility for. Rather, the `window` and `separator` options belong to `kwic()` and are natural to associate with creating a `kwic()`, but our implementation has now disassociated itself from needing to set this at creation time. But that's more about our current implementation than what the action implemented by the function calls for.

One more reason: If we do change `kwic()` again in the future, it might make more sense to set these at creation time. 

So I think it makes sense to keep them without warnings. Most users will happily never know the difference.

koheiw: It makes more sense to have `window` in `kwic()` but I am not comfortable with the same argument in the different places. How about let users to call `kwic.kwck()` to change `window` or `separator`? `max_nrow` should still be in `print()`.

kbenoit: `kwic.kwic()` makes no sense unless it would be possible to search a `kwic()` for keywords - although we could make it do so since we have the tokens stored inside it. I think that's an edge case however that will create more confusion than the benefit that might come from using such a method to (re)set window or separator.

I think it's ok to keep window and separator in both the constructor and the print and conversion methods. We just have to make it clear that the default is set when the kwic is created, but can be changed at the print or data.frame stage. My guess is that it will be a very rare user that wants to do so, so in the vast majority of cases for the vast majority of users, the kwic functions will simply function as they have before.

For `print.kwic()` I'd suggest that our default be `max_nrow = -1`. Want me to implement that?

koheiw: Let's take out `window` or `separator` from print for now. My understanding is that people do not want to run `kwic()` multiple times because it was really slow for large object, but it is no loner so was far as the resulting object is subset. Let set `max_nrow = 1000` or a bit more so that `kwic()` does not freeze.

koheiw: Thinking about why we have to make a decision to add these argument to either or both `kwic()` or `print()`... I think my design to make `print.kwic()` was the mistake. `kwic()` should be the print function that creates the concordance view.

What I should have done is creating a function that might be call `search()` or `locate()` that returns the positions of matches (the same as what `kwic.tokens()` does in this branch) and calling this function in `kwic()`. If we add a special class to this object (like "position"), we can make a `kwic.position()`. The object can be useful for information retrieval applications too.

kbenoit: That would work too. But the current structure is fine. Let's do this:
- Keep `kwic()` as in this PR.
- Remove `window` and `separator` from `print.kwic()`, but keep these for `as.data.frame.kwic()`.
- Add `max_nrow = 1000` to `print.kwic()` and make it work like the other object print functions (with the "x more rows..." output).
- Not implement yet the other functions you mention above, but keep in mind that these can be added when and if needed.

koheiw: OK. That is the middle way.

kbenoit: Alright, I did all of the above in 13fa3ea. This implements the same print options now for kwic as for other core objects, including `show_summary`. The arguments `window` and `separator` are kept in `as.data.frame.kwic()` but removed from  `print.kwic()` (but I left the internal code for these intact in case we want to add them back).

We can remove the .corpus and .character methods in a separate PR, when we address #2058 more directly.

kbenoit: Alright, I did all of the above in 13fa3ea. This implements the same print options now for kwic as for other core objects, including `show_summary`. The arguments `window` and `separator` are kept in `as.data.frame.kwic()` but removed from  `print.kwic()` (but I left the internal code for these intact in case we want to add them back).

We can remove the .corpus and .character methods in a separate PR, when we address #2058.

koheiw:
I checked how `kwic()` works. I don't think users enjoy wrapping the function by `as.data.frame()` always to see the concordance view in Data Viewer.

```r
View(kwic(corp, pattern = "secure*", valuetype = "glob", window = 3)) # shows only keywords and positions
View(print(kwic(corp, pattern = "secure*", valuetype = "glob", window = 3))) # shows nothing 
View(as.data.frame(kwic(corp, pattern = "secure*", valuetype = "glob", window = 3))) # shows concordance view
```

It is not necessary to define `View.kwic()` if `kwic()` returns concordance view (as I argued above).

> What I should have done is creating a function that might be call search() or locate() that returns the positions of matches (the same as what kwic.tokens() does in this branch) and calling this function in kwic(). If we add a special class to this object (like "position"), we can make a kwic.position(). The object can be useful for information retrieval applications too.

kbenoit:
That all sounds reasonable, but hard to follow as this PR keeps shifting.

To get View.data.frame to work with a raw kwic object, we really only need to add the pre and post columns to the kwic object. If you want them viewable from the kwic object, just add the columns from the window at the kwic creation step. 

We could keep the tokens attribute so that these could still be changed in `print.kwic()` or `as.data.frame.kwic()` since the original tokens are included, as in this PR. Or, if you no longer want to use that approach, we no longer need that attribute, but then we are not far from where we started before this PR.

How do you want to proceed?

koheiw:
I prefer to make different (exported) function only to locate pattern matches in C++, and class the object differently from "kwic" (say "matches"). Then we make `kwic.matches()`, that has `window` and `max_nrow` (or `page`) to limit rows in the concordance view and returns the KWIC object classed "kwic" (the same as v2.0).  

In this way, we can fulfill requests like in #2063, and avoid conflict with old KWIC objects.

We can move this discussion to an issue page and create a new branch if you like.

kbenoit:
OK - if that's based on this branch, then I suggest we merge this branch, and then work on a new branch for the changes you suggest. We have a lot of improvements in this PR including a better approach, in my view, for separating the construction of the kwic from its conversion and printing functions.

However I'm happy to add the pre and post columns to the data.frame part of the kwic object for now, as suggested above (https://github.com/quanteda/quanteda/pull/2045#issuecomment-784018541), if you think we should add that back to the kwic object in this branch before merging.

PR #2055, opened 2021-02-14
Implement text_field for corpus.corpus()

### LINKED ISSUE #2017: Add text_field to corpus.corpus
If `text_field` is specified in `corpus.corpus()`, the corresponding field in docvars should be the texts.

kbenoit:
Implements `text_field` for `corpus.corpus()`, as requested in #2017.

```r
> ?corpus.corpus
> df <- data.frame(doc_id = paste0("d", 1:2),
+                  en = c("The room.", "The car."),
+                  fr = c("La chambre.", "La voiture."))
> corp <- corpus(df, text_field = "en")
> corp
Corpus consisting of 2 documents and 1 docvar.
d1 :
"The room."

d2 :
"The car."

> corpus(corp, text_field = "fr")
Corpus consisting of 2 documents.
d1 :
"La chambre."

d2 :
"La voiture."

> # to preserve former text field
> corpus(corp, text_field = "fr", docvars = data.frame(en = texts(corp)))
Corpus consisting of 2 documents and 1 docvar.
d1 :
"La chambre."

d2 :
"La voiture."
```

Closes #2017


koheiw:
I was going to retract the request. My conclusion is users should keep the original data.frame if there is more than two potential text fields.

kbenoit:
I implemented your suggestion as in #2017 to drop the old text, but we could exchange it.  `text_field_old` could rename it and make it a docvar. 

Could be really useful if:
- a user has a multi-lingual corpus that they want to switch languages on while keeping all of the information in the single object
- a user has a survey dataset with several open-ended answers and they want to switch from one to another but keeping the rest of the data as docvars in the single corpus object.

It's a minor change, implementing an argument for corpus.corpus() that is already in other corpus methods, but useful. (I've had both scenarios above in my own work.) So I think your  motivation in filing the issue was sound, and that it's a good feature to implement.

koheiw:
We can record the column names of the input to `corpus.data.frame()` in the object meta and save the text in the original column.

kbenoit:
Where are we with this branch? Perhaps suggest a very specific change, or make it?

koheiw:
I don't think we should add `text_field` to `corpus.corpus()` unless it can return the same object as `corpus.data.frame()`. Please see my previous comment.

> We can record the column names of the input to corpus.data.frame() in the object meta and save the text in the original column.
> 
> A simpler approach is keep the text column but assign NULL to it; gnore NULL column in the internal get_docvars(). The text will return to the original column when corpus.corpus() is called.

kbenoit:
OK, fine with me to close this. Turns out it's easy to do anyway as below.

``` r
library("quanteda")
## Package version: 2.9.9000
## Unicode version: 10.0
## ICU version: 61.1

df <- data.frame(
  doc_id = paste0("d", 1:2),
  en = c("The room.", "The car."),
  fr = c("La chambre.", "La voiture.")
)

corp <- corpus(df, text_field = "en")
corp
## Corpus consisting of 2 documents and 1 docvar.
## d1 :
## "The room."
## 
## d2 :
## "The car."
docvars(corp)
##            fr
## 1 La chambre.
## 2 La voiture.

corp2 <- corpus(convert(corp, to = "data.frame"), text_field = "fr")
corp2
## Corpus consisting of 2 documents and 1 docvar.
## d1 :
## "La chambre."
## 
## d2 :
## "La voiture."
docvars(corp2)
##        text
## 1 The room.
## 2  The car.
```

<sup>Created on 2021-02-23 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

PR #2064, opened 2021-02-23
Tidy up documentation

### LINKED ISSUE #2056: Add party to inaugural corpus documentation
Documentation refers to 3 docvars but the object has 4 (including Party, not documented).

kbenoit:
This PR tidies up the documentation.

- adds `Party` to the documentation of the `data_corpus_inaugural` docvars
  Closes #2056 
- Removes remaining aliases and references to removed function `metacorpus()`
- Makes the following internal only: `unlist.tokens()`, `object2fixed()`, `object2id`.
- Removes the "show" methods for S4 objects (dictionary2, dfm, fcm) from the print-methods entry
- Removes the "print-quanteda" index entry, now it's just "print-methods".
- Also removes the revdep folder and .gitignores it for future use.


koheiw (changes requested):
`object2fixed()` and `object2id()` should not be internal. Don't you want  people to develop packages using quanteda?

koheiw:
I also wonder why we still have `dictionary_edit()`. I thought we agreed to remove it.

kbenoit:
I absolutely do want to encourage developers to make use of quanteda's object handling functions, but I think we should put these into a developer's guide vignette. I also have some suggestions for the object2* functions and wanted to make those separately. I'll do that in a new issue shortly.

I will move the edit functions out, just did not do it here. It involves moving them to another package and additional tests, so is trickier than the simple ones in this PR.

koheiw:
OK. Sounds like a plan.

koheiw (approved):

Issue #2058, opened 2021-02-17
Consider removing "shortcut" methods


kbenoit:
Right now we have lots of methods for constructing things that skip the tokenisation step. This is convenient, but leads to inconsistencies and in the case of #2045, inefficiencies. We should give serious consideration to dropping methods for pre-tokenization objects that allow users to do things that require tokenization internally.

https://github.com/quanteda/quanteda/pull/2045#issuecomment-779235429, is correct, because of memory management, but only if there is a tokens object for the tokens attribute of the new kwic object to point to. When we run kwic on a corpus or character, however, the tokens inside the kwic will be a duplicate.

Which ones? These are the methods for corpus and character that allow the user to leapfrog tokenization.

- `dfm()`, `fcm()`. Removing `dfm.tokens()` would mean we could also remove some of the convenience arguments (see #1602)
- `kwic()`
- `ntoken()`, `ntype()`, `nsentence()`. These are very useful but we could keep equivalent functionality in `textstat_summary.corpus()`, with a health warning.

Actually that list is not as bad as I thought! But we would have to give serious consideration to the code we would break, and consider deprecating these first.

The only reason to keep the corpus methods would be in case we implement tokenization within a corpus, similar to what happens with you call `nlp()` from spaCy, but even then we would probably want users to extract them into a tokens object after applying `remove_*` options.

Here's another thought: We could introduce a new `quanteda_option(strict = TRUE)` in v3 that when TRUE, would not allow skipping the creation of a tokens object. When `FALSE`, it will still allow the earlier behaviour. This provides a way to provide backward compatibility without all of the deprecation warnings that simply using `.Deprecate()` would create. But we'd make users turn this on, so that they are at least aware.

koheiw:
I think it is a good idea to add `strict = TRUE` (or `compatibility = "v2"`) in the package option or the function itself for this and other changes. With the option, we can remove `dfm.character()` and `dfm.corpus()` straight away.

PR #2067, opened 2021-02-26
Deprecate convenience arguments in dfm()

koheiw:
For #2058. I  support the idea of deprecating `dfm.corpus()`, but I did not because it brakes many bad examples in the quickstart guide.


kbenoit:
I was thinking that the best way to really clean this up and discourage the shortcuts would be to remove the options from the function signature, but recognise them if called, by looking for these in `...`. If found, we keep them operational but issue the warning. (We did this when we changed the `tokens()` arguments for v2.)

The only tricky one is `tolower = TRUE`. The question is whether to keep this in the function signature and `TRUE`, or to go hard-core are remove it too. The problem is that if we remove it, then we will break existing behaviour since we probably don't want an argument not listed, but deprecated and operational, to be actively changing the input object.

This is a good case for the quanteda_option we discussed somewhere in all of this...

So I'm suggesting at least for now, we:
- remove all but `tolower = TRUE` from `dfm()`, and `verbose`. And consider making this operate via a quanteda_option, but that should be in a separate PR since it could involve other changes.
- make all other arguments functional via `...` but with warnings
- deprecate `dfm.character()` and `dfm.corpus()`

So `dfm()` would work just on tokens and look like
```r
dfm(x, tolower = TRUE, verbose = quanteda_options("verbose"), ...)
```

koheiw:
It makes sense to hide deprecated arguments. 

I would keep `tolower` becasue DFM should discard information about positions of tokens (i.e. the first word in sentences), but capitalization is not always about positions (i.e. proper names). In many non-European languages, lowercasing does not change anything. I would also keep `remove` in `dfm()` or add `remove_padding = FALSE` to it to remove paddings. I think it is extremely rare that people want to keep paddings in DFM as a feature, but making it `TRUE` would upset users.

kbenoit:
How about we keep `remove` but change its default to `remove = ""` which removes pads?

koheiw:
This is exactly why I think we need `remove`.

kbenoit:
I started working on this yesterday and even small changes caused so much to start breaking. We might need to keep `remove` as is, but include the `remove = ""` in the examples. I'm happy to work more on this one.

kbenoit:
Current state of the PR:
- Implemented all previous `dfm()` argument functionality, but as non-documented, deprecated arguments. Only `valuetype` and `case_insensitive` are not working now.
- I kept the default for `remove` to `NULL`, rather than make the default value `""`. We could change that, but it breaks a lot of tests (and therefore, previous behaviour).
- Added a new Rd page called `quanteda-deprecations` where we can detail the breaking changes, as well as how to set the `strictness_warning` if we go with that (and I strongly think we should). This can also be a section of the pkgdown site.
- I also removed the previously deprecated ability to pass a dfm as a `pattern` input to `dfm_select()`.
 
Still needed:
- deprecate `kwic()` for character and corpus inputs
- add more detail to `quanteda-deprecations.Rd`
- add tests for deprecation warnings

koheiw:
What it the advantage of giving an option to suppress deprecation warnings? If users do not want to see warnings, they should update the code. If they postpone changes, they will notice the change when the functions/arguments are removed (and unable to make sure that the results are unchanged). It is better to make the deprecation messages more informative (e.g. add a link to a web page on how to update the code) than hiding them.

kbenoit:
Advantage is to ease the user pain. I don’t want to abruptly break colleagues’ teaching code, student code, researcher code, or other package author’s code, if we can avoid it as easily as adding an option to make the effect more mild, initially. With an option that has three levels of warning, we can tighten the strictness in minor releases - 3.1, 3.2, etc. Agree on the need to be informative, that’s why I created the `quanteda-depreciations.Rd`. And we can write a vignette article about the breaking changes and other improvements in v3. 

Just take a look at our vignette articles - they pretty much all need changing because of our depreciations. (I only did the main English quickStart.Rmd so far.) If it affects this way, imagine how widely it affect users.

koheiw:
I haven't see option to suppress deprecation code anywhere else. Without surprising users with warning messages, they do not update their code (I also do not think deprecation messages break users code). If you want to let users to keep using their old code, better not changing any behavior and release **quanteda3**. I prefer this approach than a slow transition.

kbenoit:
Let's merge this and see how bad it is with revdepcheck and with our own vignettes. I'll test this on some instructional materials too that we've developed.

The only open question is whether we want to go fully extreme and remove the `remove` option too. But I think we decided above that we both think it's better to keep for now.

koheiw:
Please do merge. Let's keep `remove` in `dfm()` until we decide to add `remove_padding` or something.

PR #2071, opened 2021-03-01
Deprecate corpus and character methods for dfm and fcm

kbenoit:
Addresses #2058.
- Attaches a deprecation warning to `dfm.corpus()`, `dfm.character()`, `fcm.character()`, and `fcm.corpus()`.
- Updates tests to call tokens first for these functions, except where we want to continue to test older methods, in which case it suppresses the warning.
- Updates  documentation, including examples.
- Also removes the `context()` from the header of the test files, since this is now discouraged in newer versions of **testthat**.

This was a bit of a nightmare not just for how extensively _we_ used these shortcut methods, but also by the future prospect of insane numbers of irritating warnings for the other users whose code will also issue endless warnings about this. I propose the following way around this: We add a `quanteda_options(strictness_warnings = c("never", "once", "always")` that issues none of these deprecation warnings for "never"; only the first time a method is called for "once"; and every time for "always". Default will be "once", so that it resets with every new session.

We could also add `kwic.corpus()`, `kwic.character()` to this PR.

I propose we resolve this, and add tests, and merge before returning to #2067. Although we could use the same `strictness_warning` for that as well.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # quanteda 3.0
 
+**quanteda** 3.0 is a major release that improves functionality, completes the modularisation of the package begun in v2.0, and further improves function consistency by removing previously deprecated functions, and deprecating some shortcut steps in the text analysis workflow.
+
 ## Changes
 
 * Separated the `textplot_*()` functions from the main package into a separate package **quanteda.textplots**.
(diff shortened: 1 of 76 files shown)


koheiw (approved):

koheiw:
I don't think we need `quanteda_options(strictness_warnings = c("never", "once", "always")` because `options(warn = -1)` turns off warnings. Uses would be annoyed but appreciate the tokens object shortly after.

kbenoit:
Of course they could turn off all warnings, but that would mask other warnings that they would need to see.

I think if we slap warnings onto the convenience options and direct-from-character/corpus methods, we should provide some cushion for the annoyance that results. The extensive use in our own code just goes to show how common their use probably is among users. I'll work that into a separate PR.

PR #2075, opened 2021-03-06
Fix revdep breaking issues

### LINKED ISSUE #2016: Solve revdep issues with modularised v3 release
I have identified revdep issues with the following, so far, related to them calling functions from **quanteda** that will soon be relocated to the modular packages.

| package         | fix status         | new dependency     | branch                                                          |
|:-----------------|:--------------------|:--------------------|:-----------------------------------------------------------------|
| LSX             | Fixed GitHub but not CRAN | quanteda.textstats |              |
| collector       | PR submitted | quanteda.textstats | https://github.com/kbenoit/collector/tree/fix-quanteda-v3       |
| LexisNexisTools | Fixed on CRAN | quanteda.textstats |  |
| newsmap         |    Fixed on CRAN                | quanteda.textstats |                                                                 |

kbenoit:
Makes adjustments to the deprecations already implemented for `dfm()`, to fix some of the ways that it was breaking packages that used `dfm()`.

- Fixes a problem wherein `dfm()` calls would break if called as `quanteda::dfm()` without also importing `tokens()`.
- Fixes false warnings when `verbose` is specified.
- Adds back functionality for `valuetype` and `case_insensitive`, although these remain hidden from the function signature. (@koheiw if you think we should add these back to the argument list, since we have kept `remove`, let me know - I nearly added them back myself, but thought it best to keep `dfm()` simple rather than enable the use of a functionality we are deprecating.)

Now, only **politeness** and **LexisNexisTools** are breaking, and I've issued PRs for both that have already been merged. (But neither yet re-submitted to CRAN.)

Closes #2016.

### DIFF
--- a/R/dfm.R
+++ b/R/dfm.R
@@ -4,12 +4,8 @@
 #' [tokens], or even other [dfm] object.
 #' @param x a [tokens] or [dfm] object
 #' @param tolower convert all features to lowercase
-#' @param remove a [pattern] of user-supplied features to ignore, such as "stop
-#'   words".  To access one possible list (from any list you wish), use
-#'   [stopwords()].  The pattern matching type is fixed to "glob"; if you want
-#'   greater control, use [tokens_remove()] or [dfm_remove()].  `remove` is
-#'   primarily useful for removing artefacts created during the [tokens]
-#'   creation process, such as pads.  See examples.
+#' @param remove_padding logical; if `TRUE`, remove the "pads" left as empty tokens after
+#' calling [tokens()] or [tokens_remove()] with `padding = TRUE`
 #' @param verbose display messages if `TRUE`
 #' @param ... not used directly
 #' @section Changes in version 3:
@@ -44,7 +40,7 @@
 #' dfm(toks, tolower = FALSE)
 dfm <- function(x,
                 tolower = TRUE,
-                remove = NULL,
+                remove_padding = FALSE,
                 verbose = quanteda_options("verbose"),
                 ...) {
     dfm_env$START_TIME <- proc.time()
@@ -95,11 +91,16 @@ dfm.corpus <- function(x, ...) {
 #' @export
 dfm.tokens <- function(x,
                        tolower = TRUE,
-                       remove = NULL,
+                       remove_padding = FALSE,
                        verbose = quanteda_options("verbose"),
                        ...) {
     # check for arguments passed to tokens via ...
     otherargs <- list(...)
+    # if "remove" was matched to "remove_padding"
+    if ("remove" %in% setdiff(names(as.list(sys.call())), names(as.list(match.call())))) {
+        otherargs[["remove"]] <- remove_padding
+        remove_padding <- FALSE
+    }
     otherargs_tokens <- otherargs[names(otherargs) %in% names(as.list(args(quanteda::tokens)))]
     if (length(setdiff(names(otherargs_tokens), "verbose"))) {
         .Deprecated(msg = "'...' should not be used for tokens() arguments; use 'tokens()' first.")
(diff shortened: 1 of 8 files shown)


koheiw:
It is true that `dfm()` should bave `valuetype` and `case_insensitive` along with `remove`. If we add `remove_padding = FALSE` now, we can deprecated all the three. If not now, we need to keep all of them active and visible.

kbenoit:
I can see the value of having remove, for convenient removal of stopwords, pads, punctuation, and other possible characters left over from tokens manipulations. But it's also inconsistent with our approach of deprecating the shortcuts.

So the options are what you said. Which do you prefer? I think I mildly prefer `remove_padding` only.

koheiw:
In the long run, we have to defunct `remove` and add `remove_padding`. This is the question of when. I think v3 is a good time to do.

kbenoit:
OK I'll do this later today. Or, if you are fixing LSX, merge it and we can modify the remove arguments later.

kbenoit:
Ugh, trapping `remove` without it automatically expanding to `remove_padding` is a bit of a pain. But possible.

kbenoit:
OK done with implementing `remove_padding`. It's not the prettiest, but works. And extensively tested.

I think the modular chunks of code approach taken here, while not the prettiest, is better than modularising it into utility functions, since it's special to need to partition the deprecated arguments into two sets: those originally passed to `tokens()` via `...`, and those originally named `dfm()` arguments. This way, we can make the deprecated functionality defunct in a future release simply by deleting the code blocks, rather than pulling on some thread of utility function calls that could cause more fabric to unravel.

This now works with the  current master branch of **LSX** as well. (You should not change the `dfm(x, remove = "")` in `textstat_context()` for now since that works with both this PR and v2.1.2, but you could `suppressWarnings()` on that in **LSX** to make it warning-free for v3, and update it once v3 is on CRAN).

koheiw (changes requested):
It is better to set `dfm_remove(x, "", valuetype = "fixed")`. Otherwise looks good ( the use of `do.call()`).

kbenoit:
I'm reluctant to change the (hidden) `valuetype` default from "glob" to "fixed", even if it's likely to affect very few users. That is effectively saying: We are deprecating an argument and hiding it from the function signature and documentation, but also changing its default behaviour. Your code will now work but work differently, and it will have become harder for you to figure out why. So I don't think we should change it.

R/dfm.R
```diff
@@ -182,6 +200,11 @@ dfm.tokens <- function(x,
 
     check_dots(otherargs, method = "dfm")
 
+    remove_padding <- check_logical(remove_padding)
+    if (remove_padding) {
+        x <- tokens_remove(x, "")
```
koheiw, line 205: Adding `valuetype = "fixed"` will disable auto-detection of glob in `pattern2id()` but I admit that the difference is very small.

koheiw: More importantly, it is faster to remove padding after forming a DFM.

```r
> toks <- tokens(data_corpus_inaugural, remove_punct = TRUE, padding = TRUE)
+   dfm(tokens_remove(toks, "")),
+   dfm_remove(dfm(toks), "")
+ )
Unit: milliseconds
                         expr     min       lq     mean   median       uq      max neval
 dfm(tokens_remove(toks, "")) 73.2486 94.10910 151.1463 126.0674 172.1803 399.8985   100
    dfm_remove(dfm(toks), "") 65.8952 87.95525 122.9699 101.7932 134.7257 330.3436   100
```

kbenoit: I'm getting very similar results for all three options below. Looks like the `remove_padding = TRUE` is about the same. Was that just for information or were you proposing we remove `remove_padding`?
```r
+    dfm(tokens_remove(toks, "")),
+    dfm_remove(dfm(toks), ""),
+    dfm(toks, remove_padding = TRUE)
+ )
Unit: milliseconds
     dfm(tokens_remove(toks, "")) 24.52573 27.33274 32.28436 30.10545 33.06334 149.6006   100
        dfm_remove(dfm(toks), "") 24.52290 27.62597 33.32570 30.62006 32.69043 149.8663   100
 dfm(toks, remove_padding = TRUE) 22.91938 27.99522 32.07499 30.15143 32.86733 144.3347   100
```

kbenoit: I misunderstood about the `valuetype = "fixed"` point earlier, thought you meant the deprecated `valuetype`, but you meant for `remove_padding`. Changed now in d54e6ea.

koheiw (approved):

Issue #2028, opened 2020-12-17
Implement NSE consistently for functions that reference docvars


kbenoit:
As it works for the `*_subset()` functions, we would want to extend this functionality to arguments that take `group`, or `by`, or (in `textstat_keyness()`), `target`/`reference`, etc.

- `subset` - (logical expression) - Status ✅
    - `corpus_subset()`
    - `tokens_subset()`
    - `dfm_subset()`
 
- `by`- (docvar name(s)) - Status ✅ (see #2081)
    - `corpus_sample()` 
    - `dfm_sample()`
    - `tokens_sample()`

- `groups` - (docvar name(s)) - Status ✅ (see #2081)
    - `dfm_group()`
    - `tokens_group()`
    - `dfm_group()`
    - `texts()` - See #2076 
    - `topfeatures()` - We will remove this

There are also methods in **quanteda.textmodels** -- anything with `y` -- and `textstat_keyness()` in **quanteda.textstats** where this functionality should be applied.

kbenoit:
This is not so hard to do, and I have it working in a test branch, BUT: it breaks arguments such as:

```r
dfm_group(dfm(data_corpus_inaugural[1:8]), groups = "Party")
```

_although_ this already does not work:

```r
dfm_subset(dfm(data_corpus_inaugural[1:8]), subset = "Party" == "none")
```

koheiw:
We are very good in NSE in `*_subset()`. I think the easiest way to upgrade `*_group()` is adding `by` with full NSE while deprecating `groups`.

koheiw:
By the way, we should remove `groups` from `texts()` and `topfeatures()`.

kbenoit:
OK I'll get the NSE for `*_group()` and `*_sample()` working, but with a trap and warning for passing docvar names as characters (but making them work).

Will remove `groups` from topfeatures. `textstat_frequency()` does this now.

On the groups from texts, I find this a really useful building function for some of my work. We could drop it if we were to write a `corpus_group()`. Should we add that and then drop the `groups` from `texts()`?

koheiw:
It is probably the time to export `tokens_group()` and add `corpus_group()`. They will look very similar to each other.

kbenoit:
Agreed. Let me keep working on the NSE for the existing ones first, we can add corpus_group once I've fixed the others.

[...]

PR #2081, opened 2021-03-13
Implement NSE for by in *_sample() functions

kbenoit:
Implements #2028 for by in `*_sample()` functions. @koheiw I think you will like this one since it is clean and now completely consistent, and removes hacks like `by = "document"`. The code is also simpler. It breaks some previous code however, so please take a look at whether you agree with this approach, before I implement similar changes to the `groups` arguments as per the last part of #2028.

- `by` now takes unquoted docvars. It will no longer work with quoted docvar names, nor with `"document"`, which should now be replaced with `by = docid(x)` (or alternatively, `by = docid_`).
- The documentation is made much more consistent across all three `*_sample()` functions.
- There are updates to `?quanteda-deprecations` but I am thinking now that we could remove this, and replace it with the suggestion to see `news(Version >= "2.9.0", package = "quanteda")`, which means we keep the information all in one place.


koheiw (approved):
Nice to see the code getting shorter.

kbenoit:
Agreed - although sometimes it has to get longer in order to get shorter. Shorter here cause it's a breaking change.

kbenoit:
Latest commits complete the NSE issue by extending the same functionality and consistent implement as with `by`, to the `groups` arguments.

Also removes the `quanteda-deprecations` page and now encourages users to use `news(Version >= "2.9", package = "quanteda")` instead. I've also rewritten this carefully so that it documents the changes in v3 including deprecations and suggested new usages.

Solves #2028. I'll file a separate issue for the `texts()` and `topfeatures()` functions, since they might need deprecations.

Issue #2078, opened 2021-03-07
Add corpus_group(), export tokens_group()


kbenoit:
We hesitated to do this before since it is sequence dependent. But now via `[` both corpus and tokens objects are easily reordered, without losing their special attributes. `texts(x, groups = "...")` always had the order-dependent issue. 

So let's implement `corpus_group()` so that we can implement #2076. But first, let's solve #2028, since this will affect how we implement the `groups` functionality.

So the sequence: #2028, then #2078 (this issue), then #2076.

PR #2082, opened 2021-03-15
Issue 2078

koheiw:
Add `corpus_group()` and export `tokens_group()` for #2078.


kbenoit:
This is good but some work was already done in #2081. The `tokens_group()` exporting and cleanup, including moving the `fill` argument Rd, was done in dfab982. Also here, `corpus_group()` uses the older quoted docvar names for `groups`, but I can change that to make it consistent with the `tokens_group()` and `corpus_group()` in #2081. I'll merge that #2081 now and then we can clean this PR up. 

Should we go ahead and deprecate `texts()` as well in this PR, since
- `texts(corp)` is now the same as `as.character(corp)`, but the "as" method seems more consistent with our usage throughout the rest of the package for coercing to base objects; and
- `texts(corp, groups = "something")` is now available through `as.character(corpus_group(corp, groups = "something"))`?

koheiw:
I did not make changes for NSE deliberately in this PR.

kbenoit:
OK but you agree that we should treat `groups` the same for `corpus_groups()` as for the other group fns?

koheiw:
Yes, all the `*_group()` should work the same way (there are even tests for that).

kbenoit (approved):
I'm happy with this. I changed `corpus_group()` and the tests to make it consistent with the other group functions. I also consolidated the documentation a bit more and made it consistent across the functions. 

Merge if you are happy with it.

Issue #2076, opened 2021-03-07
Deprecate `texts()` and `texts<-()`


kbenoit:
- `texts(x)` - this has long been a selling point of **quanteda** versus **tm**: it's super easy to extract your texts from the corpus objects. But now we can offer the same functionality via `as.character()`.
    ```r
    > identical(texts(data_corpus_inaugural), as.character(data_corpus_inaugural))
    [1] TRUE
    ```

- `texts(x)<-` - also already works via `<-.corpus`, so no need for that function:

``` r
library("quanteda")
## Package version: 2.9.9000

(corp <- corpus(c(d1 = "Doc one.", d2 = "Doc two."),
  docvars = data.frame(var1 = 1:2, var2 = c(TRUE, FALSE))
))
## Corpus consisting of 2 documents and 2 docvars.
## d1 :
## "Doc one."
## 
## d2 :
## "Doc two."

corp2 <- corp
corp2[2] <- "Replacement text."

corp2
## Corpus consisting of 2 documents and 2 docvars.
## d1 :
## "Doc one."
## 
## d2 :
## "Replacement text."
summary(corp2)
## Corpus consisting of 2 documents, showing 2 documents:
## 
##  Text Types Tokens Sentences var1  var2
##    d1     3      3         1    1  TRUE
##    d2     3      3         1    2 FALSE
identical(meta(corp, type = "all"), meta(corp2, type = "all"))
## [1] TRUE
```

<sup>Created on 2021-03-07 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

- `texts(x, group = "...")` - This is a functionality I still find useful, and could offer a path forward if we pre-tokenize a new form of corpus object. I also use it with transcripts, e.g. https://blog.quanteda.org/2019/02/12/text-analysis-of-the-10th-republican-presidential-candidate-debate/

Another example:
``` r
library("quanteda")
## Package version: 2.9.9000

chars <- as.character(tokens("This text, will be split!", what = "character", remove_separators = FALSE))
chars
##  [1] "T" "h" "i" "s" " " "t" "e" "x" "t" "," " " "w" "i" "l" "l" " " "b" "e" " "
## [20] "s" "p" "l" "i" "t" "!"
texts(chars, groups = rep(1, length(chars)), spacer = "")
##                           1 
## "This text, will be split!"
```

[...]

PR #2086, opened 2021-03-16
Deprecate texts() and texts<-

kbenoit:
Deprecate `texts()` and `texts<-`. Also:
- moves these functions to texts.R, makes them internal only, and marks them as deprecated in the title
- reorganizes main object methods source files
- creates a more consistent man page for corpus coercion and checking methods (is.corpus, as.corpus)
- updates NEWS
- changes examples, tests, and internal code to use as.character.corpus() instead of texts()
- makes as.character.corpus() independent and texts() now calls as.character(), instead of the other way around as it was previously.

Solves #2076.


koheiw:
It makes sense to use `as.character()` instead of `texts()`. How about defining `corpus<-`? Do you want to prevent users from changing texts in a corpus entirely?

kbenoit:
Interestingly the "[.corpus" already works for texts replacement, without messing up metadata. But it has to be used, even if the index is empty.

``` r
library("quanteda")
## Package version: 2.9.9000
## Unicode version: 10.0
## ICU version: 61.1

corp <- corpus(letters[1:3], docvars = data.frame(dvar = 101:103))
corp
## Corpus consisting of 3 documents and 1 docvar.
## text1 :
## "a"
## 
## text2 :
## "b"
## 
## text3 :
## "c"

corp[] <- c("new1", "new2", "new3")
corp
## Corpus consisting of 3 documents and 1 docvar.
## text1 :
## "new1"
## 
## text2 :
## "new2"
## 
## text3 :
## "new3"
docvars(corp)
##   dvar
## 1  101
## 2  102
## 3  103

corp[2] <- "second text"
corp
## Corpus consisting of 3 documents and 1 docvar.
## text1 :
## "new1"
## 
## text2 :
## "second text"
## 
## text3 :
## "new3"
docvars(corp)
##   dvar
## 1  101
## 2  102
## 3  103

corp <- paste0("text", 1:3)
corp
## [1] "text1" "text2" "text3"
```

But without the `[` it simply replaces the entire object. It's logical but not obvious. Perhaps we should document this more.

It's also possible we could just hide `texts()` and `texts()<-` now but turn off the deprecation warnings. They seem to get used a lot in our examples and our own code.

<sup>Created on 2021-03-19 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

koheiw:
I came across a package that actively use empty `[]` long time ago (probably network analysis package). I kind of liked it but it is very new to many users. I think we should some how allow users to modify character strings in a corpus.

kbenoit:
Agreed. But that functionality works fine now using `corp[] <-` with no or any valid indexes in the `[]`.

koheiw (approved):
OK, let's move on with `corp[]`. The manual page for `[` says:

> An empty index selects all values: this is most often used to replace all the entries but keep the attributes.

PR #2089, opened 2021-03-18
Eliminate deprecated or unnecessary code for v3

kbenoit:
- Removes the unnecessary `head.corpus/tail.corpus()`, as the base methods work fine on the corpus as character structure
- Removes the deprecated `dfm_weight()` scheme options
- Removes the deprecated `tokens()` arguments `remove_hyphens` and `remove_twitter`


koheiw:
I think `head/tail.dfm()` can be removed too unless you think the `nf` argument is important.

kbenoit:
I agree that for consistency, we should remove `head.dfm()` and `tail.dfm()`. `nf` was added before we introduced `print.dfm()`, but it does truncate columns in the returned object. The index operator can be used for that. So in order to be as standard as possible, let's rely on the base head() and tail() methods for dfm. That means we no longer use special versions of head or tail for any of our classes. 👍

kbenoit:
OK I removed any special head or tail methods (for dfm and fcm). Also made sure that print.dfm appears in the help index.

koheiw (approved):

Issue #2037, opened 2021-01-04
dfm_group(fill = TRUE) drops empty factors from the filled docvars


koheiw:
```
> corp <- corpus(data.frame(text = c("aa bb", "bb cc"), 
+                           var = factor(c("x", "y"), levels = c("x", "y", "z"))))
> dfmt <- dfm(corp)                          
> dfm_group(dfmt, "var", fill = TRUE)                          
Document-feature matrix of: 3 documents, 3 features (55.6% sparse) and 1 docvar.
    features
docs aa bb cc
   x  1  1  0
   y  0  1  1
   z  0  0  0
> docvars(dfm_group(dfmt, "var", fill = TRUE))
   var
1    x
2    y
3 <NA> 
```
The last value of `var` should be `z` because the document is created by filling.

koheiw:
There is a branch `issue-2037` in which I tried to solve this issue, but put it on hold because I could not find a fast and reliable way.

PR #2091, opened 2021-03-19
Issue2037 2

koheiw:
Redone #2037 on the current master. 

The values for the grouping variable is filled only when it is a factor in the docvars, becasue coercion of numeric to factor then to numeric can cause problems.

I also find used of `substitute()` twice is redundant, but kept the original code (second) intact to avoid changing the NSE behavior.
```r
field <- deparse(substitute(groups))
groups <- eval(substitute(groups), get_docvars(x, user = TRUE, system = TRUE), parent.frame())
```
Once agreed on how to implement, we can updated `tokens_group()` and `corpus_group()`.


koheiw:
Strange error is happening in the deprecation test... It does not replicate when I run the code manually.
```
══ Failed tests ════════════════════════════════════════════════════════════════
── Error (test-dfm.R:152:5): dfm.dfm works as expected ─────────────────────────
Error: 'length(x) = 12 > 1' in coercion to 'logical(1)'
Backtrace:
    █
 1. ├─base::suppressWarnings(...) test-dfm.R:152:4
 2. │ └─base::withCallingHandlers(...)
 3. ├─quanteda::dfm(...)
 4. └─quanteda:::dfm.dfm(...)
 5.   ├─base::do.call(...)
 6.   ├─(function (x, groups, fill = FALSE, force = FALSE) ...
 7.   └─quanteda:::dfm_group.dfm(...)
 8. ```

**Comment by kbenoit** [comment]
I experimented a bit but did not solve it. The best place I think to solve it is in `group_docvars()`, which is still structured for taking the name of a docvar (in `field`) rather than its value. In the way we now implement the `groups` to be evaluated within the scope of the docvars, it would be best to make use of that rather than trying to get the name as well. It's tricky to exploit this property of lazy evaluation, since it's easy to trigger evaluation even with the first substitute. I think that's what's breaking the call from `dfm()` in the tests shown in CI.

Just FYI, this is not unique to the way we have implemented our "NSE" - it's the same as `subset()` and therefore how we've long used this in our `*_subset()` functions.

``` r
subset2 <- function(x, subset, select, drop = FALSE, ...) {
  subset(x, subset, select, drop, ...)
}

head(subset(iris, subset = Species == "setosa"))
##   Sepal.Length Sepal.Width Petal.Length Petal.Width Species
## 1          5.1         3.5          1.4         0.2  setosa
## 2          4.9         3.0          1.4         0.2  setosa
## 3          4.7         3.2          1.3         0.2  setosa
## 4          4.6         3.1          1.5         0.2  setosa
## 5          5.0         3.6          1.4         0.2  setosa
## 6          5.4         3.9          1.7         0.4  setosa
head(subset2(iris, subset = Species == "setosa"))
## Error in subset2(iris, subset = Species == "setosa"): object 'Species' not found
```

There are other ways to do this using **rlang** (the tidyverse way) but that adds overhead we decided against.

<sup>Created on 2021-03-19 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

kbenoit:
OK, that worked, nice one. Passing the arguments as a list via do.call()  triggers evaluation of the arguments in a way that direct passing does not, so it does not work with NSE.

```r
> do.call(subset, list(iris, subset = Species == "setosa"))
Error in do.call(subset, list(iris, subset = Species == "setosa")) : 
  object 'Species' not found
```

I'm happy for you to do this with the other group functions.

kbenoit (approved):

kbenoit:
Very clever solution for `dfm()`, to let an anonymous function filter named arguments from those in `...`. It will all look cleaner when we remove the deprecations in the next major version.

The only thing that occurs to me is to remove the (e.g.) `if (is.missing(groups))` condition and code for each `*_group()` function, since the function ought not to operate without a group being specified. What do you think?

koheiw:
The test that I added in the last commit is actually for `*_group()` where groups is missing. In this case, documents are grouped by `docid`. Please merge if you think all good.

kbenoit:
I saw that, and that's what gave me pause. 
1) An argument with no default value ought to be required.
2) An argument with an effective default value should be revealed. Here, the default value can actually change the object in ways that the user will not know, without inspecting the code (the default is `docid(x)`). This changes the object if it has been previously reshaped or segmented.

```r
> corp1 <- corpus(c(textone = "This is a sentence.  Another sentence.  Yet another.",
+                  textwo = "Sentence 1. Sentence 2."))
> corp2 <- corpus_reshape(corp1, to = "sentences")
> 
> corpus_group(corp2)
Corpus consisting of 2 documents.
textone :
"This is a sentence. Another sentence. Yet another."

textwo :
"Sentence 1. Sentence 2."
```

How about we change the function to require a `groups()` argument to be supplied? If users want the current default behaviour, they can supply `groups = docid(x)` in the call. (We do this already with `by` in the `*_sample()` functions - see code snippet below.)
```r
> corpus_group(corp2, groups = docid(corp2))
Corpus consisting of 2 documents.
textone :
"This is a sentence. Another sentence. Yet another."

textwo :
"Sentence 1. Sentence 2."
```

It might be worth better documenting `docid()` since right now the man page has no details, and I've already seen a user on StackOverflow confusing this with `docnames()`. I noticed we use this in the example for `corpus_sample()` for instance.
https://github.com/quanteda/quanteda/blob/7e802cc834ff8af529bfb0da5a27dd97b48d4dd6/R/corpus_sample.R#L39-L43

koheiw:
I don't like missing values too. How about making it `dfm_group(x, groups = docid_)` if we are allowed to use NSE as the default value? If not fine with `dfm_group(x, groups = docid(x))`.

kbenoit:
Hmmm... `groups = docid_` would encourage users to access system components of our docvars object that I think is best avoided. `groups = docid()` is better since it encapsulates access in our function, but we currently don't promote that function or document it, so we would need to change that. (But we probably need to change it anyway.)

My preference remains to make this argument mandatory, with no default (like `f` in `split()`). We could improve the documentation of `docid()` and show an example there using `groups = docid()` but this would be only for very advanced users or developers.

koheiw:
If am fine with `groups` to be mandatory. `dfm_group(x, groups = docid(x))` will be the same as current behavior, but `dfm_group(x, groups = NULL)` is fine too. 

I don't have a strong opinion on this, but  hope that `dfm_group()` will not silently stop grouping when `groups` is missing. Will you raise deprecation warnings if `groups` is missing?

kbenoit:
Well, we cannot make `groups` a required argument but also leave the default as `docid(x)` when `groups` is not supplied (since then it would not be a required argument).

Would it not be more consistent, and explicit, to specify `groups = docid(x)` when you need the former (current, unless we change it) behaviour for when `groups` is omitted?

This has always seemed inconsistent (and undocumented) to me (from 2.1.2):
``` r
library("quanteda")
## Package version: 2.1.2
## 
## 
##     View

corp <- corpus(c(
  textone = "This is a sentence.  Another sentence.  Yet another.",
  textwo = "Sentence 1. Sentence 2."
))
corp2 <-

  dfmat <- corp %>%
  dfm()
dfmat
## Document-feature matrix of: 5 documents, 9 features (62.2% sparse).
##            features
## docs        this is a sentence . another yet 1 2
##   textone.1    1  1 1        1 1       0   0 0 0
##   textone.2    0  0 0        1 1       1   0 0 0
##   textone.3    0  0 0        0 1       1   1 0 0
##   textwo.1     0  0 0        1 1       0   0 1 0
##   textwo.2     0  0 0        1 1       0   0 0 1

dfm_group(dfmat)
## Document-feature matrix of: 2 documents, 9 features (38.9% sparse).
##          features
## docs      this is a sentence . another yet 1 2
##   textone    1  1 1        2 3       2   1 0 0
##   textwo     0  0 0        2 2       0   0 1 1
```

<sup>Created on 2021-03-20 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

kbenoit:
anyway, sleep on it. I'll merge this and we can adjust if we decide to make this change.

Issue #2097, opened 2021-03-21
upgrade_dfm() not working on some pre-`docid_` v2 objects


chainsawriot:
## Describe the bug

I encounter this problem while working with the [useNews](https://osf.io/uzca3/) (@cbpuschmann & @MarHai, 2020) dataset. As the data is not from me, it could be a data issue (e.g. the dfm is malformed). But displaying the dataset doesn't suggest so.

I can't directly pinpoint what is the actual cause of the error. `debug()` actually points to `dfm_match`, exactly [this line](https://github.com/quanteda/quanteda/blob/9ffe48cbab94f063048acac5f605b4322b57bc1e/R/dfm_match.R#L48). 

## Reproducible code

Please paste minimal code that reproduces the bug. If possible, please upload the data file as `.rds`.
```r
require(osfr)
require(quanteda)

osf_retrieve_node("uzca3") %>% osf_ls_files(n_max = 1, pattern = "usenews.mediacloud.wm.RData") %>% osf_download(path = ".", progress = TRUE)

load("usenews.mediacloud.wm.RData")

## mediacloud.wordmatrix2019 and mediacloud.wordmatrix2020 are a list of dfms.

mediacloud.wordmatrix2019[[1]] ## looks okay, it was created with v 1.5.2
mediacloud.wordmatrix2020[[1]] ## this one might be corrupted, but looks okay, was created with v 2.1.0 

dfm_lookup(mediacloud.wordmatrix2019[[1]], dictionary(list(Boston = c("boston")))) ## works
dfm_match(mediacloud.wordmatrix2019[[1]], "boston") ## works

dfm_lookup(mediacloud.wordmatrix2020[[1]], dictionary(list(Boston = c("boston")))) ## error
dfm_match(mediacloud.wordmatrix2020[[1]], "boston") ## print problem

```

## Expected behavior

It works for 2019 data but not for 2020 data. 2020 data should work.

## System information

```
R version 4.0.4 (2021-02-15)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.2 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.9.0
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.9.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
[...]

PR #2098, opened 2021-03-22
Issue 2097

koheiw:
Address #2097. Add check for docvars and enable upgrades of them only.


koheiw:
Is this what we want? I know why but don't think very useful.
```r
> is.dfm(fcm(tokens("a b c")))
[1] TRUE
```

kbenoit (approved):
💯 

On the `is.dfm`, we could change it to:

```r
is.dfm <- function(x) {
    is(x, "dfm") && !is(x, "fcm")
}
```

PR #2099, opened 2021-03-25
Fixes warning that occurs on Debian devel

kbenoit:
See code notes

### DIFF
--- a/R/quanteda_options.R
+++ b/R/quanteda_options.R
@@ -129,13 +129,13 @@ set_option_value <- function(key, value) {
             warning("Setting threads instead to maximum available ", thread["max"], call. = FALSE)
             value <- thread["max"]
         }
-        if (is.na(thread["tbb"])) {
+        # if (is.na(thread["tbb"])) {
             if (value != thread["max"])
                 RcppParallel::setThreadOptions(value)
-        } else {
-            if (value != thread["tbb"])
-                warning("Number of threads can be changed only once in a session", call. = FALSE)   
-        }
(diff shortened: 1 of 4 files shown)


R/quanteda_options.R
```diff
@@ -129,13 +129,13 @@ set_option_value <- function(key, value) {
             warning("Setting threads instead to maximum available ", thread["max"], call. = FALSE)
             value <- thread["max"]
         }
-        if (is.na(thread["tbb"])) {
+        # if (is.na(thread["tbb"])) {
```
kbenoit, line 132: Why did we need this condition to be TRUE before we call `RcppParallel::setThreadOptions(value)`? Because it can only be set once, and if not `NA`, then it has already been set?

R/quanteda_options.R
```diff
@@ -129,13 +129,13 @@ set_option_value <- function(key, value) {
...
             if (value != thread["max"])
```
kbenoit, line 133: And here, we don't set it if the user wants to use the max threads? But we don't issue any message?

R/quanteda_options.R
```diff
@@ -129,13 +129,13 @@ set_option_value <- function(key, value) {
...
                 RcppParallel::setThreadOptions(value)
-        } else {
-            if (value != thread["tbb"])
-                warning("Number of threads can be changed only once in a session", call. = FALSE)   
-        }
+        # } else {
```
kbenoit, line 135: See my email forwarded from KH.

Issue #2124, opened 2021-07-23
Add separator to phrase()


koheiw:
to allow 
```r
tokens_compound(toks_new, phrase(types(toks), separator  = "_"))
```
This will better address https://github.com/quanteda/quanteda/issues/1227.

PR #2129, opened 2021-08-14
Issue 2124

koheiw:
For #2124, add `separator` to `phrase()` and make `as.phrase()` for objects for which separator has no effect.


kbenoit (commented):
This is a good addition and move to make the coercion methods more consistent as well. However I see two things.

1. Better to deprecate or just make `noRd` the current `phrase()` methods that will be moved to `as.phrase()`, just so not to break existing code in a minor update. (I'm happy to do this if you want.)

2. I'm happy with the use the `separator` here. But I did take opportunity to review our various uses of `contatenator` versus `separator` to make sure our usage makes sense and is consistent. I think it is.

| Function                                      | `concatenator` | `separator` | (none) |
|-----------------------------------------------|:--------------:|:-----------:|:------:|
| `kwic()`                                      |                |      ✓      |        |
| `dictionary()`                                |                |      ✓      |        |
| `as.dictionary()`                             |                |      ✓      |        |
| `tokens_split()`                              |                |      ✓      |        |
| `tokens_ngrams()`                             |        ✓       |             |        |
| `tokens_skipgrams()`                          |        ✓       |             |        |
| `corpus_group()`                              |        ✓       |             |        |
| `spacyr::entity_consolidate()`                |        ✓       |             |        |
| `spacyr::entity_extract()`                    |        ✓       |             |        |
| `quanteda.textstats::textstat_collocations()` |                |             |    ✓   |

I think this is consistent because all of our functions that use `concatenator` are designed to take parts and put them together, using the value of `concatenator` to be the joining character. The functions that use `separator` on the other hand either take things that are already concatenated and split them using the value of `separator` to know what value to use for this split. 

So by that logic, for `as.phrase()` or `phrase()` the use of `separator` is correct.

koheiw:
I doubt that more than one or two people is using `phrase()` on these objects, but you are free to deprecate the methods.

kbenoit (approved):

PR #2136, opened 2021-09-14
Issue 2134

### LINKED ISSUE #2134: Error in grouping dfm
When grouping a dfm, I get an error (_Error in qatd_cpp_is_grouped_numeric(as.numeric(x), group) : Invalid groups_) with one docvar, while another docvar that is similar, does work. I can't figure out what the difference is and how to solve it.

koheiw:
Fix #2134

### DIFF
--- a/R/dfm_group.R
+++ b/R/dfm_group.R
@@ -69,8 +69,9 @@ dfm_group.dfm <- function(x, groups = docid(x), fill = FALSE, force = FALSE) {
 
     # remove NA groups
     x <- dfm_subset(x, !is.na(groups))
+    attrs <- attributes(x)
     groups <- groups[!is.na(groups)]
-
+    
     x <- group_matrix(x, documents = groups, fill = fill)
     build_dfm(x, colnames(x),
               unit = "documents",
--- a/R/docvars.R
+++ b/R/docvars.R
@@ -150,12 +150,8 @@ reshape_docvars <- function(x, i = NULL, unique = FALSE, drop_docid = TRUE) {
 # Reshape docvars keeping variables that have the same values within groups
 group_docvars <- function(x, group = NULL, field = NULL) {
     
-    if (is.null(group)) {
+    if (is.null(group))
         return(x)
-    } else {
-        x <- x[!is.na(group),]
-        group <- group[!is.na(group)]
-    }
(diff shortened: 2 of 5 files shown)


kbenoit (commented):
Looks good. I was thinking however whether we should add an option to group the NAs - for instance `use.na = TRUE`. I investigated a few similar functions, and they work like this:

* `split()` 
    Drops NA as a group.
    > Any missing values in f are dropped together with the corresponding values of x.

* `dplyr::group_split()`
    Keeps `NA` as a group 

``` r
df <- structure(list(
  x = 1:10,
  y = structure(c(
    NA, 1L, 1L, 1L, 1L,
    NA, 3L, 3L, 3L, 3L
  ),
  .Label = c("a", "b", "c"),
  class = "factor"
  )
),
row.names = c(NA, -10L), class = "data.frame"
)
df
##     x    y
## 1   1 <NA>
## 2   2    a
## 3   3    a
## 4   4    a
## 5   5    a
## 6   6 <NA>
## 7   7    c
## 8   8    c
## 9   9    c
## 10 10    c

# drops NA
split(df, df$y, drop = TRUE)
## $a
##   x y
## 2 2 a
## 3 3 a
## 4 4 a
## 5 5 a
## 
## $c
##     x y
## 7   7 c
## 8   8 c
## 9   9 c
## 10 10 c

# keeps NA
dplyr::group_split(df, y, .drop = TRUE)
## <list_of<
##   tbl_df<
##     x: integer
##     y: factor<af15a>
##   >
## >[3]>
## [[1]]
## # A tibble: 4 × 2
##       x y    
##   <int> <fct>
## 1     2 a    
## 2     3 a    
## 3     4 a    
## 4     5 a    
## 
## [[2]]
## # A tibble: 4 × 2
##       x y    
##   <int> <fct>
## 1     7 c    
## 2     8 c    
## 3     9 c    
## 4    10 c    
## 
## [[3]]
## # A tibble: 2 × 2
##       x y    
##   <int> <fct>
## 1     1 <NA> 
## 2     6 <NA>
```

koheiw:
This PR only makes `dfm_group()` to work in the same way as `tokens_group()` and `corpus_group()`. See 5f40e22 and compare with others.

Issue #2167, opened 2022-03-26
Make object class check more robust


koheiw:
It is easy to cheat `is.dfm()` and `is.dictionary()`. Need to check if `x` is S4 objects.
```r
> require(quanteda)
> d <- as.dfm(matrix(0))
> class(d)
[1] "dfm"
attr(,"package")
[1] "quanteda"
> isS4(d)
[1] TRUE
> is.dfm(d)
[1] TRUE
> 
> d2 <- matrix(0)
> class(d2) <- "dfm"
> isS4(d2)
[1] FALSE
> is.dfm(d2)
[1] TRUE

> ls <- list()
> class(ls) <- "dictionary2"
> is.dictionary(ls)
[1] TRUE
```

PR #2168, opened 2022-03-26
Issue 2167

koheiw:
Address #2167


koheiw:
Merged to `fix-flatten_dict`.

PR #2174, opened 2022-06-04
Fix pos

koheiw:
Fix a bug in ` tokens_select()`: only the first element is used when either `startpos` or `endpos` is a one-element vector.


kbenoit (approved):
Looks good. Would be a good practice to note this fix in `NEWS.md` too.

koheiw:
I improve the code more  and updated NEWS. Please merge if it is still good. It is time to update the CRAN.

PR #2182, opened 2022-08-26
Fixes for deprecated Matrix calls and for forthcoming Matrix 1.4-2

kbenoit:
- fixes deprecated `as()` calls (mostly `as(., "dgCMatrix")` which we'd used a lot)
- turns on errors for using deprecated versions using `options(Matrix.warnDeprecatedCoerce = 2)` in `testthat.R`
- works for forthcoming Matrix 1.4.2 release (tested from r-forge installation)

### DIFF
--- a/R/dfm-methods.R
+++ b/R/dfm-methods.R
@@ -144,7 +144,8 @@ matrix2dfm <- function(x, docvars = NULL, meta = NULL) {
     
     if (nrow(x) == 0 && ncol(x) == 0) {
         # avoid coercion to ldiMatrix
-        x <- as(as(as(matrix(nrow = 0, ncol = 0), "dMatrix"), "generalMatrix"), "unpackedMatrix")
+        x <- Matrix(x, sparse = TRUE)
+        x <- as(as(as(x, "CsparseMatrix"), "generalMatrix"), "dMatrix")
     } else {
         x <- Matrix(x, sparse = TRUE)
     }
--- a/R/dfm-methods.R
+++ b/R/dfm-methods.R
@@ -285,12 +285,12 @@ NULL
 
 #' The `Compare` methods enable relational operators to be use with dfm.
 #' Relational operations on a dfm with a numeric will return a
-#' [dgCMatrix-class][Matrix::dgCMatrix-class] object.
+#' [lgCMatrix-class][Matrix::lgCMatrix-class] object.
 #' @rdname dfm-internal
 #' @param e1 a [dfm]
 #' @param e2 a numeric value to compare with values in a dfm
 #' @export
 #' @seealso [Comparison] operators
 setMethod("Compare", c("dfm", "numeric"), function(e1, e2) {
-    as(callGeneric(as(as(as(e1, "CsparseMatrix"), "generalMatrix"), "dMatrix"), e2), "CsparseMatrix")
+    as(callGeneric(as(e1, "CsparseMatrix"), e2), "CsparseMatrix")
 })
--- a/R/dfm-methods.R
+++ b/R/dfm-methods.R
@@ -143,9 +143,7 @@ matrix2dfm <- function(x, docvars = NULL, meta = NULL) {
         meta <- make_meta("dfm")
     
     if (nrow(x) == 0 && ncol(x) == 0) {
-        # avoid coercion to ldiMatrix
-        x <- Matrix(x, sparse = TRUE)
-        x <- as(as(as(x, "CsparseMatrix"), "generalMatrix"), "dMatrix")
+        x <- make_null_dfm()
     } else {
         x <- Matrix(x, sparse = TRUE)
     }
--- a/R/dfm-methods.R
+++ b/R/dfm-methods.R
@@ -293,5 +293,5 @@ NULL
 #' @export
 #' @seealso [Comparison] operators
 setMethod("Compare", c("dfm", "numeric"), function(e1, e2) {
-    as(callGeneric(as(e1, "dgCMatrix"), e2), "lgCMatrix")
+    as(callGeneric(as(as(as(e1, "CsparseMatrix"), "generalMatrix"), "dMatrix"), e2), "CsparseMatrix")
 })
--- a/R/dfm-methods.R
+++ b/R/dfm-methods.R
@@ -144,13 +144,13 @@ matrix2dfm <- function(x, docvars = NULL, meta = NULL) {
     
     if (nrow(x) == 0 && ncol(x) == 0) {
         # avoid coercion to ldiMatrix
-        x <- as(matrix(nrow = 0, ncol = 0), "dgeMatrix")
+        x <- as(as(as(matrix(nrow = 0, ncol = 0), "dMatrix"), "generalMatrix"), "unpackedMatrix")
     } else {
         x <- Matrix(x, sparse = TRUE)
     }
     
     build_dfm(
-        as(x, "dgCMatrix"),
+        as(as(as(x, "CsparseMatrix"), "generalMatrix"), "dMatrix"),
         featname,
         docvars = docvars,
         meta = meta
--- a/R/dfm.R
+++ b/R/dfm.R
@@ -34,7 +34,8 @@
 #'     tokens_remove("b", padding = TRUE)
 #' toks
 #' dfm(toks)
-#' dfm(toks, remove = "") # remove "pads"
+#' dfm(toks) %>%
+#'  dfm_remove(pattern = "") # remove "pads"
 #'
 #' # preserving case
 #' dfm(toks, tolower = FALSE)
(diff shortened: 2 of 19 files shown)


kbenoit:
Tests above are failing but it does pass cleanly with Matrix 1.4-2. I need to test more with current CRAN release of Matrix too.

R/fcm.R
```diff
@@ -221,7 +223,7 @@ fcm.tokens <- function(x, context = c("document", "window"),
         type <- types(x)
         boolean <- count == "boolean"
         temp <- qatd_cpp_fcm(x, length(type), weights, boolean, ordered)
-        temp <- as(temp, "dgCMatrix")
+        temp <- as(as(as(temp, "CsparseMatrix"), "generalMatrix"), "dMatrix")
```
koheiw, line 226: I think we can switch between dgT and dgC only single `as()`. We know that `qatd_cpp_fcm()` returns only dgT.

``` r
require(Matrix)
#> Loading required package: Matrix
sessionInfo()
#> R version 4.1.2 (2021-11-01)
#> Platform: x86_64-pc-linux-gnu (64-bit)
#> Running under: KDE neon User - Plasma 25th Anniversary Edition
#> other attached packages:
#> [1] Matrix_1.4-2
#> 
smat <- sparseMatrix(i = c(1, 3, 3), j = c(1, 1, 2), x = c(10, 20, 30), dims = c(3, 4))
class(smat)
#> [1] "dgCMatrix"
#> attr(,"package")
#> [1] "Matrix"
smat2 <- as(smat, "CsparseMatrix")
class(smat2)
#> [1] "dgCMatrix"
#> attr(,"package")
#> [1] "Matrix"
smat3 <- as(smat2, "TsparseMatrix")
class(smat3)
#> [1] "dgTMatrix"
#> attr(,"package")
#> [1] "Matrix"
```

R/object-builder.R
```diff
@@ -27,9 +27,9 @@ build_dfm <- function(x, features,
                       docvars = data.frame(), meta = list(), 
                       class = "dfm", ...) {
     result <- new(class,
-        as(x, "dgCMatrix"),
-        docvars = docvars,
-        meta = make_meta("dfm", inherit = meta, ...)
+                  as(as(as(x, "CsparseMatrix"), "generalMatrix"), "dMatrix"),
```
koheiw, line 30: DFM is dgC or dgT when created internally. It is necessary to apply `as(as(as(x, "CsparseMatrix"), "generalMatrix"), "dMatrix")` only when `x` is unknown format. We probably need triple coercion for `as.dfm()` where `x` can be any matrix.

R/dfm-methods.R
```diff
@@ -144,13 +144,13 @@ matrix2dfm <- function(x, docvars = NULL, meta = NULL) {
     
     if (nrow(x) == 0 && ncol(x) == 0) {
         # avoid coercion to ldiMatrix
-        x <- as(matrix(nrow = 0, ncol = 0), "dgeMatrix")
+        x <- as(as(as(matrix(nrow = 0, ncol = 0), "dMatrix"), "generalMatrix"), "unpackedMatrix")
     } else {
         x <- Matrix(x, sparse = TRUE)
```
koheiw, line 149: It might be safe enough if we ensure that `x` is a dgC or  dgT Matrix here like:
```r
x <- Matrix(x, sparse = TRUE)
x <- as(as(as(x, "CsparseMatrix"), "generalMatrix"), "dMatrix")
```
Then we need only `as(x, "CsparseMatrix")` or `as(x, "TsparseMatrix")` in other functions.

kbenoit: I also changed 147 to `make_null_dfm()` since we already have a function for that.

R/object-builder.R
kbenoit: I left it here since it achieves the purpose of the original code, which was to ensure that build_dfm returns a dgCMatrix. I think you were suggesting a broader change that this glimpse at the code revealed. First things first - getting Matrix working. Did I misunderstand?

koheiw: Misunderstanding. I am suggesting to make simplify the code to make it work with Matrix 1.4-2: if `x` is dgT or dgC, you need only `as(x, "CsparseMatrix")`. `as(as(x, "generalMatrix"), "dMatrix")` is redundant.

R/dfm.R
```diff
@@ -383,7 +384,9 @@ make_null_dfm <- function(feature = NULL, document = NULL) {
         i = NULL,
         j = NULL,
         dims = c(length(document), length(feature))
-    ), "dgCMatrix")
+    ), "CsparseMatrix") %>%
```
koheiw, line 387: Please do not use pipes in the code for me. I use them in the global environment but never in packages.

R/dfm-methods.R
```diff
@@ -293,5 +292,5 @@ NULL
 #' @export
 #' @seealso [Comparison] operators
 setMethod("Compare", c("dfm", "numeric"), function(e1, e2) {
-    as(callGeneric(as(e1, "dgCMatrix"), e2), "lgCMatrix")
+    as(callGeneric(as(as(as(e1, "CsparseMatrix"), "generalMatrix"), "dMatrix"), e2), "CsparseMatrix")
```
koheiw, line 295: I wonder why you need to apply as, because DFM is always a dgCMatrix. Do you need to strip the DFM class to call the generic or something?

R/fcm-methods.R
```diff
@@ -142,7 +142,7 @@ matrix2fcm <- function(x, meta = NULL) {
     if (is.null(meta))
         meta <- make_meta("fcm")
     
-    build_fcm(as(Matrix(x, sparse = TRUE), "dgCMatrix"),
+    build_fcm(as(as(Matrix(x, sparse = TRUE), "generalMatrix"), "dMatrix"),
```
koheiw, line 145: I don't think you need `as()` here because you added `as(as(as(x, "CsparseMatrix"), "generalMatrix"), "dMatrix")` to  `build_fcm()`.

R/fcm.R
```diff
@@ -164,22 +165,23 @@ fcm.dfm <- function(x, context = c("document", "window"),
...
+        Matrix::diag(temp) <- (Matrix::diag(temp) - m) / 2
     }
-    if (tri) 
+    if (tri)
         temp <- Matrix::triu(temp)
 
     result <- build_fcm(
-        as(temp, "dgCMatrix"), featnames(x), 
-        count = count, context = context, margin = featfreq(x), 
+        as(temp, "CsparseMatrix"),
```
koheiw, line 179: I don't think you need `as()` here becasue it is done in `build_fcm()`.

R/fcm.R
```diff
@@ -230,9 +232,10 @@ fcm.tokens <- function(x, context = c("document", "window"),
             }
         }
         result <- build_fcm(
-            as(temp, "dgCMatrix"), type,
-            count = count, context = context, margin = featfreq(dfm(x, tolower = FALSE)), 
-            weights = weights, ordered = ordered, tri = tri, 
+            as(temp, "CsparseMatrix"),
```
koheiw, line 235: I don't think you need `as()` here becasue it is done in `build_fcm()`.

R/object-builder.R
```diff
@@ -27,9 +27,9 @@ build_dfm <- function(x, features,
                       docvars = data.frame(), meta = list(), 
                       class = "dfm", ...) {
     result <- new(class,
-        as(x, "dgCMatrix"),
-        docvars = docvars,
-        meta = make_meta("dfm", inherit = meta, ...)
+                  as(x, "CsparseMatrix"),
```
koheiw, line 30: If you add `as(as(as(x, "CsparseMatrix"), "generalMatrix"), "dMatrix")` in `build_fcm()`, let's do the same here too. This may contradict my earlier comment, but I think it we are better off if the two object builders work consistently.

tests/testthat.R
```diff
@@ -4,6 +4,9 @@ Sys.setenv("_R_CHECK_LENGTH_1_CONDITION_" = TRUE)
 library(testthat)
 library(quanteda)
 
+# for strong texts for Matrix deprecations
```
koheiw, line 7: Do you mean "tests"?

R/dfm-methods.R
kbenoit: Yes we do, but it could be accomplished more simply, so I changed it to 
```r
setMethod("Compare", c("dfm", "numeric"), function(e1, e2) {
    as(callGeneric(as(e1, "CsparseMatrix"), e2), "CsparseMatrix")
})
```

also I updated the documentation to reflect that it always returns an lgCMatrix.

R/fcm-methods.R
kbenoit: True. Hard to keep track of when it's needed. Was used before so I added it, but it wasn't necessary so I made it just 
```r
build_fcm(x, rowname, colname, meta = meta)
```
now.

R/object-builder.R
kbenoit: You're right, better to make sure here, since we want only a dgCMatrix returned, and not some other version of CsparseMatrix. It doesn't hurt to be sure.

koheiw (approved):
Looks good to me. Thank you for your work.

PR #2185, opened 2022-10-01
Fix potential cause of segfault

koheiw:
Use multiplication instead to bitshift


kbenoit:
Is this PR active? I saw you had some activity on it but then it was closed.

koheiw:
I was looking into the FCM issue in an example in my package. My guess was bitshif and multiplication are slightly different in TBB, but not really sure. We do not see segfault anywhere else, so it could be just false positive.
https://www.stats.ox.ac.uk/pub/bdr/donttest/LSX.out
I can reopen this PR if you want.

PR #2196, opened 2022-12-15
Build to update Google Analytics code

koheiw:
I need to rebuild the site to update the GA tracking code in the header:

https://support.google.com/analytics/answer/10089681


kbenoit (approved):
Thanks Kohei!

Issue #2199, opened 2023-01-11
How best we can make artificial paragraphs?


koheiw:
Let's imagine we need to make artificial paragraphs by combining 2 adjacent sentences, run some operation of them and restore original documents. This might be a rare case, but raises a few questions on quanteda's functions. 

I came up with the following pipeline, but it is too complex. So I propose to make a dedicated function for grouping segments of texts, although I do not know how we should call it.

At least, we should add `segid()` and `tokens_group(reset_docid = FALSE)`. With the the first, we can avoid accessing the system docvar directly, `(attr(toks, "docvars")$segid_`; with the second, we don't need to back up the original docid, `toks$docid_org <- docid(toks)`.

``` r
require(quanteda)
#> Loading required package: quanteda
#> Package version: 3.2.4
#> Unicode version: 13.0
#> ICU version: 66.1
#> Parallel computing: 4 of 4 threads used.
#> See https://quanteda.io for tutorials and examples.
dat <- data.frame(text = c("A aaa aaa. Aa aaaa.",  
                           "B bb bbb bb. Bbb b. B bbb. Bb bb bb."),
                  var = c(10, 20))
corp <- corpus(dat) %>% 
    corpus_reshape()
toks <- tokens(corp)
toks
#> Tokens consisting of 6 documents and 1 docvar.
#> text1.1 :
#> [1] "A"   "aaa" "aaa" "."  
#> 
#> text1.2 :
#> [1] "Aa"   "aaaa" "."   
#> 
#> text2.1 :
#> [1] "B"   "bb"  "bbb" "bb"  "."  
#> 
#> text2.2 :
#> [1] "Bbb" "b"   "."  
#> 
#> text2.3 :
#> [1] "B"   "bbb" "."  
#> 
#> text2.4 :
#> [1] "Bb" "bb" "bb" "."
docid(toks)
#> [1] text1 text1 text2 text2 text2 text2
#> Levels: text1 text2
docvars(toks)
#>   var
#> 1  10
#> 2  10
#> 3  20
#> 4  20
#> 5  20
#> 6  20

# back up the original docid
toks$docid_org <- docid(toks) 

f <- paste0(docid(toks), ".", ceiling(attr(toks, "docvars")$segid_ / 2))
f <- factor(f, unique(f))

# chunk sentences to create artificial paragraphs
toks_grp <- tokens_group(toks, f) # the original docid is overritten
toks_grp
#> Tokens consisting of 3 documents and 2 docvars.
[...]

PR #2201, opened 2023-01-13
Add segid()

koheiw:
Add `segid()` to discourage users to access directly to the system docvars (#2199)


koheiw:
@kbenoit I am not sure what is happening in Github actions, but this PR should pass all the checks.

kbenoit:
> @kbenoit I am not sure what is happening in Github actions, but this PR should pass all the checks.

Not sure either, but the reason I disabled that test on Linux was because in late November, our build started failing on the linux dev branches due to ongoing changes in the **Matrix** package (which caused us issues since late summer) and due to some R changes. I could not figure out the nuances from the tests, which had to do with Matrix object class coercions and how dimension names were preserved (or not), so I simply disabled the test on the failing platforms since I was getting dire warnings in early December from CRAN about the package being removed. I agree it's time to sort this out now. Sometimes things start breaking but later seem to be fixed again for the dev builds.

kbenoit:
> @kbenoit I am not sure what is happening in Github actions, but this PR should pass all the checks.

Looks like codecov is failing and the ubuntu tests are not running as a result?
https://app.codecov.io/gh/quanteda/quanteda/pull/2201

koheiw:
Matrix should be fine on Linux, at least on my machens. I will keep eyes on it.

Codecov of the project is irrelevant because the patch is 100%. Why don't we only use the patch coverage?

kbenoit:
I'm happy with turning off the project codecov!

koheiw:
The PR was not passing the tests because tests were not running. I fixed the Github action file and remove codecov/project from required list (although its is showing +0.02% increase this time).

kbenoit (approved):

Issue #2208, opened 2023-02-05
Upgrade tokens object to process larger data more efficiently


koheiw:
Hi all,

I have finished feasibility tests of upgrading tokens object to process larger a corpus with smaller memory footprint and shorter execution time. The discussion on this subject has a long history (#681), but, in short, the new tokens object is based on Rcpp's external pointer, the `XPtr` object, to keep the large vectors in the C++ memory space.

It is hugely beneficial to keep tokens in the C++ side because every time we copy C++ vectors to R vectors, it takes a few seconds consuming a huge memory space. Repeated creation and deletion of large objects also trigger R's garbage collector, which also takes more than a few second to do its job.

The example below compares the old tokens object `toks` and new (prototype) tokens object `xtoks` through a long pipeline, which (I think) is common. Since `tokens_remove()` is called four times, `toks` travels between C++ and R eight times; while `xtoks` does not at all. As a result, operations on the new object is nearly three times faster. 

`as.externalptr(toks)` converts the old tokens object to the new tokens object (I probably should give the function a better name); `xtoks`. `as.externalptr(xtoks)` deep-copies the new tokens object, so `xtoks` is not modified in the test.

Please install the branch with this command and tests the new object:
```r
devtools::install_github("quanteda/quanteda", ref = "test-xtokens")
```
The Guardian corpus is [avaialble](https://www.dropbox.com/s/kfhdoifes7z7t6j/data_corpus_guardian2016-10k.rds?dl=1), but please use a larger corpus if you have any. You should see a bigger difference between the old and new if you use tokens objects with many short texts, like sentences or social media posts.

I would appreciate anyone's feedback/comment, but especially interested in @conjugateprior @stefan-mueller @amatsuo @pablobarbera.

``` r
require(quanteda)
#> Loading required package: quanteda
#> Package version: 3.2.5
#> Unicode version: 14.0
#> ICU version: 70.1
#> Parallel computing: 4 of 4 threads used.
#> See https://quanteda.io for tutorials and examples.

corp <- readRDS('/home/kohei/Dropbox/Public/data_corpus_guardian2016-10k.rds') %>% 
    corpus_reshape()
ndoc(corp)
#> [1] 391395

toks <- tokens(corp, remove_punct = FALSE, remove_numbers = FALSE, 
               remove_symbols = FALSE)
xtoks <- as.externalptr(toks) # convert to a new tokens object
class(xtoks)
#> [1] "externalptr"
class(toks)
#> [1] "tokens"

[...]

PR #2210, opened 2023-02-15
Add tokens_xptr object

koheiw:
For #2208, I created a tokens_xptr object, a sub-class of tokens, whose underlying object is an external pointer instead of a list. All the underlying C++ code for `tokens_*` functions are also upgraded to accept the new object. 

Users can switch between tokens and tokens_xptr objects via `as.tokens()` and `as.tokens_xptr()` whenever they want to. If they apply `tokens_*` functions to old objects, `as.tokens_xptr()` is applied to the input and `as.tokens()` to the output internally, so they do not even notice the changes in underlying functions.

Apart from the detail of the implementation, remaining design issues are when the `tokens_xptr` methods should return a deep copy of the input objects. Currently, a deep copy is returned after subsetting (e.g. `[]`, `tokens_subset()` and `tokens_sample()`).

We can create `tokens_xptr.corpus()`, or `tokens.corpus(xptr = TRUE)`, to get tokens_xptr objects straight away. `tokens.tokens()` is already a wrapper around `tokens.tokens_xptr()`.

We should also add `env` to functions that perform NSE just like I did in `tokens_group()` and `tokens_sample()`. Otherwise, developers find it difficult to nest them in their functions.

For now, I kept the old code for `dfm.tokens()` and `tokens_recompile.tokens()` to compare results.


kbenoit (commented):
Really impressed by all the work you're putting in here @koheiw and looking forward to reviewing these changes in detail. I'm teaching for three more mornings then will have the clarity to give it a thorough review and test. Thanks!

koheiw:
@kbenoit please let me know long it will take for you to finish reviewing this PR.

kbenoit (commented):
Sorry @koheiw, we lost some staff and it's been crazy at DSI. I'll have this looked at this week and let's try to finalise it by next. Promise to get to it v soon - and thanks for the amazing work on this improvement!

koheiw:
I am asking because it will take full one or two week or so to properly review this PR. This is also the base for other substantive changes like [doing serializetion to C++](https://github.com/quanteda/quanteda/compare/dev-xtokens...test-tokenize).

koheiw:
@kbenoit  please test and review this in parallel to all otehrs but don't merge.

kbenoit:
> @kbenoit please test and review this in parallel to all others but don't merge.

@koheiw  `master` is up-to-date now, but I will let you update this branch by merging `master` into it. You might need to update `v4` too. Let me know when the merging is done and it's ready for review.

src/serialize.cpp
```diff
@@ -0,0 +1,123 @@
...
+    
+    //dev::Timer timer;
+    Types types = xptr->types;
+    StringTexts texts = Rcpp::as<StringTexts>(texts_);
+    
+    //dev::start_timer("Register", timer);
+    MapTypes map;
+    for (size_t g = 0; g < types.size(); g++) {
+        std::string type = types[g];
+        auto it = map.insert(std::pair<std::string, UintParam>(types[g], g + 1));
```
kbenoit, line 82: I'm seeing this compiler warning:

```
C  -falign-functions=64 -Wall -g -O2  -UNDEBUG -Wall -pedantic -g -O0 -c serialize.cpp -o serialize.o
   serialize.cpp:82:14: warning: unused variable 'it' [-Wunused-variable]
           auto it = map.insert(std::pair<std::string, UintParam>(types[g], g + 1));
                ^
   1 warning generated.
```

DESCRIPTION
kbenoit, line 1: do we still need to specify C++11? The CRAN check results suggest we not specify this if not essential.

R/tokens_xptr.R
```diff
@@ -0,0 +1,209 @@
+#' @rdname tokens_xptr
+#' @export
+is.tokens_xptr <- function(x) {
+    identical(typeof(x), "externalptr") && "tokens_xptr" %in% class(x)
+}
+
+#' Convert a tokens object to a tokens_xptr object
+#'
+#' tokens_xptr is a external pointer object to process a large number of
```
kbenoit, line 9: This needs more description here, about the implications of converting tokens to an xptr type, and what would need to be done in order to work with such objects, including how the workspace is saved.

We could also list the methods here for which this class is now defined, and explain any differences that the user should know.

The examples should show how to go to and from the xptr space and a few manipulations using standard tokens methods in the middle.

R/tokens_group.R
```diff
@@ -22,24 +24,26 @@
...
+tokens_group.tokens_xptr <- function(x, groups = docid(x), fill = FALSE, env = NULL) {
+    
+    if (is.null(env))
+        env <- parent.frame()
     if (missing(groups)) {
         field <- NULL
         groups <- docid(x)
     } else {
         field <- deparse(substitute(groups))
-        groups <- eval(substitute(groups), get_docvars(x, user = TRUE, system = TRUE), parent.frame())
```
kbenoit, line 42: I think we specified `parent.frame()` to make sure that this would work in other functions or environments, not just the base environment. Have you tested this under such conditions?

R/tokens_group.R
```diff
@@ -22,24 +24,26 @@
...
+        groups <- eval(substitute(groups), get_docvars(x, user = TRUE, system = TRUE), env)
```
kbenoit, line 46: Under what circumstances would users want to specify `env`?

kbenoit (commented):
This is a potentially game-changing improvement that could drastically improve performance.   You've done an fantastic job of making this work within the existing function ecosystem in a way that does not affect current usage.

It would be very important here to develop a vignette describing what are the external pointer tokens objects and when and how to use them. There are also important user considerations to know about, such as finishing a session and the fact that the .RData workspace will not save the external pointer references, nor will they be restored if R should crash or be terminated by the user. **data.frame** I think works this way, so we could examine their approach.

The vignette would be an excellent place also to showcase the performance gains from using xptr tokens.

See my code notes on the addition of `env` to functions such as `dfm_group()`. We should test this for robustness, since the default in your change is the base environment, not the parent environment. This will affect variables that are in a parent.frame() but not in the docvars, for instance, when the parent.frame() is not the base environment. It could affect developers, in other words.

I was thinking that you added `env` because the object to which the external pointer refers does not include the docvars data.frame? But  tokens_xptr object seems to include the standard attributes (including docvars) as regular objects, no? I was not sure why this approach is needed to get the NSE for docvars.

`NEWS.md` will need updating, of course.

> Apart from the detail of the implementation, remaining design issues are when the tokens_xptr methods should return a deep copy of the input objects. Currently, a deep copy is returned after subsetting (e.g. [], tokens_subset() and tokens_sample()).

The alternative would be to make this happen in the external pointer space, right? I think the key would be for `[`, and the other methods should then be easy. How tricky is redefining `[` for this?

Finally, consider this. Since we are introducing this as a parallel set of functions, which a normal user could ignore completely, what if we released this as a 3.3 or 3.4, with these functions as experimental, and we could then see how users get along with it? 4.0 would then come a few months later, with the full fanfare, as well as other improvements, removal of deprecations, etc.

R/tokens.R
kbenoit, line 1: Some conflicts in this file indicated in merging the newly updated `master` into this branch, please resolve and merge to update this branch.

kbenoit, line 1: Some conflicts in this file indicated in merging the newly updated master into this branch, please resolve and merge to update this branch.

R/tokens_xptr.R
```diff
@@ -0,0 +1,209 @@
...
+        e <- substitute(subset)
+        r <- eval(e, docvar, parent.frame())
+        r & !is.na(r)
+    }
+    return(x[r, drop_docid = drop_docid])
+}
+
+#' @method dfm tokens_xptr
+#' @export
+dfm.tokens_xptr <- function(x, tolower = TRUE, remove_padding = FALSE, ...) {
```
kbenoit, line 145: 👍 👍 👍 This is a real game changer, since compiling a dfm for a large tokens object can be hugely time-consuming.

R/tokens_group.R
koheiw: I needed to pass `env` from `tokens_group.tokens()` to `tokens_group.tokens_xptr()` because the global environment is not the parent frame for `tokens_group.tokens_xptr()` (it is grand parent frame). I am no expert on NSE so you are welcome to suggest a better approach.

More broadly, it is hard to use `tokens_group()` in custom functions if `eval()` searches only the parent frame. `env` is what Wickham calls "[escape hatch](http://adv-r.had.co.nz/Computing-on-the-language.html)".

R/tokens_xptr.R
koheiw: I can improve the documentation once we fixed the design. Why don't we remove old tutorials in vignettes, so that we can include more important (technological) things? @stefan-mueller can you decide which articles to keep using Google Analytics and updated them?

R/tokens_xptr.R
koheiw: It is the time to defunct all the Swiss Army arguments of `dfm()`.

R/tokens.R
koheiw: Resolved the conflicts

R/tokens_xptr.R
kbenoit: We only have one vignette (`quickstart.Rmd`), so this would classify as an article unless we included it explicitly as a vignette. But it might be time to create one for performance issues and how to optimise performance with large objects. This could introduce xtokens but also discuss multi-threading, and do some comparisons with other packages.

R/tokens_xptr.R
kbenoit: Agreed but let's leave all deprecations and defuncting to a separate PR.

R/tokens_group.R
kbenoit: That's exactly right, and our current approach also provides that escape hatch. 

But I also need to study this more. I just ran a bunch of tests and while I thought that if not found in the docvars, it would search the calling environment, that does not appear to be the case. It's also not how the tidyverse method in **dplyr** works either.

``` r
myfun1 <- function(x) {
    logvar <- sample(c(TRUE, FALSE), 
                     size = nrow(x),
                     replace = TRUE)
    myfun2(x)
}

myfun2 <- function(y) {
    dplyr::filter(y, logvar)
}

# fails
myfun1(iris)
#> Error in `dplyr::filter()`:
#> ℹ In argument: `logvar`.
#> Caused by error:
#> ! object 'logvar' not found
#> Backtrace:
#>      ▆
#>   1. ├─global myfun1(iris)
#>   2. │ └─global myfun2(x)
#>   3. │   ├─dplyr::filter(y, logvar)
#>   4. │   └─dplyr:::filter.data.frame(y, logvar)
#>   5. │     └─dplyr:::filter_rows(.data, dots, by)
#>   6. │       └─dplyr:::filter_eval(...)
#>   7. │         ├─base::withCallingHandlers(...)
#>   8. │         └─mask$eval_all_filter(dots, env_filter)
#>   9. │           └─dplyr (local) eval()
#>  10. └─base::.handleSimpleError(...)
#>  11.   └─dplyr (local) h(simpleError(msg, call))
#>  12.     └─rlang::abort(message, class = error_class, parent = parent, call = error_call)

# works
logvar <- sample(c(TRUE, FALSE), 
                 size = nrow(iris),
                 replace = TRUE)
myfun1(iris) |>
    tibble::tibble()
#> # A tibble: 79 × 5
#>    Sepal.Length Sepal.Width Petal.Length Petal.Width Species
#>           <dbl>       <dbl>        <dbl>       <dbl> <fct>  
#>  1          5.1         3.5          1.4         0.2 setosa 
#>  2          4.9         3            1.4         0.2 setosa 
#>  3          4.7         3.2          1.3         0.2 setosa 
#>  4          5.4         3.9          1.7         0.4 setosa 
#>  5          4.6         3.4          1.4         0.3 setosa 
#>  6          5           3.4          1.5         0.2 setosa 
#>  7          4.4         2.9          1.4         0.2 setosa 
#>  8          4.9         3.1          1.5         0.1 setosa 
#>  9          4.8         3            1.4         0.1 setosa 
#> 10          4.3         3            1.1         0.1 setosa 
#> # ℹ 69 more rows
```

<sup>Created on 2023-04-02 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

tests/testthat/test-spacyr-methods.R
```diff
@@ -6,14 +6,6 @@ test_that("test quanteda methods for spacy_parsed objects", {
     expect_identical(ntoken(data_spacy_parsed), c(text1 = 7L))
 })
 
-test_that("test nsentence for spacy_parsed objects", {
```
kbenoit, line 9: See prior comment on nfunctions

R/nfunctions.R
kbenoit, line 1: I agree that `nsentence()` is problematic but I don't think we should remove it here. 

1. We would need to deprecate rather than remove this, plus run a bunch of revdepchecks to see if we are breaking anyone's code before changing the behaviour of these functions.
2. Let's leave deprecations and removals for v4 to a different PR, so that we could potentially release xptrs into a 3.x release and so that at least we can isolate the effects of the v4 pruning.

R/nfunctions.R
```diff
@@ -188,55 +189,3 @@ ntype.tokens <- function(x, ...) {
...
-
-#' @export
-#' @importFrom stringi stri_detect_charclass
-nsentence.character <- function(x) {
-    upcase <- try(any(stri_detect_charclass(x, "[A-Z]")), silent = TRUE)
-    if (!is.logical(upcase)) {
-        # warning("Input text contains non-UTF-8 characters.")
-    } else if (!upcase)
-        warning("nsentence() does not correctly count sentences in all lower-cased text")
-    ntoken(tokens(x, what = "sentence"))
```
kbenoit, line 226: `nsentence.character()` and `nsentence.corpus()` are good examples of the behaviour we should deprecate or remove asap, since they have to tokenise non-tokenised objects on the fly. (Same as with `summary.corpus()`.) We should be only returning summaries on tokenised objects that are already tokenised by the user, on principle. I think we agree 100% on this. The question is just how to phase it out without breaking dependencies or alienating users too badly.

kbenoit:
@koheiw you probably had discovered this already, but after I restored the `nsentence()` functions, I began getting this, but only on `devtools::test()`, not from running the tests interactively or from `test_file("tests/testthat/test-nfunctions.R")`. This is also what is failing in CI. It seems to do with the new serializer function. Can we patch this so that `nsentence()` still works robustly?

```
     ══ Failed tests ════════════════════════════════════════════════════════════════
     ── Error ('test-nfunctions.R:75:5'): test nsentence ────────────────────────────
     Error in `nsentence.tokens(tokens(txt, what = "sentence"))`: nsentence on a tokens object only works if what = "sentence"
     Backtrace:
         ▆
      1. ├─testthat::expect_identical(...) at test-nfunctions.R:75:4
      2. │ └─testthat::quasi_label(enquo(object), label, arg = "object")
      3. │   └─rlang::eval_bare(expr, quo_get_env(quo))
      4. ├─quanteda::nsentence(tokens(txt, what = "sentence"))
      5. └─quanteda:::nsentence.tokens(tokens(txt, what = "sentence"))
```

R/tokens_group.R
koheiw: I will leave this issue with you.

koheiw:
I am strongly against releasing `tokens_xptr` as part of v3. It is the biggest change in tokens object and its functions, so there are too many ways to go wrong badly.

kbenoit:
I understand, and it would only make sense if used optionally (as `what = "word4"` for example) rather than some functions depending on them, as they do in this PR.

R/nfunctions.R
koheiw: I want to remove the function as soon as possible, but fixed the error.

kbenoit:
Should we merge `master` into `v4` before we check the ability of this branch to merge into `v4`?

R/tokens_xptr.R
kbenoit: I'd suggest at least updating `NEWS.md` for this PR. We can add a performance vignettes later as we near the 4.0 release.

R/tokens.R
```diff
@@ -157,13 +158,10 @@ tokens <-  function(x,
                     include_docvars = TRUE,
                     padding = FALSE,
                     verbose = quanteda_options("verbose"),
-                    ...) {
+                    ...,
+                    xptr = FALSE) {
```
kbenoit, line 162: I like this, it's simpler, and suggests we could flip it to `TRUE` eventually.
I know you dislike the idea of testing this in a 3.x version, but if we sneaked it in this way, we could allow tests and comparisons with just one toggle. Maybe 4.0 would set it to `TRUE`?

R/tokens_xptr.R
kbenoit: Resolving now but still think we need a performance vignette.

kbenoit: If we want to pare the articles in the pkgdown site, we should open a separate issue.

R/tokens_xptr.R
```diff
@@ -0,0 +1,209 @@
...
+#' documents efficiently.
+#' @param x a [tokens] object to convert or a tokens_xptr object to
+#'   to deep-copy.
+#' @rdname tokens_xptr
+#' @export
+as.tokens_xptr <- function(x) {
+    UseMethod("as.tokens_xptr")
+}
+
+#' @rdname tokens_xptr
```
kbenoit, line 19: We should collapse these into just `as.tokens_xtpr`, the way we do `as.tokens`, so there are not the object-specific methods in the package index.

R/tokens.R
koheiw: I have to see how many people will really use the `tokens_xptr` object. If a corpus contains only a few thousands documents, it is easier to use the classic object.

R/tokens_group.R
koheiw: I think these tests covers all the possible cases. 

https://github.com/quanteda/quanteda/blob/d737996e45061fddb36432a6d104177f445b14b0/tests/testthat/test-tokens_group.R#L127-L156

kbenoit:
I like the addition of the `xptr` logical in `tokens()`. One thing about naming, it suggests we should name this new object `tokensx`, since our grammar is `objectclass_verb`, so that `tokens_xptr` suggests a function that takes in and returns a tokens object.

So the constructor and object class could just be called `tokensx()` and `tokensx`, although I like `tokens(x, xptr = TRUE)` even better.

What do you think?

R/tokens_xptr.R
```diff
@@ -0,0 +1,209 @@
...
+# }
+# 
+# #' @export
+# segid.tokens_xptr <- function(x) {
+#     get_docvars(x, "segid_", FALSE, TRUE, TRUE)
+# }
+
+#' @method as.tokens tokens_xptr
+#' @export
+as.tokens.tokens_xptr <- function(x, ...) {
```
kbenoit, line 74: Should this (and other functions) have a `concatenator` argument, as do the other `as.tokens()` methods?

koheiw:
> I like the addition of the `xptr` logical in `tokens()`. One thing about naming, it suggests we should name this new object `tokensx`, since our grammar is `objectclass_verb`, so that `tokens_xptr` suggests a function that takes in and returns a tokens object.

`tokens_xptr` is a `sub-class` of tokens, so there is no problem for `tokens()` to return the object.
```
> class(tokens("aa", xptr = TRUE))
[1] "tokens_xptr" "tokens" 
```

R/tokens_xptr.R
koheiw: Unnecessary because `x` can only be `tokens`. Conversion of a list will be `as.tokens_xptr(as.tokens(list(), concatenator = "+"))`.

R/tokens_xptr.R
koheiw: Done.

R/tokens_xptr.R
koheiw: I can write an article on xptr only after merging this branch.

stefan-mueller: Happy to get involved in revising the existing vignettes at [quanteda.io](https://quantda.io) and editing the new article on xptr and multithreading. Probably worth discussing details in a separate issue.

kbenoit (approved):

Issue #554, opened 2017-02-15
dfm error with special characters


jhfowler:
i am trying to turn a list of tweets into a dfm and it occasionally creates an error on special characters.  here's an example:

> library(quanteda)
> x <- dfm(corpus("あﾞづいﾞ"))
Error in Matrix::sparseMatrix(i = index, p = cumsum(c(1, lengths(x))) -  : 
  NA's in (i,j) are not allowed

jhfowler:
On further inspection, it seems like the error is always caused by Hiragana.  Strings that fail: "づいﾞ",
"゛んﾞ","# たーﾟ"

kbenoit:
OK, definitely one for @koheiw. It's an issue with the hashing function:
```r
> tokens(c("づいﾞ", "゛んﾞ", "たーﾟ"))
tokens from 3 documents.
Component 1 :
[1] "づ"  NA    "いﾞ"

Component 2 :
[1] "゛"  NA    "んﾞ"

Component 3 :
[1] "た"  NA    "ーﾟ"

> tokens(c("づいﾞ", "゛んﾞ", "たーﾟ"), hash = FALSE)
tokenizedTexts from 3 documents.
Component 1 :
[1] "づ"  ""    "いﾞ"

Component 2 :
[1] "゛"  ""    "んﾞ"

Component 3 :
[1] "た"  ""    "ーﾟ"
```

jhfowler:
Thanks!  BTW I am really loving the package.   This is the first snag I've hit.
j

kbenoit:
Thanks Jim!

Testimonials always welcome at https://github.com/kbenoit/quanteda/issues/461.

Ken


On 16 Feb 2017, at 23:42, jhfowler <notifications@github.com<mailto:notifications@github.com>> wrote:

Thanks! BTW I am really loving the package. This is the first snag I've hit.
j

Issue #420, opened 2017-01-07
Create regex2id function


koheiw:
It is wasteful to get character from regex2fixed5 and convert to token IDs again in the `tokens_*` family.

kbenoit:
This is fixed in https://github.com/kbenoit/quanteda/commit/b1c8b49f1ade25fccf7571204de1d928c607d91e?

PR #2216, opened 2023-03-19
Incoporate RBBI tokenizer for v4.0

koheiw:
I drafted `tokenize_word4()` based on #2165. I want to make it the main tokenizer of quanteda v4.0. It is made available via `tokens(what = "word4")` too.

``` r
require(quanteda)
#> Package version: 3.2.5
#> Unicode version: 14.0
#> ICU version: 70.1
#> Parallel computing: 4 of 4 threads used.
tokenize_word4("a well-known website http://example.com #hashtag @username", 
               split_hyphens = FALSE, split_tags = FALSE)
#> [[1]]
#>  [1] "a"                  " "                  "well-known"        
#>  [4] " "                  "website"            " "                 
#>  [7] "http://example.com" " "                  "#hashtag"          
#> [10] " "                  "@username"
tokenize_word4("a well-known website http://example.com #hashtag @username", 
               split_hyphens = TRUE, split_tags = TRUE)
#> [[1]]
#>  [1] "a"                  " "                  "well"              
#>  [4] "-"                  "known"              " "                 
#>  [7] "website"            " "                  "http://example.com"
#> [10] " "                  "#"                  "hashtag"           
#> [13] " "                  "@username"

tokenize_word4("Qu'est-ce que c'est?", split_elisions = FALSE)
#> [[1]]
#> [1] "Qu'est-ce" " "         "que"       " "         "c'est"     "?"
tokenize_word4("Qu'est-ce que c'est?", split_elisions = TRUE)
#> [[1]]
#> [1] "Qu'"    "est-ce" " "      "que"    " "      "c'"     "est"    "?"

```

@odelmarcelle, I could not understand why you added these rules. Can you explain? 

```
### Protect variant selector & whitespace with diacritical marks
$Variant = [\uFE00-\uFE0F];
$Diacritical = [\p{whitespace}][\u0300-\u036F];
# Rules
($ALetterPlus | $Hebrew_Letter) $Variant ($ALetterPlus | $Hebrew_Letter);
($ALetterPlus | $Hebrew_Letter) $Diacritical ($ALetterPlus | $Hebrew_Letter);
```
I am hoping `split_elisions = TRUE` makes it easier to analyze French texts. Don't you think "Qu'est-ce" to shoul be tokenized to "Qu" "'", "est-ce" (separate )?


odelmarcelle:
Nice work, I'm glad you had the chance to look at it!

For your first question, I added this rule to align behaviour with https://github.com/quanteda/quanteda/blob/master/R/tokenizers.R#L45-L48, which already implied that those characters should be protected in the ICU tokenizer. 

I couldn't identify the difference between adding and not adding these rules, so I did not use them. I just gave it a try today, and I think you could drop the regex replacement at https://github.com/quanteda/quanteda/blob/dev-tokenize4/R/tokenizers.R#L242-L244:

- It drops the variation selector for emoji, which is not needed (seems to be already protected through the 'word' ICU rule).
- It does not break whitespace followed by a diacritic mark (which can be equivalently implemented with an ICU rule).

See the following examples
``` r
require(quanteda)
#> Le chargement a nécessité le package : quanteda
#> Package version: 3.2.5
#> Unicode version: 13.0
#> ICU version: 69.1
#> Parallel computing: 12 of 12 threads used.
rules_extended <- c(
  data_breakrules,
  list(
    variant =
      r"---(### Protect variant selector & whitespace with diacritical marks
    $Variant = [\uFE00-\uFE0F];
    $Diacritical = [\p{whitespace}][\u0300-\u036F];
    # Rules
    ($ALetterPlus | $Hebrew_Letter) $Variant ($ALetterPlus | $Hebrew_Letter);
    ($ALetterPlus | $Hebrew_Letter) $Diacritical ($ALetterPlus | $Hebrew_Letter);)---"
  ))


txt <- "i \u2764\ufe0f you \u2764\ufe0f\ufe0f\u2764" 
print(txt)
#> [1] "i ❤️ you ❤️️❤"
tokens(txt)
#> Tokens consisting of 1 document.
#> text1 :
#> [1] "i"   "❤"   "you" "❤"   "❤"
tokenize_word4(txt)
#> [[1]]
#> [1] "i"   " "   "❤"   " "   "you" " "   "❤"   "❤"
tokenize_custom(txt, data_breakrules)
#> [[1]]
#> [1] "i"   " "   "❤️"   " "   "you" " "   "❤️️"   "❤"
tokenize_custom(txt, rules_extended)
#> [[1]]
#> [1] "i"   " "   "❤️"   " "   "you" " "   "❤️️"   "❤"


txt <- "übër u\u0308be\u0308r \u0308ubër"
print(txt)
#> [1] "übër übër ̈ubër"
tokens(txt)
#> Tokens consisting of 1 document.
#> text1 :
#> [1] "übër"     "übërubër"
tokenize_word4(txt)
#> [[1]]
#> [1] "übër"     " "        "übërubër"
tokenize_custom(txt, data_breakrules)
#> [[1]]
#> [1] "übër" " "    "übër" " ̈"    "ubër"
tokenize_custom(txt, rules_extended)
#> [[1]]
#> [1] "übër"      " "         "übër ̈ubër"
```

Might be possible to fine-tune the diacritic ICU rule to retain `\u0308ubër` as a single token, but I am not sure there are a lof of used cases. 


For your second question, "qu" will always be used with an elision to replace "qui" or "que". I'm not sure there is any use in separating the contracted word from the elision.

odelmarcelle:
After giving it a second thought, it is probably better to tokenize "Qu'est-ce" into "Qu" "'", "est-ce" (mainly to avoid having "Qu'" and  "Qu’" as two different tokens). This can be done by adjusting the elision rule to:

```
$Elision = ([lLmMtTnNsSjJdDcC]|([jJ][u][s]|[qQ][u][o][i]|[lL][o][r][s]|[pP][u][i][s]|[qQ][u][e][l])?[qQ][u]);
    $Apostrophe = [\u0027\u2019];
    ^$Elision / $Apostrophe;
```

I also added `|[qQ][u][e][l]` in the elision part to tokenize "Quelqu'un" into "Quelqu" "'" "un"

koheiw:
@odelmarcelle thanks you for the comments. I will investigate the variant selector issue a bit more. 

You are welcome to edit this branch directly (I sent you an invite). In this way, your contribution will be registered properly.

koheiw:
You are the expert on elisions here because other do not know French. In English, extracting "aren" from "aren't" does not make sense, but "he" from "he'll" does.

koheiw:
I removed https://github.com/quanteda/quanteda/blob/dev-tokenize4/R/tokenizers.R#L242-L244. 
``` r
require(stringi)
require(quanteda)
#> Package version: 3.2.5
#> Unicode version: 14.0
#> ICU version: 70.1
#> Parallel computing: 4 of 4 threads used.

txt <- "i \u2764\ufe0f you \u2764\ufe0f\ufe0f\u2764"
tokenize_custom(txt, data_breakrules_word)
#> [[1]]
#> [1] "i"   " "   "❤️"   " "   "you" " "   "❤️️"   "❤"
tokenize_word4(txt)
#> [[1]]
#> [1] "i"   " "   "❤️"   " "   "you" " "   "❤️️"   "❤"
stri_split_boundaries(txt, type = "word")
#> [[1]]
#> [1] "i"   " "   "❤️"   " "   "you" " "   "❤️️"   "❤"

txt2 <- "übër u\u0308be\u0308r \u0308ubër"
tokenize_custom(txt2, data_breakrules_word)
#> [[1]]
#> [1] "übër" " "    "übër" " ̈"    "ubër"
tokenize_word4(txt2)
#> [[1]]
#> [1] "übër" " "    "übër" " ̈"    "ubër"
stri_split_boundaries(txt2, type = "word")
#> [[1]]
#> [1] "übër" " "    "übër" " ̈"    "ubër"
```

The better way to deal with the stray diacritical marks is post-tokenization cleaning. We can add `tokens(remove_control = TRUE)` for this. 
```r
tokens(txt2, what = "word4") %>% 
    tokens_remove("^[\\p{Z}\\p{M}\\p{C}]+$", valuetype = "regex")
#> Tokens consisting of 1 document.
#> text1 :
#> [1] "übër" "übër" "ubër"
```

koheiw:
The only problem of the new tokenizer seems to be the handling of @, for which the [rules have been changed in the ICU library](https://github.com/unicode-org/icu/commit/49d192fefe09fcc38547203487cf4e63d2dad61f).
```
── Failure ('test-tokens-word4.R:948:5'): split_tags works ─────────────────────
as.list(tokens(txt1, what = "word", split_tags = TRUE)) not identical to list(d1 = c("@", "quanteda", "@", "koheiw7", "@", "QUANTEDA_INITIATIVE")).
Component “d1”: Lengths (3, 6) differ (string compare on first 3)
Component “d1”: 3 string mismatches
```

koheiw:
> The better way to deal with the stray diacritical marks is post-tokenization cleaning. We can add `tokens(remove_control = TRUE)` for this.

I solved this in https://github.com/quanteda/quanteda/pull/2216/commits/7f90f4e439c2b598e94c6cad432b924b73e12d83.

koheiw:
> After giving it a second thought, it is probably better to tokenize "Qu'est-ce" into "Qu" "'", "est-ce" (mainly to avoid having "Qu'" and "Qu’" as two different tokens). This can be done by adjusting the elision rule to:
> 
> ```
> $Elision = ([lLmMtTnNsSjJdDcC]|([jJ][u][s]|[qQ][u][o][i]|[lL][o][r][s]|[pP][u][i][s]|[qQ][u][e][l])?[qQ][u]);
> $Apostrophe = [\u0027\u2019];
> ^$Elision / $Apostrophe;
> ```
> 
> I also added `|[qQ][u][e][l]` in the elision part to tokenize "Quelqu'un" into "Quelqu" "'" "un"

I named the argument `split_elisions`, but isn't it better to call it `split_apostrophe` (or `split_apos`) to make it less langauge dependent?

```
[a-zA-Z] / [\u0027\u2019];
```

kbenoit:
OK will do. textmodels should be back on CRAN Wednesday. I fixed the compiler warnings but they applied a bunch of very stringent documentation requirements to the resubmission that were not flagged before it was removed!

odelmarcelle:
> I named the argument `split_elisions`, but isn't it better to call it `split_apostrophe` (or `split_apos`) to make it less langauge dependent? @odelmarcelle ?
> 
> ```
> [a-zA-Z] / [\u0027\u2019];
> ```

I thought about it again, and maybe the best is to align the behavior of the tokenizer to what part-of-speech taggers do. I quickly compared with what `spacyr` and `udpipe` offer out of the box, and they both retain the apostrophe attached to the contracted token. They also indicate the relevant lemma, to which the apostrophe belongs.

Furthermore, breaking the elision and the apostrophe tends to confuse POS models used afterward. See the example below where `udpipe`uses the result of `tokens()`.

For that reason, I'll revert the change to the elision rule to retain the contracted word and the elision together.

I also believe that other languages deserve a different rule. For example, "C'est" is the contraction of "Ce", "est". The pronoun is being contracted. In contrast, "He'll" is the contraction of "He", "will". Here, the verb is being contracted. "don't" can only be meaningfully decomposed into "do" "not". This all becomes quite language-specific, and I don't believe a single rule will fit all use cases.


See the below examples for my comparison with `spacyr` and `udpipe`

``` r
require(quanteda)
#> Le chargement a nécessité le package : quanteda
#> Package version: 3.2.5
#> Unicode version: 13.0
#> ICU version: 69.1
#> Parallel computing: 12 of 12 threads used.
require(spacyr)
#> Le chargement a nécessité le package : spacyr
#> Warning: le package 'spacyr' a été compilé avec la version R 4.2.2
# spacy_download_langmodel("fr")
spacy_initialize(model="fr_core_news_sm")
#> Found 'spacy_condaenv'. spacyr will use this environment
#> successfully initialized (spaCy Version: 3.1.3, language model: fr_core_news_sm)
#> (python options: type = "condaenv", value = "spacy_condaenv")
require(udpipe)
#> Le chargement a nécessité le package : udpipe
#> Warning: le package 'udpipe' a été compilé avec la version R 4.2.2
# udmodel <- udpipe_download_model(language = "french")
udmodel <- udpipe_load_model(file = "french-gsd-ud-2.5-191206.udpipe")

txt <- "Qu'est-ce que c'est?. Ce n'est pas normal. J'aime beaucoup voyager."

# SpaCy
spacy_parse(txt)
#> Warning in spacy_parse.character(txt): lemmatization may not work properly in
#> model 'fr_core_news_sm'
#>    doc_id sentence_id token_id    token    lemma   pos entity
#> 1   text1           1        1      Qu'      qu'  PRON       
#> 2   text1           1        2      est     être   AUX       
#> 3   text1           1        3      -ce      -ce  PRON       
#> 4   text1           1        4      que      que SCONJ       
#> 5   text1           1        5       c'       ce  PRON       
#> 6   text1           1        6      est     être  VERB       
#> 7   text1           1        7        ?        ? PUNCT       
#> 8   text1           1        8        .        . PUNCT       
#> 9   text1           2        1       Ce       ce  PRON       
#> 10  text1           2        2       n'       n'   ADV       
#> 11  text1           2        3      est     être   AUX       
#> 12  text1           2        4      pas      pas   ADV       
#> 13  text1           2        5   normal   normal   ADJ       
#> 14  text1           2        6        .        . PUNCT       
#> 15  text1           3        1       J'       je  PRON       
#> 16  text1           3        2     aime     aime  VERB       
#> 17  text1           3        3 beaucoup beaucoup   ADV       
#> 18  text1           3        4  voyager  voyager  VERB       
#> 19  text1           3        5        .        . PUNCT

# UDpipe
x <- udpipe(udmodel, x = txt)
x <- as.data.frame(x)
x[, c(1, 3, 8:11)]
#>    doc_id sentence_id token_id    token    lemma  upos
#> 1    doc1           1        1      Qu'      que  PRON
#> 2    doc1           1        2      est     être   AUX
#> 3    doc1           1        3      -ce       ce  PRON
#> 4    doc1           1        4      que      que SCONJ
#> 5    doc1           1        5       c'       ce  PRON
#> 6    doc1           1        6     est?     est? PUNCT
#> 7    doc1           1        7        .        . PUNCT
#> 8    doc1           2        1       Ce       ce  PRON
#> 9    doc1           2        2       n'       ne   ADV
#> 10   doc1           2        3      est     être   AUX
#> 11   doc1           2        4      pas      pas   ADV
#> 12   doc1           2        5   normal   normal   ADJ
#> 13   doc1           2        6        .        . PUNCT
#> 14   doc1           3        1       J'       il  PRON
#> 15   doc1           3        2     aime    aimer  VERB
#> 16   doc1           3        3 beaucoup beaucoup   ADV
#> 17   doc1           3        4  voyager  voyager  VERB
#> 18   doc1           3        5        .        . PUNCT

# UDpipe based on tokens()
x <- udpipe(udmodel, x = as.list(
              tokens(txt, split_elisions = TRUE, split_hyphens = TRUE, what="word4")
            ))
x <- as.data.frame(x)
x[, c(1, 3, 8:11)]
#>    doc_id sentence_id token_id    token    lemma  upos
#> 1   text1           1        1       Qu      que SCONJ
#> 2   text1           1        2        '        ' PUNCT
#> 3   text1           1        3      est     être   AUX
#> 4   text1           1        4        -        - PUNCT
#> 5   text1           1        5       ce       ce  PRON
#> 6   text1           1        6      que      que  PRON
#> 7   text1           1        7        c       ce  PRON
#> 8   text1           1        8        '        ' PUNCT
#> 9   text1           1        9      est     être   AUX
#> 10  text1           1       10        ?        ? PUNCT
#> 11  text1           1       11        .        . PUNCT
#> 12  text1           1       12       Ce       ce   DET
#> 13  text1           1       13        n        n  NOUN
#> 14  text1           1       14        '        ' PUNCT
#> 15  text1           1       15      est     être   AUX
#> 16  text1           1       16      pas      pas   ADV
#> 17  text1           1       17   normal   normal   ADJ
#> 18  text1           1       18        .        . PUNCT
#> 19  text1           1       19        J        J     X
#> 20  text1           1       20        '        ' PUNCT
#> 21  text1           1       21     aime    aimer  VERB
#> 22  text1           1       22 beaucoup beaucoup   ADV
#> 23  text1           1       23  voyager  voyager  VERB
#> 24  text1           1       24        .        . PUNCT
```

rules/cusotm.yml
```diff
@@ -0,0 +1,26 @@
...
+elision: |
+    $Elision = ([lLmMtTnNsSjJdDcC]|([jJ][u][s]|[qQ][u][o][i]|[lL][o][r][s]|[pP][u][i][s]|[qQ][u][e][l])?[qQ][u])[\u0027\u2019];
+    ^$Elision / $ALetterPlus;
+
+
+# Split tags
+tag: |
+    [#] / [\\p{L}\\p{N}]+; 
+    [\\p{L}\\p{N}]+ / [#]; 
+    [@] / [a-zA-Z0-9_]+;
```
odelmarcelle, line 26: I would advise using a leading `^` to match only the beginning of a word. Otherwise, username such as koheiw@quanteda would be affected as well.

koheiw: Update!

R/tokenizers.R
```diff
@@ -228,3 +229,71 @@ normalize_characters <- function(x) {
...
+}
+
+#' Customizable tokenizer
+#'
+#' Allows users to tokenize texts using customized boundary rules. See the [ICU
+#' website](https://unicode-org.github.io/icu/userguide/boundaryanalysis/break-rules.html)
+#' for how to define boundary rules.
+#' @param x character vector for texts to tokenize
+#' @param rules a list of rules for rule-based boundary detection
+#' @importFrom stringi stri_split_boundaries
```
kbenoit, line 266: I'm afraid we need to get into the habit now of including @return values on all (new) documented functions. (See https://r-pkgs.org/man.html#sec-man-returns) I just spent a full day fixing this on **quanteda.textmodels**.

R/quanteda_options.R
```diff
@@ -167,9 +167,10 @@ get_options_default <- function() {
                  base_featname = "feat",
                  base_compname = "comp",
                  language_stemmer = "english",
-                 pattern_hashtag = "#\\w+#?",
+                 pattern_hashtag = "#[\\p{L}\\p{N}]+#?",
```
kbenoit, line 170: We actually had this wrong before, since according to Twitter, a purely numeric (included in \w or \p{Nd}) sequence following the `#` is not a valid hashtag. So, 
#420day OK
#420 not ok

kbenoit: Ha, the irony, GitHub converted the second into a reference to an ancient issue about regexes.

R/tokenizers.R
```diff
@@ -36,7 +37,7 @@ NULL
 #' @rdname tokenize_internal
 #' @importFrom stringi stri_replace_all_regex stri_detect_fixed stri_split_boundaries
 #' @export
-tokenize_word <- function(x, split_hyphens = FALSE, verbose = quanteda_options("verbose")) {
+tokenize_word <- function(x, split_hyphens = FALSE, verbose = quanteda_options("verbose"), ...) {
```
kbenoit, line 40: What about this: We rename the v3 "word" to "word3" and just make the default through the new options the current version. For compatibility, we could change the tokenizer to a major version number, and not change it during the major version. This could include the sentence rules too.

koheiw: Yes, that the plan. Please do if you want to do now.

R/tokens.R
```diff
@@ -505,7 +506,7 @@ removals_regex <- function(separators = FALSE,
                            url = FALSE) {
     regex <- list()
     if (separators)
-        regex[["separators"]] <- "^[\\p{Z}\\p{C}]+$"
+        regex[["separators"]] <- "^[\\p{Z}\\p{C}\\p{M}]+$"
```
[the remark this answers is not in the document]
koheiw: I added M because I remove https://github.com/quanteda/quanteda/blob/dev-tokenize4/R/tokenizers.R#L242-L244 from word4.

R/tokenizers.R
koheiw: That is the right thing to do.

R/tokenizers.R
kbenoit: OK, I'll work on a way to include the "future" v4 tokenizer in a way that can be specified in the options (I love this approach btw!) but still leave the default at v3 so that we can merge and publish it as a v3 release.

R/quanteda_options.R
koheiw: Actually we no longer need the patterns, because we have the same in our custom rules.
https://github.com/quanteda/quanteda/blob/33813d2e80c932bad2f690a7bba9cdaa42f369ae/rules/cusotm.yml#L18-L19
However, it is still nice to provide easy way to customize rules. We could make them `pattern_hashtag` and `pattern_hashtag` user name by default if you want. We could even rename it `tokens_tokenizer_rules` or something.

R/tokenizers.R
koheiw: Done

R/tokens.R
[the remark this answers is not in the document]
koheiw: I mean this https://github.com/quanteda/quanteda/blob/9d59bb201e6b62c6e7e4408d72f90e8054b19469/R/tokenizers.R#L46-L49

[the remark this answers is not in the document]
koheiw: The third token `" ̈"` is diacritical mark (wrongly) modifying a whitespace. I need M to clean it up.
```
txt2 <- "übër u\u0308be\u0308r \u0308ubër"
tokenize_word4(txt2)
#> [[1]]
#> [1] "übër" " "    "übër" " ̈"    "ubër"
```

koheiw:
@odelmarcelle I want to include you in the list of contributors. If you like the idea, please share your name and email with me.

https://github.com/quanteda/quanteda/blob/9d59bb201e6b62c6e7e4408d72f90e8054b19469/DESCRIPTION#L11-L21

kbenoit:
> You are the expert on elisions here because other do not know French. In English, extracting "aren" from "aren't" does not make sense, but "he" from "he'll" does.

Mon français est rouillé mais en ce qui concerne les règles grammaticales, je m'en souviens encore le plus!

I believe the real problem is that there is no single set of rules that covers elisions and contractions (which as I understand are a subset of elisions) in all languages.

The new word4 tokenizer gets the English wrong, but no more wrong than word3. It's better for French when `split_elision = TRUE`. spaCy gets the French tokens wrong when using the English language model.
``` r
library("quanteda")
#> Package version: 3.2.5
#> Unicode version: 14.0
#> ICU version: 71.1

txt <- c(txt_en = "He'll say we aren't.",
         txt_fr = "Est-ce qu'on peut succir, lorsqu’on essai?")
tokenize_word4(txt, split_elisions = TRUE)
#> $txt_en
#> [1] "He'll"  " "      "say"    " "      "we"     " "      "aren't" "."     
#> 
#> $txt_fr
#>  [1] "Est-ce"  " "       "qu'"     "on"      " "       "peut"    " "      
#>  [8] "succir"  ","       " "       "lorsqu’" "on"      " "       "essai"  
#> [15] "?"

library("spacyr")
spacy_initialize("en_core_web_sm")
#> Found 'spacy_condaenv'. spacyr will use this environment
#> successfully initialized (spaCy Version: 3.4.4, language model: en_core_web_sm)
#> (python options: type = "condaenv", value = "spacy_condaenv")
spacy_tokenize(txt)
#> $txt_en
#> [1] "He"  "'ll" "say" "we"  "are" "n't" "."  
#> 
#> $txt_fr
#>  [1] "Est"       "-"         "ce"        "qu'on"     "peut"      "succir"   
#>  [7] ","         "lorsqu’on" "essai"     "?"
```

But using the French language model, the English is now wrong but the French version is right.
``` r
txt <- c(txt_en = "He'll say we aren't.",
         txt_fr = "Est-ce qu'on peut succir, lorsqu’on essai?")
library("spacyr")
spacy_initialize("fr_core_news_sm")
#> Found 'spacy_condaenv'. spacyr will use this environment
#> successfully initialized (spaCy Version: 3.4.4, language model: fr_core_news_sm)
#> (python options: type = "condaenv", value = "spacy_condaenv")
spacy_tokenize(txt)
#> $txt_en
#> [1] "He'"   "ll"    "say"   "we"    "aren'" "t."   
#> 
#> $txt_fr
#>  [1] "Est"     "-ce"     "qu'"     "on"      "peut"    "succir"  ","      
#>  [8] "lorsqu’" "on"      "essai"   "?"
```

But anyway, since `split_elision = FALSE` as a default, we are not doing any unintended harm, and it's an improvement for users who know what they are doing. 

But the above illustrates why I would suggest users coerce the results of `spacy_tokenize()` into a tokens object if they really want to get the language rules right.

kbenoit:
@odelmarcelle I just realised you had done a similar analysis as to what I just posted yesterday using spacyr and udpipe. Same conclusions. I think the `split_elision` is ok since it's only passed via `...` in `tokens()`, and the idea is to allow user choice for the tokenizer. I can live with that nomenclature.

odelmarcelle:
@koheiw Thank you for the offer, I added myself to the list of contributors :).

Something else that has been puzzling me: the goal is to skip `preserve_special` in the new v4 tokenizer? In that case, modifying the option `tokens_tokenizer_word` is not sufficient as it still branch to `preserve_special`. I corrected this in a119f5e3d898188f10339f4bba2cb8125d463de2 and it leads to failures in the test suite. I will investigate it a bit later.

koheiw:
You are right. `preserve_special` and `restore_special` should be used only when it is `tokenize_word3` (the current `tokenize_word`). We should tidy up the names.

odelmarcelle:
I tried to understand why the test failed at https://github.com/quanteda/quanteda/blob/48e8ca9e2b6340573bbd3726b81dae815aab98ca/tests/testthat/test-tokens.R#L1038-L1044 when skipping `preserve_special`.

I'm afraid it is caused by the dictionary rule in the base word rules: https://github.com/unicode-org/icu/blob/68f5ef835bd69221bf1c23ba65228d0cae7e5647/icu4c/source/data/brkitr/rules/word.txt#L66

If I understand correctly, most chinese-japanese-korean words are actually broken down into smaller pieces by matching sequences to a large pre-defined dictionary: https://github.com/unicode-org/icu/blob/main/icu4c/source/data/brkitr/dictionaries/cjdict.txt

I'm not sure how to override this behavior through additional RBBI rules.

koheiw:
Sorry that I did not notice this earlier. It seems that it is not easy to turn off the dictionary-based break iterator in the middle of a text. In it not possible, we use `preserve_special` only for tags. We at least moved protection of URLs, emails and hyphens to RBBI. I will take care of this issue.

rules/custom.yml
```diff
@@ -0,0 +1,29 @@
+# https://unicode-org.github.io/icu/userguide/boundaryanalysis/break-rules.html
+
+# Keep hyphens
+keep_hyphens: |
+    $Hyphen = [\p{Pd}];
+    ($ALetterPlus | $Hebrew_Letter | $Numeric)  $ExFm* $Hyphen $ExFm* ($ALetterPlus | $Hebrew_Letter | $Numeric);
+
+# Keep URLs
+keep_url: |
+    ^(([h][t][t][p][s]?\:\/\/([w][w][w]\.)?)|([w][w][w]\.))[-a-zA-Z0-9@:%_\+\.~#=]+(\/[-a-zA-Z0-9@\:%_\+\.~#?\&=]+)*[\/]?;
```
koheiw, line 10: @odelmarcelle  I think this should be literal 
```
"('http' | 'https') '://' [-a-zA-Z0-9@:%_+.~#=]+;"
```

koheiw:
This is how it looks like if we use `tokens_compound()` for tags.
``` r
require(quanteda)
#> Package version: 3.2.5
#> Unicode version: 14.0
#> ICU version: 70.1
#> Parallel computing: 6 of 6 threads used.
txt <- c(d1 = "オリンピック延期決定！ #politics @abe #政治# #政治 #安部政権 @安部政権 ！")
tokens(txt, "word4", split_tags = FALSE) 
#> Tokens consisting of 1 document.
#> d1 :
#>  [1] "オリンピック" "延期"         "決定"         "！"           "#politics"   
#>  [6] "@abe"         "#政治#"       "#政治"        "#安部政権"    "@安部政権"   
#> [11] "！"
tokens(txt, "word4", split_tags = TRUE) 
#> Tokens consisting of 1 document.
#> d1 :
#>  [1] "オリンピック" "延期"         "決定"         "！"           "#"           
#>  [6] "politics"     "@"            "abe"          "#"            "政治"        
#> [11] "#"            "#"           
#> [ ... and 8 more ]
```

The only problem is that separators after "#" or "@" are also compounded. I might need to upgrade `tokens_compound()`  or create a dedicated function for this operation.
```r
tokens(txt, "word4", remove_separators = FALSE, split_tags = FALSE)
#> Tokens consisting of 1 document.
#> d1 :
#> [1] "オリンピック"                                      
#> [2] "延期"                                              
#> [3] "決定"                                              
#> [4] "！"                                                
#> [5] " "                                                 
#> [6] "#politics @abe #政治# #政治 #安部政権 @安部政権 ！"
```

rules/custom.yml
odelmarcelle: You're right, I wasn't aware we could specify sequences of literals like this. I've rewritten and enhanced the rule.

koheiw:
I created `tokens_restore()` to solve this problem. @kbenoit or @odelmarcelle , please review https://github.com/quanteda/quanteda/pull/2223.
``` r
require(quanteda)
require(stringi)
txt <- "オリンピック延期決定！ #politics @abe #政治# #政治 #安部政権 @安部政権 ！"
txt <- stri_replace_all_regex(txt, "@[a-zA-Z0-9_]+|#[\\p{L}\\p{N}]+#?", "\uE001$0\uE002")
as.tokens(stri_split_boundaries(txt, type = "word")) %>% 
    quanteda:::tokens_restore() %>% 
    print(-1, -1)
#> Tokens consisting of 1 document.
#> text1 :
#>  [1] "オリンピック" "延期"         "決定"         "！"           " "           
#>  [6] "#politics"    " "            "@abe"         " "            "#政治#"      
#> [11] " "            "#政治"        " "            "#安部政権"    " "           
#> [16] "@"            "安部"         "政権"         " "            "！"
```

<sup>Created on 2023-03-26 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

kbenoit:
#2223 looked fine to me, so merging was fine. Sorry I was not able to keep up with the pace of PRs this weekend.

On this PR, I am ready to address a few changes myself, namely the change to the data object versus the environment I put into the code comments, so how about you signal with this PR is stable enough for me to implement a few of those items? Happy to start on this tomorrow (mar 28).

koheiw:
This branch is apparently not ready, but  please join our discussion.

kbenoit (changes requested):
This looks very good in that it improves the default **quanteda** tokenizer in tokens.  I think that with this update, we can now modularise the tokeniser according to which major version is being used.  This means we can merge the branch into the existing master, rather than thinking this has to be in a v4 release.  Because the current default can be set at "word3", which is equivalent to the version 3 "word", the functionality will be the same during the lifecycle of v3.  When v4 is released, we can simply change the default from "word3" to to "word4".

This update - and this summary is for our understanding as well as the release notes - does the following:

Major change:
- Implements a "word4" tokeniser that is based on new RBBI rules, implemented in a new .yml file that can be edited and changed by users, but whose defaults represent a significant improvement in pattern handling for words, sentences, and other forms of patterns.


Minor changes:
- changes how elapsed time is recorded, by creating a global environment to record these in (`aaa.R`).
- improves several of the R-coded patterns that apply to "word3":
  - the hashtag pattern (`pattern_hashtag)
  - the separator pattern (by adding `\\p{M}]`)
  - the URL pattern
- creates a new `tokens_restore()`, implemented in C++, to replace the older `preserve_special()` that rejoined splits created by the default **stringi** tokeniser machinery.
- makes some technical improvements to internal tokenisation functions, such as moving the ellipsis to the end of the function, to allow more modularity in developing future tokenisers.


Some things I still would like to address or see explained:
- [ ] Can we stop illegal hashtags such as #420 ? Or are we not sticking to "Twitter" rules?
- [ ] The rules should not be in `/data/data_*.rda` files, but rather implemented as per my comments. I'm happy to implement this change, but I expected some discussion in the review comments.
- [ ] Some breaking tests need scrutiny, I will review these when I run through the tests later today.
- [ ] Needs updates to `NOTES.md`, which could be my summary above.
- [ ] Is there any reason not to merge this not into branch `v4` but rather into `master`? Why not make it part of the current v3 release, set by default to the word3 tokeniser?

R/tokenizers.R
```diff
@@ -228,3 +229,71 @@ normalize_characters <- function(x) {
 
     return(x)
 }
+
+#' @rdname tokenize_internal
+#' @export
+tokenize_word4 <- function(x, split_hyphens = FALSE, split_tags = FALSE, split_elisions = FALSE,
+                           verbose = quanteda_options("verbose"), ...) {
+    
+    rules <- quanteda::data_breakrules_word
```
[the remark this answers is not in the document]
odelmarcelle: I think there should still be options to enable/disable rules when calling the function, without modying a global object. For example, when analyzing french and english texts in a same R session, I'd like to enable the elision split only for french.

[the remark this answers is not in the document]
koheiw: I created the data object only to avoid reading custom.yml from the disk every time we run the tokenizer. It should be more easily customizable by the users admittedly.

Please note that the role of the custom RBBI shrunk in my latest commits (URLs and emails are also restored using `tokens_restore()` to leave @ and # in them intact). RBBI is used mainly for hyphens and elisions now.

R/quanteda_options.R
```diff
@@ -167,9 +167,10 @@ get_options_default <- function() {
...
                  pattern_username = "@[a-zA-Z0-9_]+",
                  tokens_block_size = 10000L,
-                 tokens_locale = "en_US@ss=standard")
+                 tokens_locale = "en_US@ss=standard",
+                 tokens_tokenizer_word = "tokenize_word3")
```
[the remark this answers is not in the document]
koheiw: I ended up using these settings in the same way as the current master. I am preparing `tokens.corpus()` for the use of `cpp_serialize()`, but we can backport the work4 tokenizer to v3x if you want to. It still need a lot of small adjustment/correction.

kbenoit:
OK, now that the tests are all passing, here's what I will work on next.

- [ ] add full modularity to the names, so that the default word tokenizer is the current major version number, and so that we will be ready to switch this over to word4 when that is released, but still can use it by changing the option or overriding it in using the `what` argument in `tokens()`
- [ ] change it so that the `quanteda_options("tokens_tokenizer_word")` takes labels instead of function names
- [ ] move the .rda rules to an environment loaded from the config files, as per https://r-pkgs.org/data.html#sec-data-state
- [ ] add some tests that will break, until we fix them again or remove them, to `test-tokenizers.R`, including
    - [ ] word4 not splitting hashtags that contain underscores, like #hash_tag
    - [ ] word3 and word4 not breaking #554 (the currently disabled tests for the special hiragana characters, namely "゛んﾞ")
    - [ ] not letting purely numeric values be valid hashtags, such as "# 420" (without the space)
    - [ ] seeing if we cannot condition the ICU etc versions for tests like the second one in `test-tokenizers.R`, which is currently disabled because it is platform-dependent

R/tokens.R
```diff
@@ -247,7 +242,8 @@ tokens.corpus <- function(x,
                           verbose = quanteda_options("verbose"),
                           ...)  {
     x <- as.corpus(x)
-    what <- match.arg(what, c("word", "word1", "sentence", "character",
+    what <- match.arg(what, c("word", "sentence", "character",
+                              "word1", "word3", "word4",
```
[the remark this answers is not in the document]
kbenoit: I implemented this in [5cbcee4](https://github.com/quanteda/quanteda/pull/2216/commits/5cbcee4daae358c360cc0711e75b2e21ce54b6e8)

rules/cusotm.yml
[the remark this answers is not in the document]
kbenoit: Done in [5753067](https://github.com/quanteda/quanteda/pull/2216/commits/5753067347fea723869b2ba2d86467a701316a9a)

kbenoit (approved):
There are still some things I'd like to see updated, but I will approve this now.

- the tests in `test-tokenizer()`

- these tests in `test-tokens.R` and `test-tokens-word4.R`, which addressed #554

```r
txts <- c("づいﾞ", "゛んﾞ", "たーﾟ")
expect_equivalent(
    as.list(tokens(txts)),
    list(c("づ", "いﾞ"), c("゛", "んﾞ"), c("た", "ーﾟ"))
)
```

- this test for valid hashtag

```r
txt <- c("Got high on #420 day")
expect_identical(
    as.list(tokens(txt, what = "word4")),
    list(text1 = c("Got", "high", "on", "#", "420", "day"))
)
```

- this test for not splitting underscores in a hashtag

```r
txt <- c("I love #tokens_compound")
expect_identical(
    as.list(tokens(txt, what = "word4")),
    as.list(tokens(txt, what = "word3")),
    list(text1 = c("I", "love", "#tokens_compound"))
)
```

tests/testthat/test-tokens-word4.R
```diff
@@ -936,7 +894,7 @@ test_that("split_tags works", {
     txt2 <- c(d1 = "#quanteda #q-x #q_y #q100 #q")
     expect_identical(
         as.list(tokens(txt2, what = "word")),
-        list(d1 = c("#quanteda", "#q-x", "#q_y", "#q100", "#q"))
+        list(d1 = c("#quanteda", "#q", "-", "x", "#q_y", "#q100", "#q"))
```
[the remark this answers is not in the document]
kbenoit: This really needs to be resolved, it's a flaw in "word4".

R/tokens.R
```diff
@@ -259,64 +255,81 @@ tokens.corpus <- function(x,
...
-                           sentence = tokenize_sentence,
-                           character = tokenize_character,
-                           fasterword = tokenize_fasterword,
-                           fastestword = tokenize_fastestword)
-
-
-    if (!remove_separators && !what %in% c("word", "word1", "character"))
+    
+    tokenizer <- switch(what,
+                        word = quanteda_options("tokens_tokenizer_word"),
```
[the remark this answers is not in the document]
kbenoit: Fixed now in [5cbcee4](https://github.com/quanteda/quanteda/pull/2216/commits/5cbcee4daae358c360cc0711e75b2e21ce54b6e8)

kbenoit:
@koheiw if you are ok with how I implemented the modularity of word4 / word3 etc, and the custom rules as part of a global environment created at startup, then please merge both to master and the base branch for this PR.

tests/testthat/test-tokens-word4.R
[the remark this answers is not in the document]
koheiw: They are different because of 

```
> stringi::stri_match_all_regex("#hash_tag", "#[\\p{L}\\p{N}]+#?")
[[1]]
     [,1]   
[1,] "#hash"

> stringi::stri_match_all_regex("#hash_tag", "#\\w+#?")
[[1]]
     [,1]       
[1,] "#hash_tag"
```
We can revert the default value for `pattern_hashtag` to the latter.

koheiw:
I like putting RBBI rules in the environment. Thanks. 

`breakrules_get()` sounds odds to me. How about `breakrules()` (like `stopwords()`) for getting and `breakrules<-` for setting values to customize?

kbenoit:
> I like putting RBBI rules in the environment. Thanks.
> 
> `breakrules_get()` sounds odd to me. How about `breakrules()` (like `stopwords()`) for getting and `breakrules<-` for setting values to customize?

Even better. I think we should keep these internal for v3 and see how they develop. Only the most advanced users are likely to make use of them. I was thinking more that this, plus the update script, provides a way for us to update the default rules easily.

kbenoit (approved):

kbenoit:
> > I like putting RBBI rules in the environment. Thanks.
> > `breakrules_get()` sounds odd to me. How about `breakrules()` (like `stopwords()`) for getting and `breakrules<-` for setting values to customize?
> 
> Even better. I think we should keep these internal for v3 and see how they develop. Only the most advanced users are likely to make use of them. I was thinking more that this, plus the update script, provides a way for us to update the default rules easily.

Actually I started implementing this now, then realised we cannot use the same function for assignment and retrieval, because `x` cannot be passed as the thing to be modified. So better to have `breakrules_get()`, breakrules_set()`. (Later we can add validators, no need for now however - this is still mainly internal.)
See https://r-pkgs.org/data.html#sec-data-state

koheiw:
Can't we modify the package-level environment at all?

kbenoit:
> Can't we modify the package-level environment at all?

Only inside functions - so it cannot be passed as the `x` in an assignment function.

I just fixed these but now see you updated it in [6525670](https://github.com/quanteda/quanteda/pull/2216/commits/6525670d611109f65ace66a5a98258e6d3a2e3be)

koheiw:
Let's discuss how to customize rules on a branch for v4. I still don't like `_set()` and `_get()` functions.

Otherwise, word4 is ready. I fixed the issues in segmentation of tags and Japanese.

kbenoit:
Take a look at how I implemented them in 21248f3, keeping in mind that these are internal for v3 and could be improved for v4, with validators for instance. But we are limited in how we define these, given the use of environments.

kbenoit:
> Take a look at how I implemented them in [21248f3](https://github.com/quanteda/quanteda/commit/21248f36038adf82d737cea14b99cd3867dde25a), keeping in mind that these are internal for v3 and could be improved for v4, with validators for instance. But we are limited in how we define these, given the use of environments.

You might need to fix this slightly because I pushed it right before running off to a seminar

koheiw:
The rules `stri_split_boundary(type = "word")` is a bit different from `breakrules()` because it is download from the ICU's repository directly. How is the result if you `stri_split_boundary("゛んﾞ", type = breakrules()$base)`.

koheiw:
We don't need to finish `breakrules()` here, but I think it should behave like `quanteda_options()`.
```r
# get
> quanteda_options("tokens_tokenizer_word")
[1] "word3"

# set 
> quanteda_options("tokens_tokenizer_word" = "word4")
```

```r
# get
> breakrules("base")
[1] "#\n# Copyright (C) 2016 and later: Unicode, Inc. and others.\n# License & terms of use: http://www.unicode.org/copyright.html\n# Copyright (C) 2002-2016, International Business Machines Corporation\n# and others. All Rights Reserved.\n#\n# file:  word.txt\n#\n# ICU Word Break Rules\n#

# set
breakrules("custom" = "xyz") 
```

kbenoit:
``` r
library("quanteda")
#> Package version: 3.2.5.9000
#> Unicode version: 14.0
#> ICU version: 71.1

stringi::stri_split_boundaries("゛んﾞ", type = breakrules_get()$base)
#> [[1]]
#> [1] "゛んﾞ"

packageVersion("stringi")
#> [1] '1.7.12'
```

<sup>Created on 2023-03-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

kbenoit:
> We don't need to finish `breakrules()` here, but I think it should behave like `quanteda_options()`.
> 
> ```r
> # get
> > quanteda_options("tokens_tokenizer_word")
> [1] "word3"
> 
> # set 
> > quanteda_options("tokens_tokenizer_word" = "word4")
> ```
> 
>  ```r
> # get
> > breakrules("base")
> [1] "#\n# Copyright (C) 2016 and later: Unicode, Inc. and others.\n# License & terms of use: http://www.unicode.org/copyright.html\n# Copyright (C) 2002-2016, International Business Machines Corporation\n# and others. All Rights Reserved.\n#\n# file:  word.txt\n#\n# ICU Word Break Rules\n#
> 
> # set
> breakrules("custom" = "xyz") 
> ```

I'd be happy with that.

They could even _be_ quanteda options!

koheiw:
I have no idea why it works differently on your Mac. It is a problem in stringi instead of our tokenizers. I am not worried about this unusual Japanese strings, so we can merge this PR.
```

> stringi::stri_split_boundaries("゛んﾞ", type = breakrules()$base)
[[1]]
[1] "゛"  "んﾞ"

> stringi::stri_split_boundaries("゛んﾞ", type = "word")
[[1]]
[1] "゛"  "んﾞ"

> packageVersion("stringi")
[1] ‘1.7.12’
```

kbenoit:
I was just about to post this as a **stringi** issue, and noticed that it worked as expected. Now this is really weird:

``` r
# just to show Unicode info
library("quanteda")
#> Package version: 3.2.5.9000
#> Unicode version: 14.0
#> ICU version: 71.1

packageVersion("stringi")
#> [1] '1.7.12'
stringi::stri_split_boundaries("゛んﾞ")
#> [[1]]
#> [1] "゛"  "んﾞ"

sessionInfo()
#> R version 4.2.3 (2023-03-15)
#> Platform: aarch64-apple-darwin20 (64-bit)
#> Running under: macOS Ventura 13.2.1
#> 
#> Matrix products: default
#> BLAS:   /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/lib/libRblas.0.dylib
#> LAPACK: /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/lib/libRlapack.dylib
#> 
#> locale:
#> [1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> other attached packages:
#> [1] quanteda_3.2.5.9000
#> 
#> loaded via a namespace (and not attached):
#>  [1] Rcpp_1.0.10        rstudioapi_0.14    knitr_1.42         magrittr_2.0.3    
#>  [5] stopwords_2.3      R.cache_0.16.0     lattice_0.20-45    rlang_1.1.0       
#>  [9] fastmatch_1.1-3    fastmap_1.1.1      styler_1.9.1       tools_4.2.3       
#> [13] grid_4.2.3         xfun_0.37          R.oo_1.25.0        cli_3.6.1         
#> [17] withr_2.5.0        htmltools_0.5.5    RcppParallel_5.1.7 yaml_2.3.7        
#> [21] digest_0.6.31      lifecycle_1.0.3    Matrix_1.5-3       purrr_1.0.1       
#> [25] vctrs_0.6.1        R.utils_2.12.2     fs_1.6.1           glue_1.6.2        
#> [29] evaluate_0.20      rmarkdown_2.21     reprex_2.0.2       stringi_1.7.12    
#> [33] compiler_4.2.3     R.methodsS3_1.8.2
```

<sup>Created on 2023-03-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

But then 

``` r
library("quanteda")
#> Package version: 3.2.5.9000
#> Unicode version: 14.0
#> ICU version: 71.1

tokens("゛んﾞ")
#> Tokens consisting of 1 document.
#> text1 :
#> [1] "゛んﾞ"
```

<sup>Created on 2023-03-30 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Which suggests it's **quanteda**, maybe something might be happening post-split in handling the re-joining for special split characters?

PR #2221, opened 2023-03-22
Check if the value is NULL explicitly

koheiw:


kbenoit (approved):
But once again please merge to `master`.

PR #2222, opened 2023-03-22
Allow extracting system fields using docvars()

koheiw:


kbenoit (changes requested):
Looks fine, but
- Let's not work on parallel branches unless they involve major breaking changes, which this does not - so let's make the base branch for this merge `master` instead of `v4`. (Then you can update `v4` from `master`.)
- Suggest you merge #2216 into `master` first.
- That means you should update `NEWS.md` with this change too please.

kbenoit (approved):

PR #2229, opened 2023-03-30
Fixes #1961

### LINKED ISSUE #1961: Reconsider handling of padding
Following a bug #1960, we have to reconsider how we handle paddings.
```
dfmt[,""] #  error
dfm_select(dfmt, "") # works
```

R treats empty names as a special case according to `?names()`:

> The name "" is special: it is used to indicate that there is no name associated with an element of a (atomic or generic) vector. Subscripting by "" will match nothing (not even elements which have no name).

kbenoit:
Allows subsetting a dfm by character names even when that character name is empty, e.g. ""

Whether we _want_ to allow this is another matter.  We probably should for features, given padding.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: quanteda
-Version: 3.2.5.9000
+Version: 3.2.5.9001
 Title: Quantitative Analysis of Textual Data
 Description: A fast, flexible, and comprehensive framework for 
     quantitative text analysis in R.  Provides functionality for corpus management,
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,8 @@
     
 * Added `user` and `system` arguments to `docvars()`, where `system = TRUE` returns internal document-level variables.
 
+* Indexing a dfm with empty feature names (`""`) is now possible.  Fixes #1961.
+
 ## Bug fixes and stability enhancements
 
 * `dfm_group()` now works correctly with an empty dfm (#2225).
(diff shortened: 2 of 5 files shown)


koheiw:
Easy to avid the error but the root cause is deep. 

### Feature names

I used `index_col[]` in the function to make the results as close as to `as.matrix(dfmt)[,x]`. If we allow slicing with "", `regex2id(valuetype = "fixed")` should be used to get a numeric index. Otherwise, the results can be different from `dfm_select()` when features have Unicode characters. 

### Document names

I used `index_row[]` to make the subsetting behaviours as close as to base R objects. Subsetting a named vector with "" returns `NA`, which triggers an error instead of returning NA rows.

``` r
require(quanteda)
#> Package version: 3.2.5.9000
#> Unicode version: 13.0
#> ICU version: 69.1
txt <- structure(c("A A A", "B B B", "C"), names = c("a", "b", ""))
lis <- as.list(txt)
corp <- corpus(txt)
toks <- tokens(corp)
dfmt <- dfm(toks) %>% 
    dfm_match(c("A", "B", "C", ""))

dfmt[,""]
#> Error in dfmt[, ""]: Subscript out of bounds
as.matrix(dfmt)[,""]
#> Error in as.matrix(dfmt)[, ""]: subscript out of bounds

txt[""]
#> <NA> 
#>   NA
lis[""]
#> $<NA>
#> NULL

corp[""]
#> Error in `[.corpus`(corp, ""): Subscript out of bounds
toks[""]
#> Error in `[.tokens`(toks, ""): Subscript out of bounds
dfmt["",]
#> Error in dfmt["", ]: Subscript out of bounds
```
I think we should keep the current behaviour with empty feature names unchanged and only disallow empty document names in corpus construction.

kbenoit:
That's fine, I suspected the fix would not be this easy. Let's keep it as is, and perhaps add a note about indexing to encourage users to use `dfm_select()` instead of indexing by name for empty features.

PR #2243, opened 2023-04-09
Improved patten2fixed more

koheiw:
Construct an index for glob patterns with * before and after words only when they exist.

```r
> require(quanteda)
> require(stringi)
> type <- stri_rand_strings(10000, 1:200)
> pat <- c(paste0("*", stopwords()), 
+          paste0(stopwords(), "*"))
+     quanteda3::pattern2fixed(pat, type),
+     quanteda::pattern2fixed(pat, type),
+     times = 10
+ )
Unit: milliseconds
 quanteda3::pattern2fixed(pat, type) 3061.4159 5475.2998 6375.6415 7347.7097 7412.6494 7515.9377    10
  quanteda::pattern2fixed(pat, type)  155.4013  157.6409  164.0289  164.7181  170.3604  171.2816    10
```

When * is only at the end, the speed is similar to the old version.

```r
> pat <- c(paste0(stopwords(), "*"))
+     quanteda3::pattern2fixed(pat, type),
+     quanteda::pattern2fixed(pat, type),
+     times = 10
+ )
Unit: milliseconds
                                expr     min      lq     mean   median      uq      max neval
 quanteda3::pattern2fixed(pat, type) 80.7532 83.6016 87.82977 85.61320 90.4319 105.9388    10
  quanteda::pattern2fixed(pat, type) 49.2142 50.7405 55.31276 55.08445 55.9890  70.9670    10
```


kbenoit (approved):
That's brilliant. Too bad we just missed it for 3.3.0, but I have a feeling there will be one more 3.3.x release while we work on 4.
We can make note of this and add it to some promotion of the performance gains for v4.

koheiw:
By making patterns to tokens ID coversion faster, we can really appreciate the xprt object.

PR #2250, opened 2023-04-12
Update pattern2id

koheiw:
Stop exporting `index_types()`. This makes it a bit more efficient.


kbenoit (approved):
👍

Good idea to keep a running set of 4.0 changes in the NOTES.md, and to update this version to 4.0.9000.

PR #2253, opened 2023-04-14
Avoid segfault error when patterns have paddings

koheiw:
This does nothing do with recent changes. It is a very old bug.


kbenoit (approved):

kbenoit:
Old bugs are the crunchiest!

PR #2254, opened 2023-04-14
Avoid segfault error when patterns have paddings in v3.3

koheiw:
We should update the v3.3 because it leads to a nasty crash.


kbenoit:
Oh that's not good. Thanks for finding and fixing it.
I introduced a slight namespace issue myself in the effort to fix one of our conflicts introduced with changing the docvars() generic. I'll submit an update tomorrow.

kbenoit (approved):

PR #2255, opened 2023-04-16
Article tokens xptr

koheiw:
A vignette article about `tokens_xptr`. Please merge #2251 first.


kbenoit:
This will need updating from master after #2251 is merged, then it will be easier to run and to see the precise diffs. I'll review it as soon as that's completed. Looking great though.

kbenoit:
I'll update `README.Rmd` for this too, since we should be talking about v4, not v3 changes any longer!

kbenoit:
I'm reviewing this, and dumb question here, but I cannot seem to build the `v3.3.x` branch after:
- renaming the package to `quanteda3` in `DESCRIPTION`
- changing the name in `quanteda-documentation.R:82`
- rebuilding with clean so that both RcppExports files are recreated

I get, on macOS:
```
── R CMD INSTALL ────────────────────────────────────────────────────────────────────────────────────────────────
─  installing *source* package ‘quanteda3’ ...
   ** using staged installation
   ** libs
   using C++ compiler: ‘Apple clang version 14.0.3 (clang-1403.0.22.14.1)’
   using SDK: ‘’
   clang++ -arch arm64 -std=gnu++17 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG -DARMA_DONT_PRINT_OPENMP_WARNING  -I../inst/include -I'/Library/Frameworks/R.framework/Versions/4.3-arm64/Resources/library/Rcpp/include' -I'/Library/Frameworks/R.framework/Versions/4.3-arm64/Resources/library/RcppParallel/include' -I'/Library/Frameworks/R.framework/Versions/4.3-arm64/Resources/library/RcppArmadillo/include' -I/opt/R/arm64/include    -fPIC  -falign-functions=64 -Wall -g -O2  -UNDEBUG -Wall -pedantic -g -O0 -c RcppExports.cpp -o RcppExports.o
   clang++ -arch arm64 -std=gnu++17 -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress -L/Library/Frameworks/R.framework/Resources/lib -L/opt/R/arm64/lib -o quanteda3.so RcppExports.o fcm.o index.o tokens_chunk.o tokens_compound.o tokens_lookup.o tokens_ngrams.o tokens_recompile.o tokens_replace.o tokens_restore.o tokens_segment.o tokens_select.o utility.o -L/Library/Frameworks/R.framework/Resources/lib -lRlapack -L/Library/Frameworks/R.framework/Resources/lib -lRblas -L/opt/gfortran/lib/gcc/aarch64-apple-darwin20.0/12.2.0 -L/opt/gfortran/lib -lgfortran -lemutls_w -lquadmath -F/Library/Frameworks/R.framework/.. -framework R -Wl,-framework -Wl,CoreFoundation
   ld: warning: could not create compact unwind for ___emutls_get_address: registers 23 and 24 not saved contiguously in frame
   installing to /private/var/folders/rc/y2gv0_dn0z19ygq271w7f9dr0000gp/T/RtmpCL8JmR/devtools_install_fba3fea0678/00LOCK-quanteda/00new/quanteda3/libs
   ** checking absolute paths in shared objects and dynamic libraries
─  DONE (quanteda3)
Error in getDLLRegisteredRoutines.DLLInfo(dll, addNames = FALSE) : 
  must specify DLL via a “DLLInfo” object. See getLoadedDLLs()
Calls: suppressPackageStartupMessages ... assignNativeRoutines -> getDLLRegisteredRoutines.DLLInfo
Execution halted

Exited with status 1.
```

koheiw:
https://github.com/quanteda/quanteda3

`useDynLib` needs to be quanteda3.

koheiw:
@kbenoit When can you merge this PR? I don't understand why you need more time to update articles longer than to review the core C++ functions. Updating articles should be much easier.

kbenoit (approved):
I did some tidying up, and added to the xptr article fairly significantly. In so doing, I also created a few new functions which @koheiw you will want to look at carefully.

These were

*  a `copy()` function that can be extended to other xptr objects later, for making deep copies. This is similar to `data.table::copy()`, another package that modifies in place and permits shallow copies.

* a `lengths()` method for `tokens_xtpr` class objects.  This coerces first to a `tokens()` object so it's probably worth writing a C++ version for, as I noticed you had for other very simple functions like `ndoc()`.

* tests for `copy()` and `lengths.tokens_xptr()` of course

I also reworked the documentation for the `tokens_xtpr` core methods so that the man page is now just `tokens_xptr`. If this looks too much like there is a `tokens_xptr()` constructor, then we should change this back to how it was before.

R/copy.R
```diff
@@ -0,0 +1,32 @@
...
+}
+
+#' @export
+copy.default <- function(x) {
+    return(x)
+}
+
+#' @export
+copy.tokens_xptr <- function(x) {
+    as.tokens_xptr(x)
```
koheiw, line 31: Don't you think `as.tokens_xptr(xtoks)` is enough? For me, it makes sense to have `copy()` only if there are other objects that users need to deep copy.

R/tokens_xptr.R
```diff
@@ -43,6 +60,13 @@ as.tokens_xptr.tokens_xptr <- function(x) {
     rebuild_tokens(result, attrs)
 }
 
+#' @method lengths tokens_xptr
+#' @noRd
+#' @export
+lengths.tokens_xptr <- function(x, use.names = TRUE) {
+    lengths(as.tokens(x))
```
koheiw, line 67: You must not use `as.tokens()` because it is very expensive. use `quanteda:::cpp_ntoken()`, which is used in `ntokens.tokens_xptr()`.

R/copy.R
koheiw: You always tried to avoid expanding the package names space.

R/tokens_xptr.R
koheiw: [Or `ntoken(x))` i](https://github.com/quanteda/quanteda/blob/df6204052012f1db20e2a7933fca0bfb08ac55d9/R/tokens_xptr.R#L80-L83C2)

koheiw: I am not sure if we should have `lengths.tokens_xptr()` because we always recommended users to use quanteda functions, `ntoken()`. It might be even misleading because it make tokens_xptr objects look like list-based objects.

R/copy.R
kbenoit: I think the semantics of `copy()` are easier and more consistent for the user than coercing each xptr object to its plain version, plus there is precedent in another package.

But it makes sense to me to avoid if it we are just talking about tokens for now. if we expand the xptr range, we can always add it back then. How about I just comment out this code for now?

R/tokens_xptr.R
kbenoit: I agree that it's not something we should promote, and it's not in the package index. But a list-like object should work with list-like functions.

kbenoit: Resolved in [95bd25a](https://github.com/quanteda/quanteda/pull/2255/commits/95bd25a33c16e2119c3cc439963d52b14ce3e869)

R/copy.R
kbenoit: Removed in [d50727d](https://github.com/quanteda/quanteda/pull/2255/commits/d50727df08d027698f433256e81eec0abd102afc)

kbenoit:
Good idea, we could add mention of that to the vignette. But once again, not as intuitive as `copy()`.

koheiw:
I suppose most of the people only use the classic tokens object. If so, we don't need to worry about deep copying.

PR #2259, opened 2023-04-21
Fix parallel

koheiw:
Pass `quanteda_options("threads")` to C++ functions to control parallel computing. Currently, the option has no effect.


kbenoit (approved):
So as I understand it, besides tidying up the `quanteda_options()` code for setting options, this now makes sure that `quanteda_options("threads")` is passed to each cpp function. But when the option is set, we don't know if it's valid besides checking this via `cpp_get_max_thread()`. So, if anything goes wrong, it will happen as a runtime error within a function relying on the threads setting. 

This is probably fine, but I wonder if an even safer approach would be to call some trivial function upon setting the threads options that multi-threads and should be silent if it works, or throw some error if not.

Your call since you understand this aspect of TBB vastly more than I.

kbenoit:
Please add an update to `NOTES.md` about this however.

koheiw:
It is a genuine concern. We check values in `quanteda_options()` but user can access directly via `options()`. So I added `get_threads()` that validate the option values. More generally, we should move validation of options from when they are entered to whey they are used.

PR #2263, opened 2023-05-08
Fix nfunctions for tokens_xptr

kbenoit:
Tweaks the nfunctions and adds the `[[` operator to the new `tokens_xptr` class. We want the xptr functions to be completely consistent with the regular tokens functions, and this PR fills in some of the gaps.

- adds `[[.tokens_xptr` (this function was missing before)
- adds `...` functionality to `ntoken.tokens_xptr()`
- adds `ntype.tokens_xptr()`, with `...` functionality
- adds tests

@koheiw I had strange issues with shallow copying of objects within the functions, when `...` was applied, so implemented a bit of a hack to shallow copy each object unless options were applied, then it deep copies. You probably know a better way to do this.


R/tokens_xptr.R
```diff
@@ -142,6 +163,13 @@ as.list.tokens_xptr <- function(x, ...) {
     )
 }
 
+#' @method [[ tokens_xptr
+#' @export
+"[[.tokens_xptr" <- function(x, i) {
+    types <- c("", get_types(x))
+    types[unclass(as.tokens(x))[[i]] + 1] # shift index to show padding
+}
```
koheiw, line 171: How about this?
```r
as.list(as.tokens(x[i]))[[1]]
```

R/tokens_xptr.R
```diff
@@ -83,7 +83,28 @@ types.tokens_xptr <- function(x) {
...
+        x2 <- x
+    }
+    ret <- structure(cpp_ntoken(x2), names = docnames(x2))
+    rm(x2)
+    ret
+}
+
+#' @export
+ntype.tokens_xptr <- function(x, ...) {
+    if (length(list(...))) {
```
koheiw, line 99: It should be like
```
length(types(tokens(x, ...)))
```

R/tokens_xptr.R
```diff
@@ -83,7 +83,28 @@ types.tokens_xptr <- function(x) {
 
 #' @export
 ntoken.tokens_xptr <- function(x, ...) {
-    structure(cpp_ntoken(x), names = docnames(x))
+    if (length(list(...))) {
```
koheiw, line 86: It should be like
```
structure(cpp_ntoken(tokens(x, ...)), names = docnames(x))
```

R/tokens_xptr.R
kbenoit: Not sure what you mean here... the `length()` conditional is just to check whether tokens needs to be called again.

tests/testthat/test-tokens_xptr.R
```diff
@@ -312,3 +312,43 @@ test_that("lengths works on tokens xptr objects", {
...
+    expect_identical(
+        ntoken(toks), 
+        ntoken(xtoks)
+    )
+    expect_identical(
+        ntoken(toks, remove_punct = TRUE), 
+        ntoken(xtoks, remove_punct = TRUE)
+    )
+    # check that the former removals did not modify xtoks
+    expect_identical(
```
kbenoit, line 337: This second check should not fail - but does if the former call is used without the x2 deep copies. But @koheiw you might know a better way. I suggest you try improving this.

R/tokens_xptr.R
kbenoit: That was the first thing that I tried, but it then fails the test if called a second time, since a call with any `...` options modifies x outside of the function scope. If this function body is just replaced with that code, it will fail in `test_that("n functions works the same on both tokens and tokens_xptr"` - see my note there.

R/tokens_xptr.R
koheiw: `tokens()` does not (should not) do anything if nothing is passed via `...`, so conditioning is redundant.

koheiw:
You are right, then
```
structure(cpp_ntoken(tokens(as.tokens_xptr(x), ...)), names = docnames(x))
```

koheiw:
On master

``` r
require(quanteda)
#> Package version: 4.0.0
#> Unicode version: 14.0
#> ICU version: 70.1
#> Parallel computing: 4 of 4 threads used.
require(testthat)
#> Loading required package: testthat
toks <- tokens(c("A a a, b.", "A b c c c?"))
xtoks <- as.tokens_xptr(toks)

identical(
    ntoken(toks, remove_punct = TRUE), 
    structure(quanteda:::cpp_ntoken(tokens(as.tokens_xptr(xtoks), remove_punct = TRUE)),
              names = docnames(xtoks))
)
#> [1] TRUE
identical(
    ntoken(toks), 
    ntoken(xtoks)
)
#> [1] TRUE

identical(
    ntoken(toks), 
    structure(quanteda:::cpp_ntoken(tokens(as.tokens_xptr(xtoks))),
              names = docnames(xtoks))
)
#> [1] TRUE
```

<sup>Created on 2023-05-08 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/tokens_xptr.R
```diff
@@ -166,8 +166,7 @@ as.list.tokens_xptr <- function(x, ...) {
 #' @method [[ tokens_xptr
 #' @export
 "[[.tokens_xptr" <- function(x, i) {
-    types <- c("", get_types(x))
-    types[unclass(as.tokens(x))[[i]] + 1] # shift index to show padding
+    as.list(as.tokens(x[i]))[[1]]
```
koheiw, line 169: It errors if `i = integer()`, so this is better. 
```
as.character(as.tokens(x[i]))
```

kbenoit:
@koheiw would you mind making these changes directly?

koheiw:
I think we should remove `...` from `ntoken.tokens` and `ntype.tokens` because  they are on-the-fly functions.  In designing quanteda, especially methods for tokens_xptr, we should assume that the objects have hundreds of millions of tokens. I am fine with adding `"[[.tokens_xptr"`, however.

kbenoit:
I agree that we should aim to remove on-the-fly operations such as the `...` in the two nfunctions.
But we should deprecate them in v4 rather than simply remove it. And if we extend those functions to `tokens_xptr`, we should make the full set of arguments functional too. Otherwise we are saying, we implemented xptr objects, use them the same way as regular tokens objects, although be careful in that we changed the behaviours we never really liked in some of the functions.

Right now it exists but is non-functonal and generates no error message, which is the worst outcome.

koheiw:
Let's deprecate these methods for tokens. We should add `...` to methods for tokens_xptr but it will be passed to `unused_dots()`.

kbenoit:
I'd prefer to make the dots functional for the new methods, and (in a separate PR) then deprecate all of them (for tokens and tokens_xptr). Let's be as consistent as possible.

koheiw:
Why don't you deprecate them in v3 and defunct in v4?

kbenoit:
I thought about that but removing functionality really needs a major version deprecation, and I think we can release v4 by summer. It's not enough time for deprecation. I can just see the issues, emails, and SO posts now! Better to phase out.

Big deprecation and defuncting wave coming, once we get these PRs resolved!

koheiw:
I think its fine to removing them quickly because only few people use it. If you need more time, we must postpone the release of v4.

kbenoit:
The point of major releases is to shift from deprecated to defunct. But since this is a new function, go ahead and add a .Defunct or some other unused message to these methods for `ntype.tokens_xptr()` and `ntoken.tokens_xptr()`. I'll sort out the other methods separately. We still need this PR since there was not an `ntype.tokens_xptr()` at all before, nor was there a `[[` method.

kbenoit:
The point of major releases is to shift from deprecated to defunct. But since this is a new function, go ahead and add a .Defunct or some other unused message to these methods for `ntype.tokens_xptr()` and `ntoken.tokens_xptr()`. I'll sort out the other methods separately. We still need this PR since there was not an `ntype.tokens_xptr()` at all before, nor was there a `[[` method.

kbenoit:
The point of major releases is to shift from deprecated to defunct. But since this is a new function, go ahead and add a .Defunct or some other unused message to these methods for `ntype.tokens_xptr()` and `ntoken.tokens_xptr()`. I'll sort out the other methods separately. We still need this PR since there was not an `ntype.tokens_xptr()` at all before, nor was there a `[[` method.

kbenoit:
The point of major releases is to shift from deprecated to defunct. But since this is a new function, go ahead and add a .Defunct or some other unused message to these methods for `ntype.tokens_xptr()` and `ntoken.tokens_xptr()`. I'll sort out the other methods separately. We still need this PR since there was not an `ntype.tokens_xptr()` at all before, nor was there a `[[` method.

kbenoit:
Closing this but let's continue the discussion in #2269.

PR #2265, opened 2023-05-08
Restore readtext

kbenoit:
Restores the readtext extensions to **quanteda**. It's much easier for the packages we maintain to have extensions in **quanteda**, since this prevents the signatures from becoming out of sync (as they became with the now-abandoned added system arguments to `docvars()`).

I cleaned this up and added tests that were not present before.


kbenoit:
Doesn't look like these tests will pass for a merge to non-master branch, so @koheiw you should run CHECK locally. Passing on my machine.

koheiw:
Why do you need special methods for a data.frame?

koheiw:
You should add `v3_3` to the YAML file to trigger CI tests like this:

https://github.com/quanteda/quanteda/blob/c92875c92e6f943033740907048f5b1ea64c2dd5/.github/workflows/check-standard.yaml#L4-L7

kbenoit:
It's to restore what already works and worked for years. Since these are quanteda methods and we don't want quanteda to be a necessary import for the readtext package - so that it can stand alone - it's easier to keep them in quanteda.

In fact the reason I removed them recently was just a quick fix because you changed the signature of docvars() which would have broken some downstream packages, including readtext. But this only underscored for me the importance of keeping the methods in quanteda, so that we would not have that sort of issue arise again if we changed the function.

koheiw:
You are making your life difficult by creating method only for a data.frame. What is the point of using `ndoc()` instead of `nrow()` and `docvars()` instead of `$`? We need these methods for corpus, tokens and DFM because they are complex objects.

I don't care about v3.x much, but definitely don't want methods for readtext in v4.x.

kbenoit:
That's fine, but let me run extensive tests on that first, since removing them is already causing issues. Meanwhile let's get this back into v3 so I can update it on CRAN.

koheiw (approved):

PR #2266, opened 2023-05-08
Fix 1961

### LINKED ISSUE #1961: Reconsider handling of padding

kbenoit:
I think this fixes #1961, but it was done earlier and then not merged, so @koheiw you might have a good reason for this (like it being superseded by more fundamental changes in v4). 

Take a look.


koheiw:
Why do you want to make selection by element names inconsistent with corps, tokens, and base-R objects? Again, `dfmt["",]` should error according to R's element name rules. 

If you want to make `dfmt[,""]` work, you should call `dfm_select(x, j, valuetype = "fixed", case_insensitive = FALSE)` so it works in the same way as `dfm_select(dfmt, "")`. Note that such changes in `[]` affect DFM and FCM.

kbenoit:
That's fine, I thought we must have agreed sth like that. I'll close the PR.

PR #2267, opened 2023-05-09
Fix v3 matrix 1.5.5

### LINKED ISSUE #2258: Fix tests incompatible with forthcoming version of the Matrix package
I received the following communication today from Mikael Jagan:

> Matrix 1.5-5 will fix a number of bugs affecting subscripts of objects
> inheriting from the virtual class 'Matrix'.  Your package quanteda seems
> to rely on the following returning character(0) instead of NULL:
> 
>     rownames(<Matrix with rownames>[integer(0), , drop = FALSE])
>     colnames(<Matrix with colnames>[, integer(0), drop = FALSE])
> 
> Matrix previously gave character(0) in some cases but going forward will
> always give NULL, to be consistent with base R's handling of traditional
> matrices.  This fix causes some of quanteda's tests to fail.
> 
> Please correct and verify that your package passes its checks under the
> development version of Matrix, which you can install with Subversion,
> like so:
> 
> $ svn checkout svn://svn.r-forge.r-project.org/svnroot/matrix/pkg/Matrix
> $ R CMD build Matrix
> $ R CMD INSTALL Matrix_1.5-5.tar.gz
> 
> {
>     Note that installing in the usual way
> 
>     > install.packages("Matrix", repos = "http://r-forge.r-project.org/")
> 
>     seems to provide an outdated version of Matrix, due to a glitch
>     in R-Forge, which I hope is just temporary ...
> 
>     If you do not have Subversion, then I can try to host a tarball on
>     GitHub for you ...
> }

kbenoit:
Fixes #2258 

- Uses a bit of a hack in `dfm_select()` to ensure the expected result when the result of selection would be to remove all of an fcm's features.
- Comments out two tests that we might not want as expected behaviours anyway

Unlike the trend that the new Matrix package is embracing - to return `NULL` instead of `character(0)` for empty dimension names, in quanteda, docnames() and featnames() are always character. If there are none, we could consider returning `NULL` instead. `length(NULL) == 0` is `TRUE`, so our conditionals based on the length of empty dimension names would still work.

I've tested this under both the current CRAN version of **Matrix** (1.5.4) and the forthcoming 1.5.5, installed using the subversion pull, build, and install suggested in #2258.


R/object-builder.R
kbenoit, line 1: @koheiw this simplified the fix, now it works with `fcmat[, integer(0)]` and still returns a `character(0)` as `featnames`. This preserves existing behaviours yet means that the package will work with Matrix 1.5.5. 

Changing to adhere to empty dimnames being NULL is a policy shift we should consider, but I suggest we do that separately for 4.0 when we can test its full consequences.

koheiw (approved):
It would be great if you could fix the compatibility issue with these small changes.

PR #2268, opened 2023-05-09
Add ntype.tokens_xptr() and [[.tokens_xptr

koheiw:
Redo #2263


koheiw:
Do we allow recursive indexing?
```
> toks <- tokens("a b c")
> toks[[c(1, 3)]]
[1] "c"
```
I thought I posted this yesterday, but went somewhere...

kbenoit:
On recursive indexing: In short, yes, because we want the behaviour to be the same as lists of character vectors. Although: I think few users are aware of this behaviour for lists, or use it. But best to be consistent, e.g.

``` r
library("quanteda")
#> Package version: 3.3.0.9001
#> Unicode version: 14.0
#> ICU version: 71.1

toks <- tokens("a b c")
toks[[c(1, 3)]]
#> [1] "c"

lis <- list(c("a", "b", "c"))
lis[[c(1, 3)]]
#> [1] "c"
```

<sup>Created on 2023-05-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

kbenoit:
@koheiw feel free to merge this when you're happy with it

koheiw:
Let me disable recursive indexing because it is unreasonably complex for tokens_xptr.

kbenoit:
I'm fine with disabling that, since I always thought it was counter-intuitive in base R anyway (and never used it).
We could consider adding a note to `tokens-class.Rd` about this, but since we don't document `[[` anyway, then I think we can just leave it as is.

PR #2273, opened 2023-05-11
Fix v4 matrix 1.5.5

kbenoit:
Solves #2258

We ought to merge this to 4.0 as well.

I've opened #2274 about the dimnames() returning NULL, versus the desired behaviour our abstraction of `featnames()` and `docnames()`, which we can consider making for the 4.0 release.


kbenoit:
`readtext` functions removed

kbenoit:
@koheiw Can we merge this?

koheiw (approved):

PR #2278, opened 2023-05-15
Fix email

koheiw:
Current regex for email is too simple.

### DIFF
--- a/R/tokenizers.R
+++ b/R/tokenizers.R
@@ -160,7 +160,7 @@ tokenize_word4 <- function(x, split_hyphens = FALSE, split_tags = FALSE, split_e
     
     ftp <- "s?ftp://[-+a-zA-Z0-9@#:.%~=_&/]+"
     http <- "(https?://)?(www.)?[-a-zA-Z0-9]+(\\.[-a-zA-Z0-9]+)+([/?#][-+a-zA-Z0-9@#:.%~=_&]+)*[/?#]?"
-    email <- "[-a-zA-Z0-9_.]+@[-a-zA-Z0-9]+(\\.[-a-zA-Z0-9]+)+\\.[a-z]{2,4}"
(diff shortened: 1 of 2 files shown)


R/tokenizers.R
```diff
@@ -160,7 +160,7 @@ tokenize_word4 <- function(x, split_hyphens = FALSE, split_tags = FALSE, split_e
     
     ftp <- "s?ftp://[-+a-zA-Z0-9@#:.%~=_&/]+"
     http <- "(https?://)?(www.)?[-a-zA-Z0-9]+(\\.[-a-zA-Z0-9]+)+([/?#][-+a-zA-Z0-9@#:.%~=_&]+)*[/?#]?"
-    email <- "[A-Za-z0-9_]+@[A-Za-z][A-Za-z0-9_]+\\.[a-z]+"
+    email <- "[-a-zA-Z0-9_.]+@[-a-zA-Z0-9]+(\\.[-a-zA-Z0-9]+)*\\.[a-z]{2,4}"
```
odelmarcelle, line 163: I would perhaps drop the quantifier at the end. I looked it up and domains longer than 4 character such as .london or .tokyo exists.

Also, the leading part of the email can sometimes accept '+'. I would add it to the first character set.

koheiw: Thanks. I updated accordingly.

kbenoit: Here are some nightmare scenarios but pretty rare. I can't see an easy way to deal with the spaces issue, so I propose we just ignore these. (The one starting with an underscore works.)
https://haacked.com/archive/2007/08/21/i-knew-how-to-validate-an-email-address-until-i.aspx/

kbenoit (approved):
Would be good to merge this into v3_3 as well, before I send that update to CRAN later this week.

PR #2280, opened 2023-05-17
Small fixes for 3.3.1 CRAN update

kbenoit:
Updates the NOTES and other items, but also removes some internal (non-exported) methods that matched exported generics. We were not using those methods anyway, and they were targeted for removal on our deprecations and slimming and trimming for 4.0 anyway.

### DIFF
--- a/R/tokens_ngrams.R
+++ b/R/tokens_ngrams.R
@@ -48,8 +48,7 @@ tokens_ngrams.default <- function(x, n = 2L, skip = 0L, concatenator = "_") {
 ## the grammatical rules of quanteda (inputs character, outputs tokens),
 ## but starts with "tokens_"
 #' @importFrom stats complete.cases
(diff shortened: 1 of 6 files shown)


R/tokens_ngrams.R
```diff
@@ -48,8 +48,7 @@ tokens_ngrams.default <- function(x, n = 2L, skip = 0L, concatenator = "_") {
 ## the grammatical rules of quanteda (inputs character, outputs tokens),
 ## but starts with "tokens_"
 #' @importFrom stats complete.cases
-tokens_ngrams.character <- function(x, n = 2L, skip = 0L, concatenator = "_") {
-
+tokens_ngrams_character <- function(x, n = 2L, skip = 0L, concatenator = "_") {
```
kbenoit, line 51: I changed this to not be a method of the generic for which it was inappropriately named, but I think we agree that for 4.0, we should remove this anyway since it shortcuts the tokenisation step.

koheiw (approved):

PR #2285, opened 2023-07-04
Fix tokens group

koheiw:
Due to the double loop in the current code for `tokens_group()`, R almost freezes when the number of documents (sentences) are very large (ndoc > 10M).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,7 @@ language handling, and customisation options.
 
 ## Bug fixes and stability enhancements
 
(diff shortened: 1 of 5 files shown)


kbenoit (approved):
👍

kbenoit:
Now that summer is here I have plenty of time for quanteda and ready to get 4.0 finalised. 🚀

koheiw:
I have been using v4.0 myself for two months or more. This is the only problem I faced.

PR #2287, opened 2023-07-18
Use .empty() where appropriate

MichaelChirico:
See https://releases.llvm.org/5.0.1/tools/clang/tools/extra/docs/clang-tidy/checks/readability-container-size-empty.html.


koheiw (approved):
Looks great. Thank you!

PR #2289, opened 2023-08-06
Added link to as.dfm() in dfm() docs

khughitt:


khughitt:
It might also be worth mentioning `as.dfm()` in the quick start guide, but I figured I would leave that up to you to decide.

kbenoit (approved):

Issue #2299, opened 2023-10-16
Error when combining more than 3 tokens objects


koheiw:
```r
> require(quanteda)
> 
> toks1 <- tokens(c("doc1" = "a a a"))
> toks2 <- tokens(c("doc2" = "b b b"))
> toks3 <- tokens(c("doc3" = "c c c"))
> toks4 <- tokens(c("doc4" = "d d d"))
> 
> c(toks1, toks2, toks3, toks4)
Error in x[[seq(3, length(x))]] : subscript out of bounds
```

PR #2300, opened 2023-10-16
Issue 2299

koheiw:
Address #2299


kbenoit (approved):

kbenoit:
@koheiw that's approved but I am not sure what's happening with ubuntu-latest (devel).

koheiw:
I think the TBB library is new on the ubuntu-latest(devel), so quanteda.textstats is failing.

PR #2304, opened 2023-10-23
Deprecations and removals for 4.0

### LINKED ISSUE #2070: Use the lifecycle package to mark functions according to maturity
This allows us to tag functions explicitly according to whether they are mature, developing, deprecated, defunct, etc.

https://github.com/r-lib/lifecycle

### LINKED ISSUE #2241: Deprecate nsentence()
It should be done in this way
```r
ntoken(tokens(txt, what = "sentence"))
```

### LINKED ISSUE #2269: Removing ... tokens options from ntoken(), ntype()
Since we had a discussion about this on #2263, I thought it could be useful to continue that discussion here, rather than it potentially holding up future PRs.

I agree, we should be limiting what is done "on the fly". This means we should disallow, with or without options:
- [ ] `ntype.character()`
- [ ] `ntype.corpus()`
- [ ] `ntoken.character()`
- [ ] `ntoken.corpus()`

But it actually makes _more_ sense to allow ... removals for `tokens_xptr` objects than for any other type of object, since the removals are efficient for already tokenised objects, and since there is no overhead of passing by value to/from C++.

The "on-the-fly" part we should focus on removing is skipping an explicit tokenisation step. But since we allow `tokens()` removal or splitting options on `tokens` objects, and designed all of the options to be from more to less operations (always removals or splitting things not split), then we should be encouraging users to "post-process" tokens objects using options they might have instead destructively applied on the first tokenisation. Especially for very large objects, this approach can save a lot of time, since it's the segmentation that takes more time than removals.

Without the ability to pass these options as an intermediate object to the ntoken/ntype functions for tokens/tokens_xtpr, the syntax is counter-intuitive as well as fairly horrible when it comes to tokens_xptr in particular, for something as simple as wanting to know what a set of post-tokenisation removals will trim. (I use this occasionally and always in teaching.)

``` r
library("quanteda")
#> Package version: 4.0.0
#> Unicode version: 14.0
#> ICU version: 71.1

toks <- tokens(head(data_corpus_inaugural, 5))
xtoks <- as.tokens_xptr(toks)

# total punctuation in corpus
ntoken(toks) - ntoken(toks, remove_punct = TRUE)
#> 1789-Washington 1793-Washington      1797-Adams  1801-Jefferson  1805-Jefferson 
#>             107              12             259             197             214

# but for xtoks: potential error
# works once
ntoken(xtoks) - ntoken(tokens(xtoks, remove_punct = TRUE))
#> 1789-Washington 1793-Washington      1797-Adams  1801-Jefferson  1805-Jefferson 
#>             107              12             259             197             214
# fails on second try
ntoken(xtoks) - ntoken(tokens(xtoks, remove_punct = TRUE))
#> 1789-Washington 1793-Washington      1797-Adams  1801-Jefferson  1805-Jefferson 
#>               0               0               0               0               0

# solution is a deep copy, which is the least efficient of all
toks <- tokens(head(data_corpus_inaugural, 5))
xtoks <- as.tokens_xptr(toks)
# works once
ntoken(xtoks) - ntoken(tokens(as.tokens_xptr(xtoks), remove_punct = TRUE))
#> 1789-Washington 1793-Washington      1797-Adams  1801-Jefferson  1805-Jefferson 
#>             107              12             259             197             214
# works twice
ntoken(xtoks) - ntoken(tokens(as.tokens_xptr(xtoks), remove_punct = TRUE))
#> 1789-Washington 1793-Washington      1797-Adams  1801-Jefferson  1805-Jefferson 
#>             107              12             259             197             214
```

<sup>Created on 2023-05-09 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### LINKED ISSUE #2276: Deprecations and removals for 4.0
- [x] Defunct the deprecated functions
- [x] Do this for legacy argument handling too - perhaps using the **lifecycle** package
- [x] Deprecate functions and arguments some not previously deprecated

See also #2269, #2249, #2241

kbenoit:
* Adopts the **lifecycle** package for better removals and deprecations, per #2070
* Implements the removals listed in #2276 
* Deprecates `nsentence()` per #2241
* Removes the on-the-fly options and `ntoken()`, `ntype()` applied to character or corpus objects, per #2269

closes #2070, closes #2276, closes #2241, closes #2269

To do still on this PR:
- [x] When issuing the PR, still need to defunct the trapped arguments to `dfm()`, see https://github.com/quanteda/quanteda/issues/2276#issuecomment-1775389275
- [x] Still need to clear `revdep_check()` for seeing how and what we break in dependencies.


kbenoit:
I issued PRs for two packages (including **LSX** 1.3.2) and also changed the lifecycle to defunct for the legacy `dfm()` arguments (deprecated in 3.0).

So this PR should be ready to go.

koheiw (approved):

Issue #2237, opened 2023-04-04
Easier way to subsetting objects by ntoken


koheiw:
I have been thinking about an easier way to subset objects by ntoken. Currently, there is only one way to do it.

```r
require(quanteda)
corp <- data_corpus_inaugural |>
    corpus_reshape()
 
toks <- tokens(corp) |>
    tokens_remove(stopwords()) 
toks <- tokens_subset(toks, ntoken(toks) > 10)
```

How about allowing `tokens_subset(toks, ntoken_ > 10)`, which is similar to `tokens_subset(toks, segid_ == 1)`? This is especially useful when we use pipes.

```r
toks <- tokens(corp) |>
    tokens_remove(stopwords()) |>
    tokens_subset(toks, ntoken_ > 10)
```
We can implement this just by adding `ntoken_` as a system-level document variable, which can be extracted by `docvars(toks, system = TRUE)`. If we want to do the same with corpus, the name of the variable would be `length_` or `n_` to include both `ntokens` and `nchar`.

kbenoit:
Wouldn't this require recomputing and updating these system docvars each time the object was subsetted or ['d? What if a user circumvented these?

These functions are super fast, e.g. `lengths()` for tokens, `length()` for corpus, so it's not a problem to compute them each time, but I agree the syntax is a bit cumbersome.

What about having some special functions like `.SD` in data.table, so that `.ntoken` referred to the tokens in `x` and was computed on the fly? so:

```r
toks <- tokens(corp) |>
    tokens_remove(stopwords()) |>
    tokens_subset(.ntoken > 10)
```

and this triggers the `ntoken(x)` code inside the function?

koheiw:
I don't like how data.table works much. I thought adding a docvar is best for the `*_subset` but we can add `min_ntoken` and `max_ntoken` arguments like `*_trim()`.

kbenoit:
It's a bit more like **magrittr's** use of `.` than **data.table**'s use of functions like `.SD()`, since we don't need a function, but rather a reference to the `x`, like **magrittr**.

So for functions that would operate on `x`, we could just use `.ntoken`, `.min_ntoken`, etc as part of the logical. Then we evaluate the expression by calling the function or argument on `x`. This is also simpler than cacheing elements like the number of tokens and trying to keep it updated.

koheiw:
[...]

PR #2319, opened 2023-11-27
Add min_ntoken and max_ntoken

koheiw:
I think adding `min_ntoken` and `max_ntoken` is the best solution to #2237


kbenoit (approved):
I'm fine with this. Should we also add these options to `tokens_subset()`?

koheiw:
They are added to both `tokens_subset()` and `dfm_subset()`.

PR #2320, opened 2023-11-28
Save tokenizer

koheiw:
Address #2313. We might want to drop `what` from the object meta fields because it is unreliable and redundant.


kbenoit (commented):
Nice change. Good to record this.

Issue #2322, opened 2023-12-06
Add pass argument to tokens functions to return documents intact


koheiw:
Sometimes, tokens selection need to be conditional on other tokens or document variables, but we cannot do this efficiently using existing functions. In the example, "C" and "R" refer to computer languages in the first texts but they are trademark or copyright symbols in the second. How can we remove them from the second text while keeping the first text intact?

``` r
require(quanteda)
txt <- c("R and C are languages",
         "Windows (R), Quanteda (C)")
toks <- tokens(txt, remove_symbols = TRUE, remove_punct = TRUE)
toks
#> Tokens consisting of 2 documents.
#> text1 :
#> [1] "R"         "and"       "C"         "are"       "languages"
#> 
#> text2 :
#> [1] "Windows"  "R"        "Quanteda" "C"
# Languages are removed
tokens_select(toks, min_nchar = 2)
#> Tokens consisting of 2 documents.
#> text1 :
#> [1] "and"       "are"       "languages"
#> 
#> text2 :
#> [1] "Windows"  "Quanteda"
````
Currently, we can achieve this by splitting tokens into two, apply `tokens_select()` and combine them. This is very inefficient.
```r
# Currently method
c(toks[1], tokens_select(toks[2], min_nchar = 2))
#> Tokens consisting of 2 documents.
#> text1 :
#> [1] "R"         "and"       "C"         "are"       "languages"
#> 
#> text2 :
#> [1] "Windows"  "Quanteda"
```
So, how about adding an argument that simply pass documents without changes when `TRUE`?
```r
# Proposed method
tokens_select(toks, min_nchar = 2, pass = c(TRUE, FALSE))
#> Tokens consisting of 2 documents.
#> text1 :
#> [1] "R"         "and"       "C"         "are"       "languages"
#> 
#> text2 :
#> [1] "Windows"  "Quanteda"

# This allows conditioning based on patters or docvars
tokens_select(toks, min_nchar = 2, 
              pass = ntoken(tokens_select(toks, "language*")) > 0)
tokens_select(toks, min_nchar = 2, 
              pass = toks$year > 1960)
```
I can add this to `tokens_select()`, `tokens_compound()`, `tokens_tolower/upper()`, `tokens_split()` and `tokens_replace()`.

kbenoit:
[...]

PR #2323, opened 2023-12-07
Issue 2322

koheiw:
Add a new argument for #2322. It is tentatively named `condition` but could be `cond`, `run_if`, `iff`, `filter` etc. I had an ambition to enable NSE but it did not because it over-complicate the functions. I am planing to add the same argument to `tokens_compound()`.

``` r
require(quanteda)
#> Package version: 4.0.0
#> Unicode version: 15.1
#> ICU version: 74.1
dat <- data.frame(text = c("R and C are languages",
                           "Windows (R), Quanteda (C)",
                           "Sizes are X=10, Y=20, Z=30"),
                  topic = c("language", "software", "hardware"),
                  month = c(NA, 4, 12))
corp <- corpus(dat)
toks <- tokens(corp)

toks1 <- tokens_select(toks, min_nchar = 2, condition = toks$topic != "language")
toks1
#> Tokens consisting of 3 documents and 2 docvars.
#> text1 :
#> [1] "R"         "and"       "C"         "are"       "languages"
#> 
#> text2 :
#> [1] "Windows"  "Quanteda"
#> 
#> text3 :
#> [1] "Sizes" "are"   "10"    "20"    "30"

docname <- docnames(toks)
toks2 <- toks %>% 
    tokens_select(c("R", "C"), condition = docname == "text1") %>% 
    tokens_select(c("windows", "quanteda"), condition = docname == "text2") %>% 
    tokens_select("\\d", valuetype = "regex", condition = docname == "text3")
toks2
#> Tokens consisting of 3 documents and 2 docvars.
#> text1 :
#> [1] "R" "C"
#> 
#> text2 :
#> [1] "Windows"  "Quanteda"
#> 
#> text3 :
#> [1] "10" "20" "30"

toks3 <- toks |>
    tokens_keep(c("R", "C"), condition = toks$topic == "language") %>% 
    tokens_remove(min_nchar = 6, condition = toks$topic != "language")
toks3
#> Tokens consisting of 3 documents and 2 docvars.
#> text1 :
#> [1] "R" "C"
#> 
#> text2 :
#> [1] "Windows"  "Quanteda"
#> 
#> text3 :
#> character(0)

toks4 <- toks |>
    tokens_remove(min_nchar = 2, condition = toks$month) |>
    tokens_remove(stopwords())
toks4
#> Tokens consisting of 3 documents and 2 docvars.
#> text1 :
#> [1] "R"         "C"         "languages"
#> 
#> text2 :
#> [1] "Windows"  "Quanteda"
#> 
#> text3 :
#> [1] "Sizes" "10"    "20"    "30"
```

<sup>Created on 2023-12-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


koheiw:
I added the `condition` argument to `tokens_compound()` and tokens_replace()`.

kbenoit (commented):
Looks good. A few comments. 

1. What if call this something more descriptive? For instance, `document_mapif` or `document_mask`? I'm fine with the latter if you want to avoid tidyverse language. `condition` seems too general.  
2. Let's put the documentation for this in its own function and then @inheritParams it, like valuetype etc.
3. I like how it's positioned last, just before `verbose`. But then I noticed that `tokens_compound()` and `tokens_split()` have no verbose option. Should they?
4. This is added to `tokens_compound()` -- should it not also be added to `tokens_split()`?

koheiw:
`condition` is too general but `document_mapif` are `document_mask` too long. Since not many people would use this, this argument can be less visible. My favorite is `iff` or `run_if` for this reason. It is not really mapping. 

I think documents should be processed when `TRUE` to be consistent with `tokens_subset()`, but `mask` is not clear what happens when `TRUE` (mask or unmask?).  Other possible names are `include_doc` or `execute`.

I can add the new argument to `tokens_split()`. Generally, verbose messages needs to be improved package wide.

kbenoit:
Fair point on the use of "mask", it's ambiguous. What if we call it `map_if`, which has a recognisable precedent in a widely used package? (just as we adopted some of the **stringi** argument names) 

`?purrr::map_if`:

`.p` ...Only those elements where `.p` evaluates to `TRUE` will be modified

kbenoit:
Also I think this should work without `exclusive = FALSE`. Sometimes you just want the tokens looked up and their keys.

kbenoit:
How about:
- [ ] I change the argument to `map_if` and document this as suggested above
- [ ] You add this to `tokens_split()`
- [ ] I'll raise the verbose review as a new issue

koheiw:
Mapping seems to imply switching between documents to assign (or function to apply) depending on a given condition. In our case, it is always the same document.

I have only minimally inspired by tidyverse packages, but `modify_if` sounds fine:

> modify_if() only modifies the elements of x that satisfy a predicate and leaves the others unchanged.

kbenoit (approved):

Issue #2324, opened 2023-12-07
Upgrading tokens_replace() to keep tokens and keys togather


koheiw:
We sometimes keep track of tokens matched to dictionary patters but it is not easy (see https://github.com/quanteda/quanteda/issues/2063). `tokens_replace()` can be used add keys to original tokens (e.g. "United States/Countries") but only with fixed patterns.  

``` r
require(quanteda)
txt <- c(d1 = "The United States is bordered by the Atlantic Ocean and the Pacific Ocean.",
         d2 = "The Supreme Court of the United States is seldom in a united state.")
toks <- tokens(txt, remove_punct = TRUE)

dict <- dictionary(list(Countries = c("United States", "Federal Republic of Germany"),
                        Cceans = c("Atlantic Ocean", "Pacific Ocean")), tolower = FALSE)

tokens_lookup(toks, dict)
#> Tokens consisting of 2 documents.
#> d1 :
#> [1] "Countries" "Cceans"    "Cceans"   
#> 
#> d2 :
#> [1] "Countries"
```

To use `tokens_replace()`, we need to know all the matches beforehand. It is not possible when patterns are glob.
```r
# fixed dictionary

pat <- unlist(dict, use.names = FALSE)
rep <- paste0(pat, "/" ,rep(names(dict), lengths(dict)))
tokens_replace(toks, phrase(pat), rep)
#> Tokens consisting of 2 documents.
#> d1 :
#>  [1] "The"                     "United States/Countries"
#>  [3] "is"                      "bordered"               
#>  [5] "by"                      "the"                    
#>  [7] "Atlantic Ocean/Cceans"   "and"                    
#>  [9] "the"                     "Pacific Ocean/Cceans"   
#> 
#> d2 :
#>  [1] "The"                     "Supreme"                
#>  [3] "Court"                   "of"                     
#>  [5] "the"                     "United States/Countries"
#>  [7] "is"                      "seldom"                 
#>  [9] "in"                      "a"                      
#> [11] "united"                  "state"

# glob dictionary

dict2 <- dictionary(list(Countries = c("* States", "Federal Republic of *"),
                         Cceans = c("* Ocean")), tolower = FALSE)

pat2 <- unlist(dict2, use.names = FALSE)
rep2 <- paste0(pat2, "/" ,rep(names(dict2), lengths(dict2)))
tokens_replace(toks, phrase(pat2), rep2)
#> Tokens consisting of 2 documents.
#> d1 :
[...]

PR #2327, opened 2023-12-14
Add concatenator to tokens()

koheiw:
Add concatenator to `tokens()` for #2324. If `tokens.tokens()` applied, concatenator is updated to a different character.
``` r
require(quanteda)
#> Package version: 4.0.0
#> Unicode version: 15.1
#> ICU version: 74.1
txt <- c(d1 = "The United States is bordered by the Atlantic Ocean and the Pacific Ocean.",
         d2 = "The Supreme Court of the United States is seldom in a united state.")
toks <- tokens(txt, remove_punct = TRUE, concatenator = " ")
quanteda:::get_concatenator(toks)
#> [1] " "
toks <- tokens(toks, concatenator = "+")
#> Warning: concatenator changed from " " to "+"
quanteda:::get_concatenator(toks)
#> [1] "+"
```

<sup>Created on 2023-12-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>


koheiw:
This is probably a better behavior. Not only update the meta field but also replace character in types.

``` r
require(quanteda)
#> Package version: 4.0.0
#> Unicode version: 15.1
#> ICU version: 74.1
txt <- c(d1 = "The United States is bordered by the Atlantic Ocean and the Pacific Ocean.",
         d2 = "The Supreme Court of the United States is seldom in a united state.")
toks <- tokens(txt, remove_punct = TRUE, concatenator = " ")

dict <- dictionary(list(Countries = c("* States", "Federal Republic of *"),
                        Oceans = c("* Ocean")), tolower = FALSE)

toks <- tokens_compound(toks, dict)
quanteda:::get_concatenator(toks)
#> [1] "_"
print(toks)
#> Tokens consisting of 2 documents.
#> d1 :
#>  [1] "The"            "United_States"  "is"             "bordered"      
#>  [5] "by"             "the"            "Atlantic_Ocean" "and"           
#>  [9] "the"            "Pacific_Ocean" 
#> 
#> d2 :
#>  [1] "The"           "Supreme"       "Court"         "of"           
#>  [5] "the"           "United_States" "is"            "seldom"       
#>  [9] "in"            "a"             "united"        "state"

toks <- tokens(toks, concatenator = "+")
quanteda:::get_concatenator(toks)
#> [1] "+"
print(toks)
#> Tokens consisting of 2 documents.
#> d1 :
#>  [1] "The"            "United+States"  "is"             "bordered"      
#>  [5] "by"             "the"            "Atlantic+Ocean" "and"           
#>  [9] "the"            "Pacific+Ocean" 
#> 
#> d2 :
#>  [1] "The"           "Supreme"       "Court"         "of"           
#>  [5] "the"           "United+States" "is"            "seldom"       
#>  [9] "in"            "a"             "united"        "state"
```

<sup>Created on 2023-12-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

koheiw:
If it makes sense to use `tokens()` to set or update the concatenator, we should deprecate `concatenator` or change its default value to`NULL` in other functions.

kbenoit:
OK so setting the concatenator on `tokens.character()` sets this attribute but it will be reset by any function that specifies this argument, including by their default values, such as `tokens_compound()` in the example above. So the only functionality difference is to allow `tokens.tokens()` to change the concatenator?

Would it be simpler then to create an accessor and replacement function set called `concatenator()` and `concatenator<-()`?

koheiw:
I thought about making acceessor but `tokens.tokens` is more than that. It replaces existing concatenator to new one to avoid mix of different concatenators in the same object.

koheiw:
I added `concatenator()` but `concatenator<-()` because tampering with `concatenator` causes serious problems. 

In this example, I want to make "異常気象" but it becomes "異常_気象" because of the default value is the tokens constructor is "_". This is strange.
``` r
library(quanteda)
#> Package version: 4.0.0
#> Unicode version: 15.1
#> ICU version: 74.1

dict <- dictionary(list("weather" = "異常 気象")) # means "unusual weather"

toks <- tokens("異常気象")
toks_comp <- tokens_compound(toks, dict, concatenator = concatenator(toks))
print(toks_comp) # looks strage 
#> Tokens consisting of 1 document.
#> text1 :
#> [1] "異常_気象"
tokens_lookup(toks_comp, dict) # matches
#> Tokens consisting of 1 document.
#> text1 :
#> [1] "weather"
```
If I modify "concatenator" in the tokens object, `tokens_lookup()` does not work as expected. This is really wrong.
```r
# tamper with concatenator
quanteda:::set_concatenator(toks_comp) <- ""
tokens_lookup(toks_comp, dict) # no match
#> Tokens consisting of 1 document.
#> text1 :
#> character(0)
```
The best thing to do is set the concatenator when tokens objects are constructed. The results are all good.
```r
toks2 <- tokens("異常気象", concatenator = "")
toks_comp2 <- tokens_compound(toks2, dict, concatenator = concatenator(toks2))
print(toks_comp2) # looks good 
#> Tokens consisting of 1 document.
#> text1 :
#> [1] "異常気象"
tokens_lookup(toks_comp2, dict) # matches
#> Tokens consisting of 1 document.
#> text1 :
#> [1] "weather"
```

It works well if I create a new tokens object with a different concatenator.
```r
toks3 <- tokens(toks, concatenator = "|")
toks_comp3 <- tokens_compound(toks3, dict, concatenator = concatenator(toks3))
print(toks_comp3) # looks good 
#> Tokens consisting of 1 document.
#> text1 :
#> [1] "異常|気象"
tokens_lookup(toks_comp3, dict) # matches
#> Tokens consisting of 1 document.
#> text1 :
#> [1] "weather"
```

kbenoit:
I see your point and how this is especially important for Japanese and other languages without whitespace delimiters between words. 

I'm still a bit unclear myself on how this works, however, since the setting at the tokens creation stage does not carry over into downstream operations such as `tokens_compound()`, `tokens_ngrams()`, or `tokens_lookup()`. Should we change the defaults to look like what you have added above then, for example with the default for `tokens_compound()` set as I did [above](https://github.com/quanteda/quanteda/pull/2327#issuecomment-1855799852) and how you used it in the example above too?

Can you also describe the behaviour in the `tokens()` argument for `concatenator`, preferably in an expanded section in Details?

R/tokens.R
```diff
@@ -334,6 +343,7 @@ tokens.corpus <- function(x,
         types = NULL, 
         what = what,
         tokenizer = tokenizer,
+        concatenator = concatenator,
```
koheiw, line 346: `tokens.tokens(concatenator = "_")` is saved in the object meta and used in the downstream. This is the main change.

R/tokens.R
```diff
@@ -449,6 +462,14 @@ tokens.tokens_xptr <-  function(x,
     if (!include_docvars)
         docvars(x) <- NULL
     
+    if (!identical(get_concatenator(x), concatenator)) {
+        #warning('concatenator changed from "', 
+        #        get_concatenator(x), '" to "', concatenator, '"', call. = FALSE)
+        set_types(x) <- stri_replace_all_fixed(get_types(x), get_concatenator(x),
```
koheiw, line 468: Substitute the old concatenator in types with new one and updated the object meta.

R/tokens.R
koheiw: The concatenator in the object meta matters a lot for `tokens_lookup()` because phrases are automatically concatenated. 

https://github.com/quanteda/quanteda/blob/d1a73cd31635c835d92a7320903903f236bde09f/R/tokens_lookup.R#L154-L155

koheiw:
> Should we change the defaults to look like what you have added above then, for example with the default for `tokens_compound()` set as I did [above](https://github.com/quanteda/quanteda/pull/2327#issuecomment-1855799852) and how you used it in the example above too?

I think all the `concatenator` should default to `concatenator(x)`.

R/tokens.R
```diff
@@ -727,3 +749,18 @@ types.tokens <- function(x) {
     set_types(x) <- value
 }
 
+#' @rdname types
```
kbenoit, line 752: This needs more documentation to describe concatenator. As part of explaining this as requested below, could you flesh this out more here?

kbenoit: We could make this a separate function and describe the behaviour in that Rd entry, rather than in Details in `tokens()`.

kbenoit:
> > Should we change the defaults to look like what you have added above then, for example with the default for `tokens_compound()` set as I did [above](https://github.com/quanteda/quanteda/pull/2327#issuecomment-1855799852) and how you used it in the example above too?
> 
> I think `concatenator` in all the functions but `tokens()` and `as.tokens()` should default to `concatenator(x)`.

I tried this but it created the following error (with the default being `concatenator = concatenator(x)` in the signature of `tokens_compound()`):

```r
> tokens("New York City is located in the United States.") |>
+     tokens_compound(pattern = phrase(c("New York City", "United States")))
Error in check_null(x, arg) : 
  promise already under evaluation: recursive default argument reference or earlier problems?
7.
check_null(x, arg) at validator.R#105
6.
check_character(concatenator) at tokens_compound.R#97
5.
tokens_compound.tokens_xptr(as.tokens_xptr(x), ...) at tokens_compound.R#74
4.
tokens_compound(as.tokens_xptr(x), ...) at tokens_compound.R#118
3.
as.tokens(tokens_compound(as.tokens_xptr(x), ...)) at tokens_compound.R#118
2.
tokens_compound.tokens(tokens("New York City is located in the United States."),
pattern = phrase(c("New York City", "United States"))) at tokens_compound.R#74
1.
tokens_compound(tokens("New York City is located in the United States."),
pattern = phrase(c("New York City", "United States")))
```

koheiw:
We need to rename `concatenator()` to `concat()` or something, because the argument and the accessor cannot be the same name.
See [SO](https://stackoverflow.com/questions/4357101/promise-already-under-evaluation-recursive-default-argument-reference-or-earlie) post.

``` r
x <- list(a = 10, b = 3)

get_b <- function(x) {x$b}
fun1 <- function(x, b = get_b(x)) {
    return(x$a * b)
}
fun1(x)
#> [1] 30

b <- function(x) {x$b}
fun2 <- function(x, b = b(x)) {
    return(x$a * b)    
}
fun2(x)
#> Error in fun2(x): promise already under evaluation: recursive default argument reference or earlier problems?
```

kbenoit:
`concatenator_get()` ?

R/tokens.R
kbenoit: See [bc0b4ca](https://github.com/quanteda/quanteda/pull/2327/commits/bc0b4ca5510d06868b5b5dc54c942565808f4e52) where I implemented this.

koheiw:
If we don't want an alias like `concat()`, we should set the default value `NULL` and extract the concatenator internally.

kbenoit:
I'm fine with having the two functions `concat()` and `concat()` as they are.

kbenoit (approved):

Issue #2329, opened 2023-12-15
Review verbose behaviours


kbenoit:
Some functions don't even have the `verbose` argument, e.g. `tokens_split()`. We should review the functions to see which need it, and review the behaviours when `verbose = TRUE` to make sure they are consistent.

Not sure anyone uses it much, but for the sake of consistency it's still worth reviewing.

koheiw:
We tried to make verbose messages more consistent using `message_select()`. We could do something similar across methods for corpus, tokens and dfm objects.

https://github.com/quanteda/quanteda/blob/84ecce80ff0e840c2896a1ca921e477e616ca103/R/message.R#L56-L81

However, it is not easy to provide detailed information on the operations that will be performed in C++. For example, the message below says removing more features than actually exist. Since 20215 is only the possible sequence of tokens that the pattern would match (and be removed), we only know actual number of tokens removed (4584) only after the operation. 

``` r
require(quanteda)
toks <- tokens(data_corpus_inaugural, remove_numbers = TRUE)
length(types(toks))
#> [1] 10090

toks2 <- tokens_remove(toks, phrase("a *"), verbose = TRUE)
#> removed 20,180 features
#> 
length(types(toks2))
#> [1] 9942

sum(ntoken(toks)) - sum(ntoken(toks2))
#> [1] 4584
```

Further, the repeated use of `types()` is not a good idea because it triggers recompilation of `tokens_xptr`, reducing the new objects' performance gain.

The best approach would be to simplify the message including only the number of documents (and/or tokens), the type of operation (remove/keep, lookup, ngrams etc) and, maybe, a main parameter (e.g. `pattern`, `dictionary`, `n`). This make it easy to create a fit-for-all messaging function easier too.

koheiw:
How about making `message_tokens()` and `message_dfm()` that can be used in all the methods?

``` r
require(quanteda)
#> Loading required package: quanteda
#> Package version: 4.0.0
#> Unicode version: 15.1
#> ICU version: 74.1
#> Parallel computing: 16 of 16 threads used.
#> See https://quanteda.io for tutorials and examples.
toks <- tokens(data_corpus_inaugural, remove_numbers = TRUE)

stats_tokens <- function(x) {
    list(ndoc = ndoc(x),
         ntoken = sum(ntoken(x, remove_padding = TRUE)))
}

message_tokens <- function(operation, pre, post) {
[...]

PR #2336, opened 2024-01-04
Add remove_padding to ntoken()

koheiw:
If `ntoken(x, remove_padding = TRUE)`, it returns the number of tokens ignoring paddings. The results are the same as `ntoken(tokens_remove(x, pattern = ""))` but much more efficient. This will be useful in showing the number of tokens removed in verbose messagess (#2329).

I kept `...` in `ntoken()` but I think the argument should be deprecated as it does not work in `ntoken.dfm()`. User should just run `ntoken(tokens(x))` if necessary. The documentation also says "... not used".

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,8 @@
 
 * Added new arguments `append_key`, `separator` and `concatenator` to `tokens_lookup()`. These allow tokens matched by dictionary values to be retained with their keys appended to them, separated by `separator`.  The addition of the `concatenator` argument allows additional control at the lookup stage for tokens that will be concatenated from having matched multi-word dictionary values. (#2324)
 
+* Added a new argument `remove_padding` to `ntoken()` that allows for not counting padding that might have been left over from `tokens_remove(x, padding = TRUE`). (#2336)
+
 ## Removals
 
 * `bootstrap_dfm()` was removed for character and corpus objects.  The correct way to bootstrap sentences is not to tokenize them as sentences and then bootstrap them from the dfm.  This is consistent with requiring the user to tokenise objects prior to forming dfms or other "downstream" objects.
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -89,8 +89,8 @@ cpp_ndoc <- function(xptr) {
     .Call(`_quanteda_cpp_ndoc`, xptr)
 }
 
-cpp_ntoken <- function(xptr) {
-    .Call(`_quanteda_cpp_ntoken`, xptr)
+cpp_ntoken <- function(xptr, padding = TRUE) {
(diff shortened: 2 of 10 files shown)


koheiw:
We can compute how many tokens were removed easily even when `padding = TRUE`.
``` r
require(quanteda)
#> Package version: 4.0.0
#> Unicode version: 13.0
#> ICU version: 69.1
toks <- tokens(data_corpus_inaugural, remove_numbers = TRUE)

toks2 <- tokens_remove(toks, phrase("a *"), verbose = TRUE, padding = TRUE)
#> removed 20,180 features
#> 

sum(ntoken(toks)) - sum(ntoken(toks2))
#> [1] 0
sum(ntoken(toks)) - sum(ntoken(toks2, remove_padding = TRUE))
#> [1] 4584
```

kbenoit:
Nice addition. 

On removing `...`: Right now the `...` in `ntoken()` and `ntype()` for `char` and `corpus` are already deprecated because those entire functions are. Calling it on-the-fly in `ntoken.tokens()` is not nearly as expensive since it's a removal only, not re-tokenising. And keeping the default `remove` options via `...` would be consistent with adding the new `remove_padding` option. Just as `remove_padding` adds the capability of computing these quickly if a count is desired, the other remove options provide the chance to do this if a user wants to compare the token counts with and without punctuation, for instance.

kbenoit (approved):
Should we add this to `ntype()` for consistency? (It's only a 1 token difference, but still, why not be consistent?)

koheiw:
> Should we add this to `ntype()` for consistency? (It's only a 1 token difference, but still, why not be consistent?)

We can but we probably need to add `types(remove_padding = TRUE)` if we do so. The default value will be different between `ntype()` and `ntoken()` 🤔

kbenoit:
Thinking about this, `remove_padding = FALSE` ought to be the default for both. If a dfm creates the pad as a "feature" by default, then it should be considered a token/type by default. So I'd say include it for both and make the default false for both.

Might even make sense then to include it in the methods for dfms too then.

kbenoit:
This is more consistent now, since formerly, it was:

```r
> tokens("a b c") |>
+     tokens_remove("b", padding = TRUE) |>
+     ntype()
text1 
    2 
> tokens("a b c") |>
+     tokens_remove("b", padding = TRUE) |>
+     ntoken()
text1 
    3 
```

Now it's 3 for both.

And for dfms created from this, it's also consistent, although we don't have a removal option.
```r
> tokens("a b c") |>
+     tokens_remove("b", padding = TRUE) |>
+     dfm() |>
+     ntoken()
text1 
    3 
> tokens("a b c") |>
+     tokens_remove("b", padding = TRUE) |>
+     dfm() |>
+     ntoken(remove_padding = TRUE)
text1 
    3 
```

PR #2337, opened 2024-01-09
Upgrade lookup

koheiw:
Add `modify_if` to `tokens_lookup()`, but its behavior in the exclusive mode is a bit different in that "text2" is still modified (but matches are not recorded).

``` r
require(quanteda)
#> Package version: 4.0.0
#> Unicode version: 15.1
#> ICU version: 74.1
dat <- data.frame(text = c("R and C are languages",
                           "Windows (R), Quanteda (C)"),
                  topic = c("language", "software"))
dict <- dictionary(list(language = c("C", "R")))
corp <- corpus(dat)
toks <- tokens(corp)

tokens_lookup(toks, dict, modify_if = toks$topic == "language")
#> Tokens consisting of 2 documents and 1 docvar.
#> text1 :
#> [1] "language" "language"
#> 
#> text2 :
#> character(0)
tokens_lookup(toks, dict, modify_if = toks$topic == "language", exclusive = FALSE)
#> Tokens consisting of 2 documents and 1 docvar.
#> text1 :
#> [1] "LANGUAGE"  "and"       "LANGUAGE"  "are"       "languages"
#> 
#> text2 :
#> [1] "Windows"  "("        "R"        ")"        ","        "Quanteda" "("       
#> [8] "C"        ")"
```


kbenoit:
I like this extension, just wonder if we should call it something else. It's modified pretty fundamentally if `exclusive = TRUE`. At the extreme, if the condition were not met by any document, we would be returning an empty list of length `ndoc`. 

Maybe call this `apply_if` or `lookup_if`, and and add a note in details about how this works and what we recommend using it for? I'd prefer `apply_if` since that might be something we use elsewhere.

koheiw:
I am happy with `apply_if`. It can used in other functions where `modify_if` is.

kbenoit:
Things to do:
- [ ] add a note to Details about why empty documents are returned for documents not matching the condition, when `exclusive = TRUE`
- [ ] change to `apply_if`
- [ ] add to `dfm_lookup()`

koheiw:
`R-CMD-check / ubuntu-latest (devel) (pull_request) ` is failing because of the warning from Rcpp.

```
Found the following significant warnings:
  /home/runner/work/_temp/Library/Rcpp/include/Rcpp/iostream/Rstreambuf.h:53:20: warning: field precision specifier ‘.*’ expects argument of type ‘int’, but argument 2 has type ‘std::streamsize’ {aka ‘long int’} [-Wformat=]
  /home/runner/work/_temp/Library/Rcpp/include/Rcpp/iostream/Rstreambuf.h:57:21: warning: field precision specifier ‘.*’ expects argument of type ‘int’, but argument 2 has type ‘std::streamsize’ {aka ‘long int’} [-Wformat=]
  /home/runner/work/_temp/Library/Rcpp/include/Rcpp/print.h:30:19: warning: format not a string literal and no format arguments [-Wformat-security]
```

kbenoit (approved):

kbenoit:
Looks like something we cannot fix then. So merge?

PR #2338, opened 2024-01-12
Update quickstart

koheiw:
There are many defunct functions. I don't know if these were causing errors on CRAN though.


kbenoit (approved):
A much-needed refurb of the ageing documentation. Thanks Kohei!

Issue #2339, opened 2024-01-13
Make cpp_kwic()


koheiw:
I should make C++ function to produce concordance views for KWIC. Currently, these commands are failing in Rmd probably because repeated deep-copying of the very long document (whole book).

https://github.com/quanteda/quanteda/blob/02c79e556c9d6dbf819d8fbd5bcbfbf482958a77/vignettes/pkgdown/replication/digital-humanities.Rmd#L122-L123

kbenoit:
Interesting... is it the second one with the glob? or the fact that it's being run twice?

PR #2340, opened 2024-01-14
Dev cpp kwic

koheiw:
#2339

### DIFF
--- a/R/RcppExports.R
+++ b/R/RcppExports.R
@@ -9,6 +9,10 @@ cpp_index <- function(xptr, words_, thread = -1L) {
     .Call(`_quanteda_cpp_index`, xptr, words_, thread)
 }
 
+cpp_kwic <- function(xptr, documents_, pos_from_, pos_to_, window, delim_ = " ", thread = -1L) {
+    .Call(`_quanteda_cpp_kwic`, xptr, documents_, pos_from_, pos_to_, window, delim_, thread)
+}
+
 cpp_index_types <- function(patterns_, types_, glob = TRUE) {
     .Call(`_quanteda_cpp_index_types`, patterns_, types_, glob)
 }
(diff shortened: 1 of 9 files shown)


kbenoit (approved):
Approved, although I prefer to keep things in R only unless C++ really improves performance.

koheiw:
The C++ version is a lot faster. This is also a preparation to port the package from R to other languages. We must rely on other package less because they are not available in other languages.
```
> require(quanteda)
Loading required package: quanteda
Package version: 4.0.0
Unicode version: 15.1
ICU version: 74.1
Parallel computing: 16 of 16 threads used.
> corp <- readRDS("C:/Users/watan/Dropbox/Public/data_corpus_guardian2016-10k.rds")
> toks <- tokens(corp)
> index <- index(toks, "econom*")
+   kwic(toks, index = index),
+   times = 10
+ )
Unit: milliseconds
 kwic(toks, index = index) 209.9983 230.5482 253.8923 257.2767 270.6691 298.5482    10
 
 
> require(quanteda)
Loading required package: quanteda
Package version: 3.3.1
Unicode version: 15.1
ICU version: 74.1
Parallel computing: 16 of 16 threads used.
Warning message:
In .recacheSubclasses(def@className, def, env) :
  undefined subclass "ndiMatrix" of class "replValueSp"; definition not updated
> corp <- readRDS("C:/Users/watan/Dropbox/Public/data_corpus_guardian2016-10k.rds")
> toks <- tokens(corp)
> index <- index(toks, "econom*")
+   kwic(toks, index = index),
+   times = 10
+ )
Unit: milliseconds
 kwic(toks, index = index) 714.4407 732.9401 847.8043 847.6982 880.2857 1077.305    10
```

kbenoit:
Fair enough, performance is the key motivator here, and you're right, this will make porting to e.g. Python easier, if that should ever happen.

PR #2341, opened 2024-01-18
Fix cpp

koheiw:
Address potential stability problem. `pkgdown::build_article("pkgdown/replication/digital-humanities_test")` still doesn't work, but the following works fine.

```r
require(quanteda)
load("vignettes/pkgdown/replication/data_char_mobydick.rda")
quanteda_options(threads = 8)
for (i in seq(100)) {
  cat(i, "\n")
  xtoks <- toks <- tokens(data_char_mobydick, xptr = TRUE)
  out <- index(xtoks, pattern = "whale*")
}

for (i in seq(100)) {
  cat(i, "\n")
  xtoks <- toks <- tokens(data_char_mobydick, xptr = TRUE)
  type <- quanteda:::cpp_get_types(xtoks)
  out <- quanteda:::index_types("whale*", type, "glob")
}
```


koheiw:
@kbenoit Let's merge this PR. If we need to change more, I will issue an another PR.

kbenoit (approved):

PR #2347, opened 2024-02-21
Make as.dictionary() generic

koheiw:
Change `as.dictionary()` to pass additional arguments. I want to make this function to work: https://github.com/koheiw/newsmap/pull/74/commits/34cbdc082756735f339bcc98ddb74f56573a5652


kbenoit (approved):
Looks fine. 

Should we add a `as.dictionary.list()` method that wraps `dictionary.list()`? Seems odd that we cannot coerce an ordinary named list into a dictionary.

koheiw:
Might be other way around: `as.dicitonary.list()` is called when `x` is a list in `dictionary()`.

PR #2348, opened 2024-02-26
Address winbuilder issues

koheiw:


kbenoit:
Great that you are addressing this since it's currently holding up a (re)submission to CRAN from late last week. 

Is this envisioned as a temporary fix? Should we not add some note on how to set the options to enable TBB to work?
Or am I missing something? The startup message indicates that multi-threading is disabled but the threads option query says otherwise.

```r
> library(quanteda)
Package version: 4.0.0
Unicode version: 14.0
ICU version: 71.1
Parallel computing: disabled
> quanteda_options("threads")
[1] 10
```

koheiw:
The serial code is used only when TBB is not available. I suspected that winbuilder uses the serial code but probably not because it does not build without this patch. 

We should add Github actions that run tests disabling TBB. The serial code will not be used on modern platforms but help us to separate multi-threading issues.

koheiw:
I can install the winbuilder binary `install.packages("https://win-builder.r-project.org/cow0DAW5f98q/quanteda_4.0.0.zip")` and run tests successfully. We might be better off if we upload the Windows binary to Github Releases and forget about CRAN (or submit serial version only).

kbenoit:
What if instead of skipping tests, we just  

```r 
if (identical(unname(Sys.info()["sysname"]), "Windows")) {
    quanteda_options(threads = 1) 
    # or in C++:
    #define QUANTEDA_USE_TBB false 
}
```
and re-enable multi-threading in C++ when and we or another package fixes this?

kbenoit (approved):

PR #2358, opened 2024-03-27
Fix tokens restore

koheiw:
`match("\uE001", letters)` returns `NA` when the special character is not found. This leads to error in conversion to `unsingned int` in C++.


kbenoit:
Our tests are [too thorough](https://www.businessinsider.com/trump-stop-coronavirus-testing-right-now-have-very-few-cases-2020-6)! 

Thanks @koheiw.

kbenoit (approved):

PR #2361, opened 2024-04-03
Include tbb

koheiw:
Trying to find a way to bundle TBB, but not sure how to yet 😞


kbenoit:
Ugh, I tried working with this myself, looks like this PR is trying to eliminate RcppParallel and the requirement to install TBB locally by bundling it. But so far it only works on Windows. And the way that RcppParallel includes the source and links to the appropriate files is very different. Are you sure this is the best approach?

Is there no chance of making this work with just RppParallel which is using [TBB 2019 Update 8](https://github.com/RcppCore/RcppParallel/blob/master/src/tbb/README.md) ?

PR #2362, opened 2024-04-03
Remove quanteda extras

kbenoit:
Simplify package dependences and remove any dependency, even weak ones, on the other quanteda packages.

This prevents needing to build **quanteda.textstats**, **quanteda.textplots**, and **quanteda.textmodels** during CRAN checks or CI, simplifying package management and making updates easier.

All packages used only in tests and with `skip_if_not_installed()` are moved to `Enhances:` in `DESCRIPTION`. I think this will avoid having to build them, but no one seems sure!


kbenoit:
@koheiw i think this is ready to merge and (re)submit. What do you think?

koheiw (approved):

PR #2363, opened 2024-04-04
Update pkgdown

koheiw:
Added commands to install the TBB library.

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -4,6 +4,7 @@ output: github_document
 
 ```{r, echo = FALSE}
 knitr::opts_chunk$set(
+ warning = FALSE,
  collapse = TRUE,
  comment = "##",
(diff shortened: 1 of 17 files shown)


stefan-mueller:
I just saw this now. Could we be more specific about which binary package Windows and Mac users should download to use Intel oneAPI Threading Building Blocks? The README currently states, “Windows and Mac users can download a binary package from CRAN”, but I cannot find the name of that package in the documentation.

kbenoit:
Good point. It means that if you install from CRAN, you don't need to worry about TBB, since the object files are already compiled in the binary. Only Linux users or those trying to compile it need to install TBB on their Windows or Mac machines. But you're right, we could add more documentation about how to install TBB for those compiling locally from source.

PR #2365, opened 2024-04-05
Update installation instructions

kbenoit:
Following @stefan-mueller's [comment](https://github.com/quanteda/quanteda/pull/2363#issuecomment-2039270393) on https://github.com/quanteda/quanteda/pull/2363, I think we need to be more specific on the requirements for building the package from source. 

This applies to all Linux users as well as to anyone else building from source.

(@koheiw did I get the Windows instructions right, in that TBB is included in RTools?)


stefan-mueller (approved):
Thanks for improving the instructions, @kbenoit. I just tried to add one more sentence which states very clearly that Windows and macOS users do not need additional software when installing **quanteda** from CRAN. Feel free to merge the PR after reviewing this sentence.

kbenoit:
Unfortunately I still get 

```
   checking for tbb/tbb.h... no
   configure: WARNING: parallel computing is disabled because tbb/tbb.h (Intel TBB) is not found
```

when compiling on macOS, even with TBB installed, but I will raise that as a separate issue.

kbenoit:
@koheiw are you happy with the accuracy of these updated instructions?

koheiw:
I don't know how to install TBB on MacOS, but looks good otherwise.

koheiw (approved):

PR #2368, opened 2024-04-05
Test for absence of libtbb in configure.ac

rsbivand:
This seems to work when tbb-devel is present or absent, setting or not setting the Makvars variables.

### DIFF
--- a/configure
+++ b/configure
@@ -1569,6 +1569,53 @@ printf "%s\n" "$ac_res" >&6; }
   eval $as_lineno_stack; ${as_lineno_stack:+:} unset as_lineno
 
 } # ac_fn_cxx_check_header_compile
+
+# ac_fn_cxx_try_link LINENO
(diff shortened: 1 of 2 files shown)


rsbivand:
Yes, and @koheiw just got an issue with the new `proxyC` version failing to install when libtbb is not present, for the same reason as here. I feel that TBB cannot be considered useful if a standard AC_CHECK_HEADER test passes when the install platform does not have the tbb-devel package installed. The proposed change here exercises both the header and the library, so fails correctly.

koheiw:
Thanks you very much! You saved a lot of my time with your expertise. I will merge first to [fix-configure](https://github.com/quanteda/quanteda/tree/fix-configure) and change mini TBB program a bit; then to master.

barracuda156:
Maybe it is possible to add a configure argument to specify install location of OneTBB?

Issue #2364, opened 2024-04-05
Install failure Fedora 39 g++ 13.2.1 quanteda 4.0.0; configure.ac error


rsbivand:
The CRAN Fedora test result for g++ 13.2.0 https://cran.r-project.org/web/checks/check_results_quanteda.html also found for 13.2.1. Your configure test passes when `tbb-devel` is absent. When `tbb-devel` is installed, package installation succeeds. First the error without `tbb-devel` :

```
trying URL 'https://cloud.r-project.org/src/contrib/quanteda_4.0.0.tar.gz'
Content type 'application/x-gzip' length 4095204 bytes (3.9 MB)
==================================================
downloaded 3.9 MB

* installing *source* package 'quanteda' ...
** package 'quanteda' successfully unpacked and MD5 sums checked
** using staged installation
checking whether the C++ compiler works... yes
checking for C++ compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether the compiler supports GNU C++... yes
checking whether g++ -std=gnu++17 accepts -g... yes
checking for g++ -std=gnu++17 option to enable C++11 features... none needed
checking how to run the C++ preprocessor... g++ -std=gnu++17 -E
checking for gcc... gcc
checking whether the compiler supports GNU C... yes
checking whether gcc accepts -g... yes
checking for gcc option to enable C11 features... none needed
checking whether the compiler supports GNU C++... (cached) yes
checking whether g++ -std=gnu++17 accepts -g... (cached) yes
checking for g++ -std=gnu++17 option to enable C++11 features... (cached) none needed
checking for stdio.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for strings.h... yes
checking for sys/stat.h... yes
checking for sys/types.h... yes
checking for unistd.h... yes
checking for tbb/tbb.h... yes
configure: creating ./config.status
config.status: creating src/Makevars
** libs
using C++ compiler: 'g++ (GCC) 13.2.1 20240316 (Red Hat 13.2.1-7)'
[...]

PR #2371, opened 2024-04-07
add pkg-config to configure.ac

rsbivand:
Re: https://github.com/quanteda/quanteda/issues/2364#issuecomment-2041454789, this PR adds support for `pkg-config` to `configure.ac`. If `pkg-config` is found and `tbb` is installed, it now proceeds to the viability test. If either `pkg-config` is not found, or `pkg-config` finds no `tbb`, no viability test is used and `tbb` is assumed not to be present.


koheiw (approved):
Looks great!

PR #2373, opened 2024-04-09
Address the CRAN issue, support macOS

kbenoit:
Attempting to:
* address the CRAN comments from #2372 
* resolve the macOS not finding TBB issue (#2367)

Issue #2372, opened 2024-04-09
configure script could use tidying up


kbenoit:
Some notes from the CRAN publication of [v4.0.1](https://github.com/quanteda/quanteda/releases/tag/v4.0.1):

```
checking Intel TBB existence:... yes
./configure: line 4315: -I/usr/local/share/pkgconfig/../../include: No such file or directory
checking tbb available for compiling and linking:... no

configure: Package LIBS: -L/usr/local/share/pkgconfig/../../lib/intel64/gcc4.8 -L/usr/local/share/pkgconfig/../../lib -ltbb
```
so

a) the output of

```
pkg-config --exists tbb
```
should be captured/suppressed.

b) pkg-config --exists does not really do what it says (e.g. it requires all the Requires, even Requires.private ones for dynamic linking), and in this case it seems to have found detritus left by Intel MKL with nonsense entries.

(c) `pkg-config --libs` usually gives paths for linking not loading (no -rpath or similar), so a loading test is essential.

(d) `configure` should allow the detection to be overridden.

kbenoit:
@rsbivand You can see some of my attempts to address these in the closed PR #2373 (closed since I should not have been modifying `configure`). 

Making this issue a checklist:

- [ ] the output of `pkg-config --exists tbb` should be captured/suppressed.

     **solved** in https://github.com/quanteda/quanteda/blob/8d0179b370252504b835e75b7dbf3ae27c0bd0ae/configure.ac#L29

- [ ] `pkg-config` --exists does not really do what it says (e.g. it requires all the Requires, even Requires.private ones for dynamic linking), and in this case it seems to have found detritus left by Intel MKL with nonsense entries.

- [ ] `pkg-config --libs` usually gives paths for linking not loading (no -rpath or similar), so a loading test is essential.

- [ ] `configure` should allow the detection to be overridden.

PR #2374, opened 2024-04-09
Fix TBB on macOS w/homebrew

### LINKED ISSUE #2367: Homebrew installation of TBB still not found on macOS when compiling
When compiling from source, even after 

```
brew install TBB
```

I get
```
   checking for tbb/tbb.h... no
   configure: WARNING: parallel computing is disabled because tbb/tbb.h (Intel TBB) is not found
```

hence
```
> library("quanteda")
Package version: 4.0.0
Unicode version: 14.0
ICU version: 71.1
Parallel computing: disabled
```

kbenoit:
Helps the configure file locate TBB on macOS 

- Works on M1 (where homebrew is in /opt/homebrew)
- Works on x86 (where homebrew is in /usr/local/Homebrew)

Closes #2367.


rsbivand:
@kbenoit  Re: https://github.com/quanteda/quanteda/pull/2373#issuecomment-2045790167 This looks promising, and probably addresses a) and b). c) might be accomplished by executing `libtbb_test` if the output of compiling and linking is `yes`, and only choosing $? 0 for `yes`, which would mean that tbb is fully viable. d) would involve a configure argument such as `--with-tbb`, default `yes`, but with `no` falling through to not using tbb at all even if present and usable. We use configure args in `sf`'s configure.ac. Will you try this or should I - tomorrow a bit tight for time?

rsbivand:
If the homebrew so0lution was to use `PKG_CONFIG_PATH`, I guess you'll update the docs to show how to update the environment variable, for example in Readme.md but also elsewhere. `Sys.setenv` should be feasible, but maybe needs `Sys.getenv` first to prepend the new path to any existing paths.

kbenoit:
Actually the script finds something set by homebrew's installation, without my having set any sort of environment variable. It worked for two different platforms with two different locations of homebrew.

I think the point d) of #2372 was that we should allow external environment variables to override the defaults. But the current script works fine with whatever settings homebrew's general installation is setting up. (All I can see are links to /opt/homebrew/bin and /opt/homebrew/sbin in my $PATH.

```r
> sum(stringi::stri_detect_regex(Sys.getenv(), "tbb"))
[1] 0
> sum(stringi::stri_detect_regex(names(Sys.getenv()), "tbb"))
[1] 0
> Sys.getenv()[["PATH"]]
[1] "/Users/kbenoit/.virtualenvs/r-spacyr/bin:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/opt/X11/bin:/Library/Apple/usr/bin:/usr/local/MacGPG2/bin:/Library/TeX/texbin:/Users/kbenoit/.cargo/bin:/usr/ucb:/Users/kbenoit/Applications/quarto/bin:/usr/texbin:/Applications/RStudio.app/Contents/Resources/app/quarto/bin:/Applications/RStudio.app/Contents/Resources/app/bin/postback"

```

kbenoit:
> @kbenoit Re: [#2373 (comment)](https://github.com/quanteda/quanteda/pull/2373#issuecomment-2045790167) This looks promising, and probably addresses a) and b); with your edits CMD check only has a couple of notes with and without tbb installed. c) might be accomplished by executing `libtbb_test` if the output of compiling and linking is `yes`, and only choosing $? 0 for `yes`, which would mean that tbb is fully viable. d) would involve a configure argument such as `--with-tbb`, default `yes`, but with `no` falling through to not using tbb at all even if present and usable. We use configure args in `sf`'s configure.ac. Will you try this or should I - tomorrow a bit tight for time?

I'd be very grateful if you gave it a try, but no rush.

koheiw (approved):
I added a test to check if TBB is enabled [0b9fbe0](https://github.com/quanteda/quanteda/pull/2374/commits/0b9fbe061d25ea77949f433bfda9b44294297947). This PR passes the test!

kbenoit:
When I remove TBB (`brew uninstall tbb`) and then recompile, I see

```
   configure: WARNING: Intel TBB not installed; install TBB devel package for parallel processing or update PKG_CONFIG_PATH environment variable```
but I can't find any mention in the code (`configure.ac` or anywhere else) that uses that environment variable. Should we be searching for tbb using that path in the configuration files, in case a user has installed it somewhere non-standard?

**Comment by kbenoit** [comment]
Without tbb, the test from https://github.com/quanteda/quanteda/pull/2374/commits/0b9fbe061d25ea77949f433bfda9b44294297947 fails:

```
✖ | 2        0 | tbb                                                                                   
───────────────────────────────────────────────────────────────────────────────────────────────────────
Failure (test-tbb.R:6:3): TBB is enabled
info$enabled is not TRUE

`actual`:   FALSE
`expected`: TRUE 

Failure (test-tbb.R:7:3): TBB is enabled
info$max_threads > 1 is not TRUE

`actual`:   FALSE
`expected`: TRUE 
```

and a WARNING in 
```
✔ |   1  1  18 | quanteda_options                                                                      
───────────────────────────────────────────────────────────────────────────────────────────────────────
Warning (test-quanteda_options.R:72:5): quanteda_options works with threads
Setting threads instead to maximum available 1
Backtrace:
    ▆
 1. └─quanteda::quanteda_options(threads = 2) at test-quanteda_options.R:72:5
 2.   └─quanteda:::set_option_value(key, args[[key]]) at quanteda/R/quanteda_options.R:104:13
 3.     └─quanteda:::check_threads(value) at quanteda/R/quanteda_options.R:135:5
```

In #2372 based in BDR's CRAN submission comments, the comment
> (c) pkg-config --libs usually gives paths for linking not loading (no -rpath or similar), so a loading test is essential.
meant a loading test in `configure.ac`.

rsbivand:
1) @kbenoit `PKG_CONFIG_PATH` is used by `pkg-config`, see https://linux.die.net/man/1/pkg-config. I have `export PKG_CONFIG_PATH=/usr/local/lib64/pkgconfig:/usr/local/lib/pkgconfig` in `.bashrc` on Linux to add the non-standard locations of `*.pc` files to the default search path. Users should set the environment variable before beginning installation, so that `configure`'s use of `pkg-config` picks it up.

2) I'll get back to updating `configure.ac` for the load test and a configure argument.

3) The problem with the test in https://github.com/quanteda/quanteda/pull/2374#issuecomment-2046642907 seems to be that the test ought first to see if tbb is enabled, not expecting that it is, and only if it is enabled test the threads.

rsbivand:
@kbenoit @koheiw Please do let me know where your various commits are. They are not in the master branch, not in the fix-2373 branch. For now, I have no coherent clone of your repo on which to act. In order to create a PR, I need to have a clone that is current with your latest commits.

kbenoit:
@rsbivand I've just added you to the repo settings with write permissions, so you won't have to fork anything and we can work on any branch fixes more directly.

This PR solves the macOS issue so I'll merge and close it now. The other issues in #2372 should be addressed in a new PR. I don't think there are any commits for that, just discussion happening here (but referenced in the issue).

rsbivand:
@kbenoit Thanks for the invitation, but that isn't a good idea. I only need to know what is the current code branch containing the test-tbb file and the current version of configure.ac. I cannot commit to offering assistance going forward beyond fixing the immediate problem. I only found the installation problem because my installation has almost 3000 installed packages, needed to run reverse dependency checks on geographical/spatial packages for which I have been responsible; I don't use quanteda, although the proximity measures in proxyC are in some way tangential to measures of proximity in geographical space.

Issue #2399, opened 2024-07-19
Keep original unigrams in tokens_compound()


koheiw:
It is possible to keep unigrams in `tokens_ngrams()` by setting `n = 1` but it is not in `tokens_compound()`.  Keeping unigrams makes it easy to apply pattern matching in the downstream. How about adding an option like `keep_unigram`?

kbenoit:
OK, good idea. `keep_unigrams = FALSE` (plural), or
```
action = c("replace", "add")
```
but the `action` would only make sense if there were a possible third option, and I cannot think of any.

PR #2402, opened 2024-07-25
Issue 2399

koheiw:
Add `keep_unigrams` as discussed in #2399. Compared to other arguments of the function like `join` the name is rather long, so it could be just `keep`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: quanteda
-Version: 4.0.2
+Version: 4.0.3
 Title: Quantitative Analysis of Textual Data
 Description: A fast, flexible, and comprehensive framework for 
     quantitative text analysis in R.  Provides functionality for corpus management,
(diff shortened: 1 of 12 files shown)


koheiw:
```
> txt2 <- "we like high quality sound"
> toks2 <- tokens(txt2)
> seqs2 <- phrase(c("high quality", "quality sound"))
> 
> tokens_compound(toks2, seqs2, join = TRUE, keep_unigrams = TRUE)
text1 :
[1] "we"                 "like"               "high"               "quality"            "sound"             
[6] "high_quality_sound"

> tokens_compound(toks2, seqs2, join = FALSE, keep_unigrams = TRUE)
text1 :
[1] "we"            "like"          "high"          "high_quality"  "quality"       "quality_sound"
[7] "sound"        
```

kbenoit (approved):

koheiw:
I wasn't entirely sure about the order of original and compounded tokens. Do you think it is OK as is?

kbenoit:
I thought about that too... It doesn't matter a great deal, but it would probably make sense to insert the compounds only after the unigrams.

So above, the first example is fine, but the second example ought to be:

```r
> tokens_compound(toks2, seqs2, join = FALSE, keep_unigrams = TRUE)
text1 :
[1] "we"            "like"          "high"          "quality"       "sound"    "high_quality" 
[7] "quality_sound"
```

koheiw:
I think I can change to this way. I will issue a separate PR.

kbenoit:
Then this PR seems ready for merge. 🛫

Issue #2406, opened 2024-08-08
CRAN problems: UBSAN


kbenoit:
From: Prof Brian Ripley <ripley@stats.ox.ac.uk>
Date: Monday, 5 August 2024 at 17:27
To: Benoit,KR <K.R.Benoit@lse.ac.uk>
Cc: CRAN@R-project.org <CRAN@R-project.org>
Subject: CRAN package quanteda
Dear maintainer,

Please see the problems shown on
<https://cran.r-project.org/web/checks/check_results_quanteda.html>.

Please correct before 2024-08-19 to safely retain your package on CRAN.

Do remember to look at the 'Additional issues'.

The CRAN Team

## Issue

See [clang-UBSAN](https://www.stats.ox.ac.uk/pub/bdr/memtests/clang-UBSAN/quanteda)

Looks like:

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /data/gannet/ripley/R/test-clang/Rcpp/include/Rcpp/proxy/AttributeProxy.h:74:32 
/data/gannet/ripley/R/test-clang/Rcpp/include/Rcpp/proxy/AttributeProxy.h:74:32: runtime error: reference binding to address 0x7fff06ce7a50 with insufficient space for an object of type 'Rcpp::Vector<13>'
0x7fff06ce7a50: note: pointer points here
 ed 7f 00 00  18 77 d9 de 72 55 00 00  88 a7 d9 de 72 55 00 00  50 7a ce 06 ff 7f 00 00  0a 74 79 70
              ^ 
    #0 0x7fed20c79d41 in Rcpp::AttributeProxyPolicy<Rcpp::Vector<13, Rcpp::PreserveStorage> >::attr(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) /data/gannet/ripley/R/test-clang/Rcpp/include/Rcpp/proxy/AttributeProxy.h:74:16
    #1 0x7fed20c79d41 in cpp_as_list(Rcpp::XPtr<TokensObj, Rcpp::PreserveStorage, &(void Rcpp::standard_delete_finalizer<TokensObj>(TokensObj*)), false>) /data/gannet/ripley/R/packages/tests-clang-UBSAN/quanteda/src/tokens_xptr.cpp:36:12

[...]

PR #2410, opened 2024-08-15
Use C++ iterator

koheiw:
This is the only thing I can fix for #2406.


kbenoit (approved):
I tried to check the UBSAN behaviour but it's hard to know exactly. I'll try later on r-hub.

PR #2417, opened 2024-08-25
Remove armadillo

koheiw:
We are not using RcppArmadillo anymore.


kbenoit:
I wondered if we still needed it. Good riddance.

kbenoit (approved):

Issue #2428, opened 2024-11-12
R crash if ntype is used on tokens with empty documents


koheiw:
Sadly, it is easy to replicate.
```
require(quanteda)
toks <- tokens(c("a b c", ""))
ntype(toks)
```

PR #2429, opened 2024-11-12
Fix ntype.tokens()

koheiw:
Address #2428


kbenoit (approved):

kbenoit:
Crazy how such a simple, but fatal edge case can elude detection for so long. Great that you caught this and we now test for it.

PR #2431, opened 2024-12-04
Implement issue 2430

kbenoit:
Implements new URL regex and tries tests ... which are currently failing for extended URLs.


koheiw:
Actually we need to change this patter:
https://github.com/quanteda/quanteda/blob/97878413ddf1d9b6e806f8e3adc6b32f7bacf82a/R/tokenizers.R#L162

koheiw (approved):
Please merge if it looks good for you.

PR #2443, opened 2025-03-01
Add corpus_chunk()

kbenoit:
We need a simple equivalent for a corpus object of `tokens_chunk()`. The main use case will be chunking documents before passing to LLMs that have fixed token sizes. It will be very convenient to be able to recombine the documents to the original, pre-chunk version after combining LLM returns. In the extension package [**quanteda.llm**](https://github.com/quanteda/quanteda.llm), we are designing ways to round-trip corpus data to LLMs, returning the objects as docvars or some other attached data. With `corpus_chunk()`, we can replace attempts to do this outside of **quanteda** such as https://github.com/quanteda/quanteda.llm/blob/main/R/text_split.R.

Of course `corpus_chunk()` could also be useful for many other purposes.

I created an argument here called `inflation_factor` (not a great name) since most **quanteda** tokens will be just 0.8 of an LLM token, as they use various subword tokenisation schemes.

Additional arguments could include the splitting unit (or tokenisation unit) since a smarter version would not split sentences (or paragraphs).


koheiw:
The size should be specified by the number of characters because corpus is a set of characters. This must be a lot faster too.

``` r
require(stringi)
txt <- stri_wrap(quanteda::data_char_ukimmig2010[1], 512)
head(txt)
#> [1] "IMMIGRATION: AN UNPARALLELED CRISIS WHICH ONLY THE BNP CAN SOLVE. - At current immigration and birth rates, indigenous British people are set to become a minority well within 50 years. - This will result in the extinction of the British people, culture, heritage and identity. - The BNP will take all steps necessary to halt and reverse this process. - These steps will include a halt to all further immigration, the deportation of all illegal immigrants, a halt to the \"asylum\" swindle and the promotion of"   
#> [2] "the already existing voluntary repatriation scheme. - The BNP recognises the right of legally settled and law-abiding minorities to remain in the UK and enjoy the full protection of the law, on the understanding that the indigenous population of Britain has the right to remain the majority population of our nation. - The BNP will deport all foreigners convicted of crimes in Britain, regardless of their immigration status. - The BNP will review all citizenship grants awarded since Labour entered government"  
#> [3] "in 1997, based on that party's admission that they orchestrated mass immigration to change forcibly Britain's demographics and to gerrymander elections. - The BNP will repeal the Race Relations Act and all other far leftist social engineering projects, such as the Equalities and Human Rights Commission aimed at enforcing multiculturalism. - The BNP will regain sovereign control of our borders by increasing funding to the UK Border Agency. The Scale of the Crisis Britain's existence is in grave peril,"       
#> [4] "threatened by immigration and multiculturalism. In the absence of urgent action, we, the indigenous British people, will be reduced to minority status in our own ancestral homeland within two generations. We, alone of all the political parties, have a decades-long record of truthfully articulating our concerns, often with foresight, and long before many of the problems associated with multiculturalism entered the public consciousness. British People Set to be a Minority within 30 - 50 Years: The Official"   
#> [5] "Figures - According to the Office for National Statistics (ONS), legal Third World immigrants made up 14.7 percent (7.5 million) of the population of England in 2004 (\"Population review of 2004 and 2005: England and Wales\", Karen Gask, Office for National Statistics). - According to the ONS, 10.9 percent (or 6.7 million) of the currently resident population of Britain was not born in this country (\"One in 10 of population born abroad\", Telegraph, 09 Dec 2009). - In 2005, the ONS issued a separate report"
#> [6] "which said that 36 percent of all births in England and Wales were not \"white British\" (\"Birthweight and gestational age by ethnic group, England and Wales 2005: introducing new data on births\", Kath Moser, Office for National Statistics). - This 2005 birth rate figure does not include births to second and third generation immigrant mothers. Figures released by the ONS in January 2009 revealed that the Muslim population in Britain has grown by more than 500,000 to 2.4 million in just four years. Their"
stri_length(txt)
#>  [1] 507 510 505 509 508 506 509 511 509 504 507 503 511 509 508 511 509 511 507
#> [20] 509 508 509 504 508 505 509 507 507 511 511 509 505 507 504 509 510 138
```
Then, it should be something like
```r
corups_chunk <- function(x, size = 512) {
  attrs <- attributes(x)
  lis <- lapply(x, stri_wrap, width = size)
  attrs[["docvars"]] <- reshape_docvars(attrs[["docvars"]],
                                        rep(seq_along(lis), lengths(lis)), 
                                        drop_docid = FALSE)
  result <- rebuild_corpus(unlist(lis, use.names = FALSE), attrs)
  return(result)
}
```
See https://github.com/quanteda/quanteda/blob/bef32cb2ca47d0db9d0d63abfde649f24c91b5e2/R/tokens_chunk.R#L56-L69

koheiw:
Relatedly but separately, I sometimes convert loss-less tokens to characters to save texts in a data.frame by `lapply(toks, paste, collapese = " ")`. It might be nice to have `as.character.tokens()` to do this.

kbenoit:
What would `as.character.tokens()` produce? We already have these:

```r
> toks <- tokens(c("a a b d e", "a a b c d e"))
> toks
Tokens consisting of 2 documents.
text1 :
[1] "a" "a" "b" "d" "e"

text2 :
[1] "a" "a" "b" "c" "d" "e"

> as.list(toks)
$text1
[1] "a" "a" "b" "d" "e"

$text2
[1] "a" "a" "b" "c" "d" "e"

> as.character(toks)
 [1] "a" "a" "b" "d" "e" "a" "a" "b" "c" "d" "e"
```

koheiw:
```r
c("a a b d e", "a a b c d e")
```
It is like `corpus.tokens()`, but I am reluctant to make such a method.

kbenoit:
I understand, it's not something to encourage for part of a workflow (to go "upstream") but for dealing with LLM context windows, we do need some way to a) split a corpus with roughly controlled lengths and b) to know, approximately at least, how many tokens this will be counted from these chunks by the LLMs.  

Character splitting is the fastest method but it can results in words split in half.  Chunking on sentences would be avoid that and might make a sensible default option.

Agree though that this ought to be fast(er).  Compared to the speed of passing the chunks through LLMs, however, chunking time is not currently very significant.

koheiw:
The great thing about `stri_wrap()` is that it only split texts only the boundaries, so words are not cut into half.

kbenoit:
You're right - I meant sentences being cut in half. It's not an unheard of way to chunk for LLMs, but some form of semantic unit splitting would be better. See https://js.langchain.com/v0.1/docs/modules/data_connection/document_transformers/.

koheiw:
That makes sense. We already have `corpus_reshape()` to split long texts into sentences, but sentences can be longer than 512 characters or a lot shorter than that. What do you want to do in those cases?

kbenoit:
I was thinking that we would need chunk delimiter options similar to langchain's document splitters. So splitting on words using `stri_wrap()` makes sense. Or, linking these to tokens (`what`) which includes sentences already, but could also be extended to LLM-style tokenizers such as `transformers::BertTokenizer`, since it's those lengths that will be used when it comes to LLM model limits.

kbenoit:
With sentences longer than 512 tokens, we would need to issue a warning and/or force sentence splitting, possibly as an option. The default for short context window LLMs like BERT is just to ignore the extra tokens. So a non-semantic split is unlikely to be worse than that.

kbenoit:
@koheiw What changes should we make to pass this? How about: An option for where to split (words or sentences) and for words, using `stringi::stri_wrap()`. For sentences where the tokens exceed a limit (like 512 for BERT), we have an option for dropping to splitting on words. (This will be an option, since BERT and similar models can take longer sentences but just truncate what exceeds the token limit.)

koheiw:
I am not entirely sure what you want, but how about this? If `drop = TRUE`, texts longer than `width` are truncated.

``` r
require(quanteda)
#> Package version: 4.2.1
#> Unicode version: 15.1
#> ICU version: 74.1

corups_chunk <- function(x, size = 512, drop = FALSE) {
    attrs <- attributes(x)
    lis <- lapply(x, function(v) {
        if (drop) {
            head(stringi::stri_wrap(v, width = size), 1)
        } else {
            stringi::stri_wrap(v, width = size)
        }
    })
    attrs[["docvars"]] <- quanteda:::reshape_docvars(attrs[["docvars"]],
                                                     rep(seq_along(lis), lengths(lis)), 
                                                     drop_docid = FALSE)
    result <- quanteda:::rebuild_corpus(unlist(lis, use.names = FALSE), attrs)
    return(result)
}

txt <- "We shall permit genuine students, so long as they possess health insurance. This latter concession will also assist with our long-term policy of re-building Third World economies in order to facilitate the voluntary return of their nationals or descendants, under our long-term resettlement programme."

corpus_reshape(corpus(txt)) %>% 
    corups_chunk(size = 100, drop = FALSE)
#> Corpus consisting of 4 documents.
#> text1.1 :
#> "We shall permit genuine students, so long as they possess he..."
#> 
#> text1.2 :
#> "This latter concession will also assist with our long-term p..."
#> 
#> text1.3 :
#> "economies in order to facilitate the voluntary return of the..."
#> 
#> text1.4 :
#> "long-term resettlement programme."

corpus_reshape(corpus(data_char_ukimmig2010)[1]) %>% 
    corups_chunk(size = 100, drop = TRUE)
#> Corpus consisting of 88 documents.
#> BNP.1 :
#> "IMMIGRATION: AN UNPARALLELED CRISIS WHICH ONLY THE BNP CAN S..."
#> 
#> BNP.2 :
#> "The Scale of the Crisis Britain's existence is in grave peri..."
#> 
#> BNP.3 :
#> "In the absence of urgent action, we, the indigenous British ..."
#> 
#> BNP.4 :
#> "We, alone of all the political parties, have a decades-long ..."
#> 
#> BNP.5 :
#> "British People Set to be a Minority within 30 - 50 Years: Th..."
#> 
#> BNP.6 :
#> "Figures released by the ONS in January 2009 revealed that th..."
#> 
#> [ reached max_ndoc ... 82 more documents ]
```

<sup>Created on 2025-04-19 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

kbenoit:
That's what I was thinking. I'll work with that and push the new code. We leave sentence-based chunking for the next version.

kbenoit:
@kohei I implemented a few versions but the only one that accomplishes the token-based chunking needed for feeding corpora to LLMs is the tokens approach. The principal objection currently is that it's slow, since it round-trips to tokens and back to a corpus. But the behaviour is what is called for. 

I've marked this function as "experimental" using *lifecycle*. Are you ok with this for now, and we find ways to optimise performance later? or some link from our tokens to LLM-friendly tokens, so that we can use `tokens_chunk()`?

koheiw:
Can you explain why `stri_wrap()` did not work? It should be more or less be the same as concatenating tokens.

If tokeniziation is necessary, users should apply `tokens_chunk()` and convert to a character vector using a new version of `as.character.tokens()`. In this way, we can avoid conversions in reverse the direction (i.e. from tokens to corpus).

kbenoit:
The problem with `stri_wrap()` is that `width` is expressed in terms of code points (characters), not tokens. The chunking needs to break the text into segments of tokens that are below a particular model's token limit (which will be the linguistically/quanteda defined tokens plus an inflation factor, due to how LLMs tokenise subwords).

koheiw:
Have you considered adjusting the `width` with the average token length?

``` r
require(stringi)
txt <- "We shall permit genuine students, so long as they possess health insurance. This latter concession will also assist with our long-term policy of re-building Third World economies in order to facilitate the voluntary return of their nationals or descendants, under our long-term resettlement programme."

size <- 10 # 10 tokens
e <- stri_length(txt) / stri_count_boundaries(txt)
stri_wrap(txt, width = size * e)
#> [1] "We shall permit genuine students, so long as they possess"     
#> [2] "health insurance. This latter concession will also assist with"
#> [3] "our long-term policy of re-building Third World economies in"  
#> [4] "order to facilitate the voluntary return of their nationals or"
#> [5] "descendants, under our long-term resettlement programme."
```

<sup>Created on 2025-04-30 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

koheiw (approved):

koheiw:
Approved but I don't know if we really need `inflation_factor` because users can simply set smaller `size` if sub-word tokenization will be performed. You also don't like the name much. It is a lot harder to remove or change an argument than to add one after the release.

kbenoit:
You're right! I removed this and added an explanation in the details. Take a final look and see if you agree with it.

koheiw:
Looks a lot better!

PR #2444, opened 2025-03-07
Reduce memory usage

koheiw:
Reduce memory usage by destroying intermediate objects.


kbenoit (approved):

kbenoit (changes requested):
Same issue as https://github.com/quanteda/quanteda/pull/2445#pullrequestreview-2762876521

```r
── Failure ('test-collocations.R:207:5'): char_select works with collocations pattern ──
char_keep(txt, patt) not identical to c(c1 = "aa bb", c3 = "aa bc").
Lengths differ: 0 is not 2
── Failure ('test-collocations.R:211:5'): char_select works with collocations pattern ──
char_remove(txt, patt) not identical to c(c2 = "ab", c4 = "bcd", c5 = "bcd").
Lengths differ: 5 is not 3
```

PR #2454, opened 2025-04-21
Add apply_if tokens_ngrams()

koheiw:
#2390


koheiw:
One line is not covered by the tests but it shouldn't be a problem.

kbenoit (approved):
🙏🙏🙏

PR #2423, opened 2024-09-01
Fix UBSAN issue

Enchufa2:
Closes #2406. I still think that this is a sanitizer issue, but if we can avoid confusing it, you don't have to argue with CRAN. 

Here `ListOf` (`Tokens`) is a very particular class. But given that `as_list` returns a `List` in the first place, there is no gain in transforming it into a `ListOf`. So keeping everything as a `List` seems to content UBSAN. Using the `rocker/r-devel-ubsan-clang` container, I see:

```bash
$ ASAN_OPTIONS=detect_leaks=0 RDscript -e 'remotes::install_github("Enchufa2/quanteda", ref="test")'
$ RDscript -e 'quanteda::as.tokens(list("test"))'
Tokens consisting of 1 document.
text1 :
[1] "test"
```

PR #2456, opened 2025-05-13
R 4.5 fixes

kbenoit:
Tidies up a few things such as a change in test behaviour related to one straight v curly quote issue, fixes a few URLs, increments the quanteda version, and attempts a fix at the clang-UBSAN issues currently listed on CRAN. @koheiw please review the latter especially.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,5 +1,5 @@
 Package: quanteda
-Version: 4.2.1
+Version: 4.3.0
 Title: Quantitative Analysis of Textual Data
 Description: A fast, flexible, and comprehensive framework for 
     quantitative text analysis in R.  Provides functionality for corpus management,
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -23,7 +23,7 @@ Authors@R: c(
     )
 License: GPL-3
 Depends:
-    R (>= 3.5.0),
+    R (>= 4.1.0),
     methods
 Imports:
     fastmatch,
--- a/src/tokens_xptr.cpp
+++ b/src/tokens_xptr.cpp
@@ -11,9 +11,9 @@ TokensPtr cpp_xptr() {
 }
 
 // [[Rcpp::export]]
-TokensPtr cpp_as_xptr(const List text_, 
+TokensPtr cpp_as_xptr(const List text_,
                       const CharacterVector types_) {
-    
+
     Texts texts = Rcpp::as<Texts>(text_);
     Types types = Rcpp::as<Types>(types_);
     TokensObj *ptr = new TokensObj(texts, types);
@@ -22,8 +22,8 @@ TokensPtr cpp_as_xptr(const List text_,
 
 // [[Rcpp::export]]
 TokensPtr cpp_copy_xptr(TokensPtr xptr) {
-    TokensObj *ptr_copy = new TokensObj(xptr->texts, 
-                                        xptr->types, 
+    TokensObj *ptr_copy = new TokensObj(xptr->texts,
+                                        xptr->types,
                                         xptr->recompiled,
                                         xptr->padded);
     return TokensPtr(ptr_copy, true);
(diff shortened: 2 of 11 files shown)


kbenoit:
BTW: GitHub now suggesting Copilot as a possible reviewer for PRs!!!

DESCRIPTION
```diff
@@ -23,7 +23,7 @@ Authors@R: c(
     )
 License: GPL-3
 Depends:
-    R (>= 3.5.0),
```
koheiw, line 26: I wonder if it is really necessary to change to R (>= 4.1.0).

src/tokens_xptr.cpp
```diff
@@ -39,16 +39,16 @@ List cpp_get_attributes(TokensPtr xptr) {
 // [[Rcpp::export]]
 List cpp_as_tokens(TokensPtr xptr) {
     xptr->recompile();
-    List texts_ = as_list(xptr->texts);
```
koheiw, line 42: The UBSAN is https://github.com/quanteda/quanteda/pull/2423 again.

`wrap()` is used to convert native C++ objects to Rcpp objects. `as_list()` returns a list, so its redundant. The same for `ecnode()`.

DESCRIPTION
kbenoit: The NOTE at current https://cran.r-project.org/web/checks/check_results_quanteda.html indicates:

```
Result: NOTE 
    Missing dependency on R >= 4.1.0 because package code uses the pipe
    |> or function shorthand \(...) syntax added in R 4.1.0.
    File(s) using such syntax:
      ‘char_select.R’ ‘char_select.Rd’ ‘convert-wrappers.Rd’ ‘convert.Rd’
      ‘corpus-addsummary-metadata.R’ ‘corpus_sample.Rd’ ‘dfm.Rd’
      ‘dfm_select.Rd’ ‘dfm_tfidf.Rd’ ‘dictionary.Rd’ ‘docfreq.Rd’
      ‘docnames.Rd’ ‘index.Rd’ ‘nsentence.Rd’ ‘tokens_compound.Rd’
      ‘tokens_replace.Rd’ ‘tokens_sample.Rd’ ‘tokens_split.Rd’
      ‘topfeatures.Rd’
```
so we would need to change these to the (older) **magrittr* `%` if we want to make it compatible. Even then, I am not sure all of the dependencies would work with that now-ancient R version.

src/tokens_xptr.cpp
kbenoit: This was my attempt to resolve the clang-UBSAN issues flagged at https://www.stats.ox.ac.uk/pub/bdr/memtests/clang-UBSAN/quanteda/tests/testthat.Rout. We need to resolve these before resubmitting, and I thought the wrapper would ensure that the type is strictly defined and thus avoids the UB.

The UBSAN warning:

`reference binding to address … with insufficient space for an object of type ‘Rcpp::Vector<13>’`

seems to points to a problem in the code assigns attributes:

`texts_.attr("types") = encode(xptr->types);`

which assumes that texts_ is a proper R object that supports attributes, and that `Rcpp::Vector<13>` — which corresponds to `Rcpp::CharacterVector` — can be safely assigned via `attr()`.

The issue seems to be that `texts_` is the result of:

`Tokens texts_ = as_list(xptr->texts);`

Here, `Tokens` is likely an alias for a List or List-like class, but if `as_list(...)` does not return a full-fledged R object (i.e., a SEXP that can accept attributes), then assigning attributes to it via `.attr()` can trigger undefined behaviour. So the "redundant" code attempts to guarantee that the object being set via `.attr(...)` on is a List, i.e. a VECSXP. If as_list(...) returns something more minimal (e.g. raw pointers or non-alloc’d objects), it may cause UBSAN to complain.

But @koheiw you're the C++ expert, I'm just drawing on AI assistance and my vague memory of C++!

koheiw: `encode()` should return a `CharacterVector`.

https://github.com/quanteda/quanteda/blob/ef07e43b4f644c331a298370feee3cbe1f560e87/inst/include/lib.h#L179-L187

How about this?

```cpp
String type_ = wrap(types[i]);
```

`as_list()` should return a `List`.

https://github.com/quanteda/quanteda/blob/ef07e43b4f644c331a298370feee3cbe1f560e87/inst/include/lib.h#L196-L204

This might avoid confusing UBSAN.
```
List texts_ = as_list(xptr->texts);
```

kbenoit:
Sure, push that and we’ll try our luck!

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

DESCRIPTION
koheiw: I don't think `|>` is a strong case but your are probably right that dependencies do not work on R 3.5.

koheiw (approved):

Issue #2458, opened 2025-05-18
fcm() might be wrong


koheiw:
We saw instances of unexpected results from `fcm()`. Related to #1413. 

``` r
require(quanteda)
#> Loading required package: quanteda
#> Package version: 4.3.0
#> Unicode version: 15.1
#> ICU version: 74.1
#> Parallel computing: 8 of 8 threads used.
#> See https://quanteda.io for tutorials and examples.
toks <- tokens(c("a b", "a b", "b a"))

# should be [b, a] = 1
fcm(toks, ordered = TRUE, tri = TRUE)
#> Feature co-occurrence matrix of: 2 by 2 features.
#>         features
#> features a b
#>        a 0 3
#>        b 3 0

# should be [b, a] = 1
fcm(toks, ordered = TRUE, tri = FALSE)
#> Feature co-occurrence matrix of: 2 by 2 features.
#>         features
#> features a b
#>        a 0 3
#>        b 3 0

# looks fine
fcm(toks, ordered = FALSE, tri = FALSE)
#> Feature co-occurrence matrix of: 2 by 2 features.
#>         features
#> features a b
#>        a 0 3
#>        b 3 0

# looks fine but should be [b, a] = 0 i s misleading.
fcm(toks, ordered = FALSE, tri = TRUE)
#> Feature co-occurrence matrix of: 2 by 2 features.
#>         features
#> features a b
#>        a 0 3
#>        b 0 0
```

<sup>Created on 2025-05-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

koheiw:
The interaction between arguments `fcm()` difficult to use.

One option is dedicate `fcm.tokens()` to `context = window` and `ordered = TRUE` (therefore `tri = FALSE` always) and `fcm.dfm()` to `context = document` (therefore `ordered = FALSE` and `tri = TRUE` always`.  

Another option is print warnings when `context = window and ordered = TRUE` or  `ordered = TRUE & tri = TRUE)`.

kbenoit:
The first example is correct, although I can see how it would be confusing, because `context = "document"` and the explanation for the `ordered` argument states that 'Only makes sense for context = "window"'.
```r
# should be [b, a] = 1
fcm(toks, ordered = TRUE, tri = TRUE)
#> Feature co-occurrence matrix of: 2 by 2 features.
#>         features
#> features a b
#>        a 0 3
#>        b 3 0
```

[...]

PR #2459, opened 2025-05-18
Fix 2458

koheiw:
For #2458. We can get correct values by not calling `fcm.dfm()` when `context = documents`.

``` r
require(quanteda)
#> Package version: 4.2.1
#> Unicode version: 15.1
#> ICU version: 74.1
toks <- tokens(c("a b", "a b", "b a"))

fcm(toks, ordered = TRUE, tri = TRUE)
#> Warning in fcm.tokens_xptr(as.tokens_xptr(x), ...): cannot return only upper
#> triangle when ordered = TRUE
#> Feature co-occurrence matrix of: 2 by 2 features.
#>         features
#> features a b
#>        a 0 2
#>        b 1 0
fcm(toks, ordered = TRUE, tri = FALSE)
#> Feature co-occurrence matrix of: 2 by 2 features.
#>         features
#> features a b
#>        a 0 2
#>        b 1 0
fcm(toks, ordered = FALSE, tri = FALSE)
#> Feature co-occurrence matrix of: 2 by 2 features.
#>         features
#> features a b
#>        a 0 3
#>        b 3 0
fcm(toks, ordered = FALSE, tri = TRUE)
#> Feature co-occurrence matrix of: 2 by 2 features.
#>         features
#> features a b
#>        a 0 3
#>        b 0 0
```

<sup>Created on 2025-05-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


R/fcm.R
```diff
@@ -210,41 +210,43 @@ fcm.tokens_xptr <- function(x, context = c("document", "window"),
...
-            if (!is.null(weights)) {
-                weights <- check_double(weights, max_len = Inf)
-                if (length(weights) != window)
-                    stop("The length of weights must be equal to the window size")
-            } else {
-                weights <- 1 / seq_len(window)
-            }
+    }
+    if (context == "document")
+        window <- max(ntoken(x))
```
kbenoit, line 219: See my comment [here](https://github.com/quanteda/quanteda/issues/2458#issuecomment-2893300183). Setting   this to the max length is not the same as context = "document" since it still allows different behaviour when `ordered = TRUE`. The idea is not just one of length, it's one of document cooccurrence that has no concept of order. That's why in my suggestion, we remove context = "document" entirely from tokens, and suggest that users create a dfm if that's what they want.

If we want to provide an option as in this line, we could either provide `-1` to mean `max(ntoken(x))` or we could just suggest that users input this themselves for window or some other effectively equivalently large value.

koheiw: Let's remove `context == "document"`. This might be even more confusing.

PR #2461, opened 2025-05-19
Explicitly set threads to 2 (kwic)

kbenoit:
To fix this NOTE:

Check: examples, Result: NOTE
  Examples with CPU time > 2.5 times elapsed time
        user system elapsed ratio
  kwic 0.624  0.466   0.297  3.67


koheiw:
How about use `donttest{}` instead?

kbenoit:
OK, let's try that. If you approve and merge, I'll resubmit to CRAN today.

koheiw (approved):

koheiw:
@kbenoit thanks🤞

PR #2478, opened 2026-01-30
Add drop0.dfm method for dfm hard-thresholding

mahjabinoyshi:
This PR adds a drop0() generic and a drop0.dfm() method to perform hard thresholding on dfm objects.

* New generic:
  - drop0(x, ...) dispatches on the class of x.

* New method:
  - drop0.dfm(x, tol = 0, ...) zeros out dfm cell values with absolute value < tol using Matrix::drop0().
  - Preserves docvars, docnames, and featnames by saving and restoring metadata around the Matrix-based operation.
  - Keeps the dfm dimensions unchanged (only increases sparsity).

* Tests:
  - tests/testthat/test-drop0.R verifies:
    - values below tol are set to zero,
    - docvars are preserved,
    - dim, docnames, and featnames are preserved,
    - tol = 0 leaves the dfm unchanged.

All new tests pass locally with testthat.


koheiw:
There are not files called 'Untitled.R' but it is in NAMESPACE. Use **roxygen2** to update NAMESPACE instead of editing manually.

R/drop0.R
```diff
@@ -0,0 +1,58 @@
...
+    sx <- methods::as(x, "dgCMatrix")
+
+    # apply Matrix::drop0 to threshold values
+    sx <- Matrix::drop0(sx, tol = tol, ...)
+
+    # rebuild dfm using build_dfm(), like dfm_compress()
+    result <- build_dfm(
+        sx,
+        colnames(sx),
+        unit    = "documents",
```
koheiw: New DFM should inherit all the attributes of the original DFM including the "unit". To copy all the attributes, `rebuild_dfm()` might be easier. See:

https://github.com/quanteda/quanteda/blob/d83caf473ed8c7cf1766c8576e5e9daa6cdece9e/R/dfm_match.R#L58

Relatedly we need to test if all the attributes are unchanged in `test-drop0.R`.

R/drop0.R
```diff
@@ -0,0 +1,58 @@
...
+#' @param ... additional arguments passed to \code{Matrix::drop0()}.
+#'
+#' @return A dfm with small cell values set to zero, with docvars preserved.
+#'
+#' @details
+#' This performs "hard thresholding" (matrix sparsification) on the dfm:
+#' it increases the number of zero cells but does not change the dfm
+#' dimensions. For removing features entirely, use \code{\link{dfm_trim}}.
+#'
+#' @export
```
kbenoit: Suggest you use @keywords internal here, since this is more of a developer's function. We can add a note to `dfm_trim()` that this function from the **Matrix** package has been extended to preserve dfm attributes. Or @koheiw do you think this belongs in the package index?

DESCRIPTION
```diff
@@ -70,6 +70,7 @@ RoxygenNote: 7.3.3
 Roxygen: list(markdown = TRUE)
 Collate: 
     'RcppExports.R'
+    'Untitled.R'
```
kbenoit: @koheiw is correct, if you build using **roxygen2** then this will naturally disappear.

kbenoit (changes requested):

R/drop0.R
```diff
@@ -0,0 +1,44 @@
...
+#'
+#' @details
+#' This performs "hard thresholding" (matrix sparsification) on the dfm:
+#' it increases the number of zero cells but does not change the dfm
+#' dimensions. For removing features entirely, use \code{\link{dfm_trim}}.
+#'
+#' @keywords internal
+#' @export
+drop0.dfm <- function(x, tol = 0, ...) {
+    x <- quanteda::as.dfm(x)
```
koheiw: Remove `quanteda::`.

koheiw (changes requested):
It is failing tests. Please see my comments.

Issue #2480, opened 2026-02-02
normalize_characters() is not called


koheiw:
For some reason, [quanteda:::normalize_characters()](https://github.com/quanteda/quanteda/blob/d83caf473ed8c7cf1766c8576e5e9daa6cdece9e/R/tokenizers.R#L371) is not called before tokenization. We cannot removed tokens with Unicode quotation mark easily. @kbenoit do you remember any reason not calling the function?

``` r
library(quanteda)
#> Package version: 4.3.1
#> Unicode version: 15.1
#> ICU version: 74.1
#> Parallel computing: 16 of 16 threads used.
#> See https://quanteda.io for tutorials and examples.
txt <- "i’m not happy"
stringi::stri_escape_unicode(txt)
#> [1] "i\\u2019m not happy"
(corp <- corpus(txt))
#> Corpus consisting of 1 document.
#> text1 :
#> "i’m not happy"
(toks <- tokens(corp))
#> Tokens consisting of 1 document.
#> text1 :
#> [1] "i’m"   "not"   "happy"
tokens_remove(toks, stopwords())
#> Tokens consisting of 1 document.
#> text1 :
#> [1] "i’m"   "happy"
```

<sup>Created on 2026-02-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

koheiw:
Test coverage shows that it is really not used anywhere.
https://app.codecov.io/gh/quanteda/quanteda/blob/master/R%2Ftokenizers.R#L373

I am not sure how we should normalize hyphens but quotation marks should be replaced by the ASCII single quote to match the stopwords.

kbenoit:
Normalizing hyphens would not be a bad idea either, although, typographically, there are important differences. The problem is that (until AI started writing text at least) most typists were unaware of the differences between say a hyphen and a hyphen-minus, or an em and en dash, or they were unaware of how to make their keyboards produce them. So I'm agnostic on the hyphens. 

Smart quotes on the other hand are the solution to a purely visual problem and I'm all in favour of replacing them with standard single- and double- straight quotes.

PR #2482, opened 2026-02-08
Use normalize_character()

koheiw:
For #2480. Unlike earlier version this substitutes only subsets of Pi, PF, Pb character class before tokenization. We could also make this optional by adding `tokens(x, normalize = FALSE)`.


kbenoit (changes requested):
I like the change. Hyphens, smart quotes, and EOL are the bane of working with text in applications where the inputs are inevitably a hodgepodge of different formats.

We already intervene discreetly for some of this at the corpus level:

https://github.com/quanteda/quanteda/blob/d1461509e731eec3034fdf9ccb66ceb545dc6214/R/corpus.R#L145-L150

I think the most conservative approach to not change behaviour, however, would be (at least until the next major release) to add this to `tokens()` as `normalize = FALSE`, as you suggest.

PR #2492, opened 2026-04-08
Fix dict separator

koheiw:
Currently, `separator` does not have any effect when dictionary is a YAML file, leading to no matches of multi-word expressions. This is a major bug.


kbenoit (approved):
Please update NEWS.md for this and the other PR

koheiw:
@kbenoit thank you for reviewing. Please update the CRAN whenever earliest.

PR #2495, opened 2026-05-27
Add as.Matrix.tokens()

koheiw:
We added `as.tensor.tokens()` recently but `as.Matrix.tokens()` would be more useful because we cannot easily subset tensor objects. In this example, data loaders take documents from a row-oriented sparse matrix to train a convolutional network.

https://github.com/koheiw/torch-test/blob/master/text-classification_quanteda.R

```r
> toks <- tokens(c(doc1 = "a b c d e f g",
+                  doc2 = "g d e e",
+                  doc3 = ""))
> 
> as.Matrix(toks)
3 x 7 sparse Matrix of class "dgRMatrix"
                  
doc1 1 2 3 4 5 6 7
doc2 7 4 5 5 . . .
doc3 . . . . . . .
```

In order to split data into training and test sets, I also need to add `tokens_subset(recompile = FALSE)` option. Revisiting the old issue https://github.com/quanteda/quanteda/issues/370.


koheiw:
Admittedly, this is an ugly code. 

https://github.com/koheiw/torch-test/blob/581408348ef0f1eaf14431caab54dadd8c360b1c/text-classification_quanteda.R#L37-L40

```r
train_ds <- movie_dataset(tokens_subset(toks, recompile = FALSE, split == "train"), 
                          output_length)
test_ds <- movie_dataset(tokens_subset(toks, recompile = FALSE, split == "test"), 
                         output_length)
```

If we had "document-position matrix" (DPM) or something, I can write a cleaner code. It should be just a dgRMatrix with document variables.

```r
dpmt <- dpm(toks, length = output_length)
train_ds <- movie_dataset(dpm_subset(dpmt, split == "train"))
test_ds <- movie_dataset(dpm_subset(dpmt, split == "test"))
```

koheiw:
Close in favor of #2497

Issue #2436, opened 2025-01-08
Integrate passing corpora to LLMs


kbenoit:
A possible model: https://skllm.beastbyte.ai/

koheiw:
It is hard to maintain API callers, but there is a package that supports many APIs: 

https://github.com/MichelNivard/gptstudio

Issue #1145, opened 2017-12-14
Add convert(x, to = "kerasR") functionality


kbenoit:
Following the discussion in #1138, we are thinking of extending `convert()` so that we could go from a dfm using:

```r
convert(anydfm, to = "kerasR")
```

Problem is we haven't used keras or **kerasR**.

@statsmaths can you tell us what format of object would be the best to provide from **quanteda**? 

Would extending `convert()` be most appropriate? We can also provide conversion from a tokens object, but it seems more natural to convert a dfm.  Would/should it possibly include additional document variables? 

We're ready to implement it but would appreciate the guidance.

statsmaths:
I think that is a great idea! Plus, the exact same output could be used with the **keras** package (R studio's port to make keras more "R like") and my **kerasR** (which follows the Python interface more closely).

Ultimately, keras wants a matrix with one row per document and one column for the maximum length (in tokens) of a given document. As an intermediate step, you'll need a list with a single element per document. Each element contains a vector of (ordered) word indices. Both packages provide `pad_sequences`, which allows for going from the intermediate step to the final version.

To give you a working example, here  is the first step using **tokenizers**:

```{r}
docs <- c("I like R.", "I really like quanteda.", "I love fasttext")
tokens <- tokenizers::tokenize_words(docs)
vocab <- unique(unlist(tokens))
intermediate_results <- sapply(tokens, function(v) match(v, vocab))
intermediate_results
```
```
[[1]]
[1] 1 2 3

[[2]]
[1] 1 4 2 5

[[3]]
[1] 1 6 7
```

Which vocabulary element gets mapped to an index is irrelevant to keras, though I assume users would want to have access to the object `vocab` (it is also needed if someone is applying the same model to a new dataset). Once someone has this list, they can quickly convert to the format needed in keras:

```{r}
keras_input <- kerasR::pad_sequences(intermediate_results)
keras_input
```
```
     [,1] [,2] [,3] [,4]
[1,]    0    1    2    3
[2,]    1    4    2    5
[3,]    0    1    6    7
```

I would recommend returning just `intermediate_results` in **quanteda**. There are some important  options in `kerasR::pad_sequences`/`keras::pad_sequences` that users will likely want to adjust.

I hope that helps describe the output format. If you'd like some help with building the functionality, just let me know.

kbenoit:
[...]

PR #2497, opened 2026-05-30
Add as.matrix.tokens()

koheiw:
## Summary

Adds an `as.matrix()` method for `tokens` objects and re-implements `as.tensor.tokens()` to return a *dense* rather than a sparse tensor. Both are backed by a new C++ routine that operates directly on the `tokens_xptr` external-pointer representation. The PR also adds `docvars()` and `docvars<-()` methods for `tokens_xptr`, so document variables can be read and assigned without round-tripping through a `tokens` object (which forces recompilation).

## Motivation

`as.tensor.tokens()` (introduced in #2495) returned a sparse COO tensor. For the intended use — feeding tokenised documents into torch for batch training — the natural representation is a dense [document × position] matrix of integer token IDs, not a sparse one. Constructing it in R was also inefficient and required materialising the whole corpus. The new design moves construction into C++ and adds an `extract` argument so that torch can pull arbitrary subsets of documents (mini-batches) directly from a `tokens_xptr`, avoiding repeated token recompilation.

## Changes

- **New `as.matrix.tokens()`** (`R/tokens-methods.R`): returns an integer matrix, one row per document and one column per token position; values are token IDs, `0` for padding/empty positions. Arguments:
  - `length` — number of columns; documents are right-padded with `0` or truncated to this length. Defaults to the maximum token count across the selected documents.
  - `extract` — integer document indices to include, in the order given, enabling random-access subsetting and reordering.
  - `drop` — if `TRUE` (default) and the result has a single row, returns a vector.
  - Non-`tokens_xptr` input is coerced via `as.tokens_xptr()` internally.
- **New C++ backend `cpp_as_matrix()`** (`src/tokens_xptr.cpp`, exported through `RcppExports`): recompiles the pointer once, then fills a `G × length` integer matrix for the requested indices, with bounds-checking on `extract`.
- **`as.tensor.tokens()` re-implemented**: now calls `as.matrix()` and wraps the result in `torch::torch_tensor()`. It returns a dense tensor, gains `extract` and `...` (the latter passed to `torch::torch_tensor()`), and shifts token IDs by `+1L` so padding maps to `1` (1-based indexing for torch/R).
- **New `docvars.tokens_xptr()` and `docvars<-.tokens_xptr()`** (`R/docvars.R`): read and assign docvars on the pointer in place; assignment previously went through `as.tokens()` and triggered recompilation. `docvars<-.tokens()` now coerces with `as.tokens()` first.
- **NAMESPACE**: registers `as.matrix.tokens`, `docvars.tokens_xptr`, `docvars<-.tokens_xptr`.
- **Tests**: new `tests/testthat/test-as.matrix.R` (basic conversion, `length`, `extract`, `drop`, empty and single-document cases, and `data_corpus_inaugural`); updates to `test-as.tensor.R`, `test-tokens.R`, `test-tokens_xptr.R`, `test-default-methods.R`.

## Example

```r
library(quanteda)
xtoks <- tokens(c(doc1 = "a b c d e f g",
                  doc2 = "g d e e",
                  doc3 = ""), xptr = TRUE)

as.matrix(xtoks)
#>      [,1] [,2] [,3] [,4] [,5] [,6] [,7]
#> doc1    1    2    3    4    5    6    7
#> doc2    7    4    5    5    0    0    0
#> doc3    0    0    0    0    0    0    0

as.matrix(xtoks, length = 5, extract = c(1, 3))
#>      [,1] [,2] [,3] [,4] [,5]
#> doc1    1    2    3    4    5
#> doc3    0    0    0    0    0
```


koheiw:
The **quanteda** tokens leads to more accurate results than the **tok** tokens. It is also faster than other implementations. Note that I did not include the time for corpus construction because it happens only once.

## tok tokenizer byte-pair encoding 
```
loss: 0.3716
acc: 0.8329
   user  system elapsed 
 460.90  129.22  249.94 
```

## tok tokenizer whitespace

```
loss: 0.3587
acc: 0.8395
   user  system elapsed 
 480.44  133.25  265.23 
```

## quanteda tokens with sparse matrix

```
loss: 0.335
acc: 0.8543
   user  system elapsed 
 696.39   67.28  460.03 
```

## quanteda tokens with dense matrix (this PR)

```
loss: 0.3404
acc: 0.8516
   user  system elapsed 
 519.00   61.58  154.97 
```

koheiw:
I probably should make **quanteda.torch** for this function. It might be better this way because I will make many utility and visualization functions.

kbenoit:
> I probably should make **quanteda.torch** for this function. It might be better this way because I will make many utility and visualization functions.

Makes sense. The coercion methods are useful to have in core **quanteda**, as you have designed them. They depend only on the tokens structure, add no heavy dependency, and a dense document × position integer matrix is the format  **torch** (and **keras**) expect. That's the format Taylor Arnold recommended back in #1145, to return the indexed sequence and leave pad_sequences to **keras**, so `as.matrix.tokens()` effectively closes that long-standing request too. (There's also the call for broader integration from #2436.)

**torch** seems to be the happening package for native deep learning in R, with most of the surrounding ecosystem (like **luz**, **brulee**, **tabnet**, **mlr3torch**) now building on top of it rather than on **keras** or **TensorFlow**.

kbenoit (approved):

PR #2499, opened 2026-06-04
Add tokens_match()

koheiw:
Add `tokens_match()` which works like `dfm_match()`. Also `dfm()` got the `trim` argument to make the two matching functions compatible.

``` r
library(quanteda)
#> Package version: 4.4.1
#> Unicode version: 15.1
#> ICU version: 74.1
txt <- c("a b c", "b b b a", "c c b")
toks <- tokens(txt)

dfm(toks) %>% 
    dfm_match(c("a", "b", "c", "x"))
#> Document-feature matrix of: 3 documents, 4 features (41.67% sparse) and 0
#> docvars.
#>        features
#> docs    a b c x
#>   text1 1 1 1 0
#>   text2 1 3 0 0
#>   text3 0 1 2 0

tokens_match(toks, c("a", "b", "c", "x")) %>% 
    dfm(trim = FALSE)
#> Document-feature matrix of: 3 documents, 4 features (41.67% sparse) and 0
#> docvars.
#>        features
#> docs    a b c x
#>   text1 1 1 1 0
#>   text2 1 3 0 0
#>   text3 0 1 2 0

tokens_match(toks, c("a", "b", "c", "x")) %>% 
    as.matrix()
#>       [,1] [,2] [,3] [,4]
#> text1    1    2    3    0
#> text2    2    2    2    1
#> text3    3    3    2    0
    
tokens_match(toks[2:3], c("a", "b", "c", "x")) %>% 
    as.matrix()
#>       [,1] [,2] [,3] [,4]
#> text2    2    2    2    1
#> text3    3    3    2    0

tokens_match(toks, c("a", "b", "c", "x")) %>% 
    as.matrix(extract = 2:3)
#>       [,1] [,2] [,3] [,4]
#> text2    2    2    2    1
#> text3    3    3    2    0
```
There are also various improvements in internal functions to achieve consistent behaviours.


src/tokens_match.cpp
```diff
@@ -0,0 +1,76 @@
+#include "lib.h"
+//#include "dev.h"
+
+using namespace quanteda;
+
+Text match(Text tokens,
```
kbenoit: Rename to avoid possible conflict with `base::match()`?

Also, `tokens` is passed by value here, but the function only reads it and builds a fresh `tokens_new` — so this is an extra deep copy of every document's token vector, on top of the copy already made in `cpp_tokens_match`. Suggest a `const` reference:

```cpp
Text match(const Text &tokens,
           const std::vector<int> &ids) {
```

(`ids` is likewise never mutated, so `const` there too.)

Note this differs from `replace()`/`join_comp()`, which take `Text` by value — but those mutate the copy in place, whereas `match()` doesn't, so there's no reason to pay for the copy.

src/tokens_match.cpp
```diff
@@ -0,0 +1,76 @@
...
+Text match(Text tokens, 
+           std::vector<int> &ids){
+    
+    if (tokens.empty()) return {}; // return empty vector for empty text
+    
+    Text tokens_new;
+    tokens_new.reserve(tokens.size());
+    for (std::size_t i = 0; i < tokens.size(); i++) {
+        //Rcout << tokens[i] << ": ";
+        if (tokens[i] <= 0) {
```
kbenoit: Can't be negative, so change to `== 0`. Claude says: gcc/clang emit -Wtype-limits / -Wtautological-compare here, which CRAN flags.

src/tokens_match.cpp
```diff
@@ -0,0 +1,76 @@
+#include "lib.h"
+//#include "dev.h"
+
+using namespace quanteda;
+
+Text match(Text tokens, 
+           std::vector<int> &ids){
```
kbenoit: `std::vector<int> &ids` should be `const`. It's never mutated; `const std::vector<int> &ids` is correct and enables optimization.

src/tokens_match.cpp
```diff
@@ -0,0 +1,76 @@
...
+                texts[h] = match(texts[h], ids);
+            }    
+        });
+    });
+#else
+    for (std::size_t h = 0; h < H; h++) {
+        texts[h] = match(texts[h], ids);
+    }
+#endif
+    xptr->texts = texts;
```
kbenoit: This deep-copies the entire corpus into `texts` and then copies it all back into `xptr->texts` at the end — two full copies of potentially large data. Since each iteration writes to a distinct index `texts[h]`, working directly on the object via a reference is safe under `parallel_for` and avoids both copies:

```cpp
Texts &texts = xptr->texts;
```

Then the trailing `xptr->texts = texts;` can be dropped.

I realize `tokens_ngrams.cpp` and `tokens_replace.cpp` use this same copy-in/copy-out pattern, so this is consistent with the house style rather than a new problem — flagging it as optional cleanup, and possibly worth applying to those too.

kbenoit (approved):

src/tokens_match.cpp
koheiw: There are several places where `const Text &tokens` is more appropriate. I will make changes.

src/tokens_match.cpp
koheiw: Good point! Copying to `texts` is the legacy from the times before `tokens_xptr` was introduced. I will change all of them.

PR #2501, opened 2026-06-09
Add drop0 for dfm and fcm

koheiw:
Improving #2478, add `drop0.dfm()` and `drop0.fcm()` as S4 methods.


kbenoit (approved):

kbenoit:
@koheiw did we intend to export the `drop0()` methods for `dfm`/`fcm`?

There's no `exportMethods(drop0)` in `NAMESPACE`. The tests pass because `load_all()` makes the internal generic visible, but in an installed package `drop0(mydfm)` won't reach this method, but instead falls through to `Matrix::drop0()`, which strips the `dfm` class, docvars, and meta when **Matrix** is loaded, or just fails if not.

If it's meant to be user-facing we should add `@exportMethod drop0` to both methods. (If it's meant to be internal only, all good as merged.)

koheiw:
True, we want to export drop0() for dfm and fcm.

PR #2504, opened 2026-07-08
Export drop0

koheiw:


kbenoit:
I made one change. `drop0` is an S4 method (`setMethod`), so the `@method drop0 dfm`
/ `@method drop0 fcm` tags are S3-specific and redundant here (the export works via
`@export` being auto-detected as `exportMethods()`). I dropped the `@method` lines and used `@exportMethod drop0` instead.

kbenoit (approved):

PR #2506, opened 2026-07-18
Redesign C++ functions for tokens_xptr 

koheiw:
This PR redesigns type management for `tokens_xptr` objects so that read-only query operations no longer trigger hidden recompilation.

- optimizes C++ recompilation by reindexing token IDs in place
- adds `tokens_recompile()` for explicit or forced vocabulary compaction
- centralizes recompilation after token transformations
- updates case conversion, DFM construction, type counting, and Rcpp bindings
- preserves the complete type table when subsetting documents, aligning the behaviour of `tokens`, `tokens_xptr`, and `dfm`
- adds regression tests for duplicate and unused types, padding, encoding, metadata, and subsetting

These changes retain the unique-types contract while avoiding unnecessary recompilation during read-only operations.


R/tokens_xptr.R
```diff
@@ -78,7 +78,7 @@ ndoc.tokens_xptr <- function(x) {
 
 #' @export
 types.tokens_xptr <- function(x) {
-    cpp_get_types(x, TRUE)
+    get_types(x, all = FALSE)
```
kbenoit: Worth a one-line comment here noting that `all = FALSE` is deliberately *not* the same
as the old `recompile = TRUE` behaviour (no merging of duplicate strings) — future
readers will otherwise assume this is a drop-in equivalent, which is the source of the
regression flagged above.

src/tokens_xptr.cpp
```diff
@@ -153,11 +153,31 @@ IntegerVector cpp_get_freq(TokensPtr xptr, bool no_padding = false,
...
+        return encode(xptr->types);
+    
+    std::size_t G = xptr->types.size();
+    std::vector<bool> flag(G);
+    std::vector<std::string> types;
+    types.reserve(G);
+    
+    std::size_t H = xptr->texts.size();
+    for (std::size_t h = 0; h < H; h++) {
+        Text text = xptr->texts[h];
```
kbenoit: const Text& text = xptr->texts[h];

Minor: this copies each document's token vector by value on every call. Since avoiding
the cost of `recompile()` is the whole point of this PR, worth taking a `const&` here
instead — no behaviour change, just skips an O(total tokens) copy.

src/tokens_xptr.cpp
```diff
@@ -153,11 +153,31 @@ IntegerVector cpp_get_freq(TokensPtr xptr, bool no_padding = false,
 
 
 // [[Rcpp::export]]
-CharacterVector cpp_get_types(TokensPtr xptr, bool recompile = false) {
-    //Rcout << "cpp_types()\n";
-    if (recompile)
-        xptr->recompile();
-    return encode(xptr->types);
+CharacterVector cpp_get_types(TokensPtr xptr, bool all = true) {
```
kbenoit: `all = FALSE` filters by used **token ID**, but doesn't dedupe by **string content**.
If `xptr->recompiled == false` (e.g. right after `tokens_tolower()`/`tokens_wordstend()`,
which intentionally leave case/stem collisions unmerged until the next recompile), this
will return duplicate strings — see the repro in the review summary. This breaks the
`types()` contract in `man/types.Rd` ("unique types").

Suggestion: either call `xptr->recompile()` here when `!xptr->recompiled` before scanning,
or dedupe the `types` vector by content (e.g. via a `std::unordered_set`/seen-map) before
`return encode(types);` at line 180.

tests/testthat/test-tokens_xptr.R
```diff
@@ -184,6 +184,17 @@ test_that("tokens are not recompiled", {
         quanteda:::cpp_as_list(tokens_subset(xtoks, b)),
         lis[b]
     )
+    
+    expect_true(
```
kbenoit, line 188: This test covers the subsetting case but not the case-folding/stemming path that
actually triggers `recompiled = FALSE` with genuine string collisions. Suggest adding:

​```r
xtoks_dup <- tokens_tolower(as.tokens_xptr(tokens("Apple apple Banana")))
expect_length(unique(quanteda:::cpp_get_types(xtoks_dup, all = FALSE)),
              length(quanteda:::cpp_get_types(xtoks_dup, all = FALSE)))
​```

would have caught the duplicate-string regression before merge.

kbenoit (changes requested):
This is a good follow-up to some of the mutation-on-read side effects uncovered in recent PRs, from `types()`. And the `all = TRUE` default preserved for internal callers looks correct (I [and Claude 🙂] checked all call sites of `get_types()` and none of them regress).

However, I built this branch against `master` and confirmed a behavioural regression:
`types()` can now return duplicate strings, which contradicts `man/types.Rd` ("Get
**unique** types of tokens from a tokens object").

Repro:​    
```r
toks <- tokens_tolower(as.tokens_xptr(tokens("Apple apple Banana")))
types(toks)
# master:      "apple" "banana"
# this branch: "apple" "apple" "banana"
```

Root cause: `tokens_tolower()`/`tokens_wordstem()` replace the type table via
`set_types<-`, which sets `recompiled = FALSE` specifically because case-folding/stemming
can collapse previously-distinct IDs onto the same string — expecting a later
`recompile()` to merge them. The new `all = FALSE` scan filters by used ID only, so it
never merges those collisions, and (unlike before) never calls `recompile()` at all.

See inline comments for the specific spot and a couple of smaller notes.

koheiw:
I am still working on this branch so please wait. I will request review when it is ready.

kbenoit:
My bad - I will wait next time until your request a review!

koheiw:
This PR is apparently not ready for review yet, but I though it is better to discuss a bit. 

As you noticed, the main problem I wanted to address in this PR is the changes in tokens_xptr objects by the query functions like `types()`. The (bad) design was my attempt to minimize expensive recompilation. For example, recompilation takes about the half of the execution time in `tokens_tolower()`

Screenshot 2026-07-20 085117

There are two questions (1) where recompilation should happen, and (2) how to count unique types in verbose message.

For 1, `rebuild_tokens()` is a good place because the function is called by many `tokens_*()` functions. Yet, I conditioned it by `global$recompile` to internally defer recompilation in `tokens()`.

https://github.com/quanteda/quanteda/blob/6afa139d1204db6c04ace44c276410fc4610363f/R/object-builder.R#L119

For 2, we can count unique types in `stats_tokens()` while allowing (potentially) duplicated types.

https://github.com/quanteda/quanteda/blob/6afa139d1204db6c04ace44c276410fc4610363f/R/message.R#L124

With 1 and 2, the types and messages become like below. Currently, `tokens_remove()` triggers recompilation; `tokens_tolower()` does not but the message correctly says `1 types` despite the duplicate.

``` r
library(quanteda)
#> Package version: 4.4.1
#> Unicode version: 15.1
#> ICU version: 74.1
#> Parallel computing: 28 of 28 threads used.
quanteda_options(verbose = TRUE)
xtoks <- tokens("a A b", xptr = TRUE)
#> Creating a tokens_xptr from a character object...
#>  ...starting tokenization
#>  ...tokenizing 1 of 1 blocks
#>  ...preserving hyphens
#>  ...preserving elisions
#>  ...preserving social media tags (#, @)
#>  ...removing separators
#>  ...3 unique types
#>  ...complete, elapsed time: 0.01 seconds.
#> Finished constructing tokens_xptr from 1 document
types(xtoks)
#> [1] "a" "A" "b"
xtoks <- tokens_remove(xtoks, "b")
#> tokens_remove() changed from 3 types (1 documents, 3 tokens) to 2 types (1 documents, 2 tokens)
types(xtoks)
#> [1] "a" "A"
tokens_tolower(xtoks)
#> tokens_tolower() changed from 2 types (1 documents, 2 tokens) to 1 types (1 documents, 2 tokens)
#> Tokens consisting of 1 document (pointer to 0x25e8ce5c7e0).
#> text1 :
#> [1] "a" "a"
types(xtoks)
#> [1] "a" "a"
```
We could also defer recompilation in other `tokens_*()` functions but this is a policy decision. One idea is adding `quanteda_options(auto_recompile = FALSE)` like `global$recompile` and ask users to run `tokens_recompile()` manually before `as.tensor()` or `as.matrix()`.

koheiw:
I manage to make the recompilation function faster: it is about twice faster, only taking 30 milliseconds, so I can apply this to tokens objects more frequently. 

image

image

koheiw:
Now, `tokens_tolower()` returns unique types.

``` r
library(quanteda)
#> Package version: 4.4.1
#> Unicode version: 15.1
#> ICU version: 74.1
#> Parallel computing: 28 of 28 threads used.
quanteda_options(verbose = TRUE)
xtoks <- tokens("a A b", xptr = TRUE)
#> Creating a tokens_xptr from a character object...
#>  ...starting tokenization
#>  ...tokenizing 1 of 1 blocks
#>  ...preserving hyphens
#>  ...preserving elisions
#>  ...preserving social media tags (#, @)
#>  ...removing separators
#>  ...3 unique types
#>  ...complete, elapsed time: 0.02 seconds.
#> Finished constructing tokens_xptr from 1 document
types(xtoks)
#> [1] "a" "A" "b"
xtoks <- tokens_remove(xtoks, "b")
#> tokens_remove() changed from 3 types (1 documents, 3 tokens) to 2 types (1 documents, 2 tokens)
types(xtoks)
#> [1] "a" "A"
xtoks <- tokens_tolower(xtoks)
#> tokens_tolower() changed from 2 types (1 documents, 2 tokens) to 1 types (1 documents, 2 tokens)
types(xtoks)
#> [1] "a"
```

koheiw:
I am almost done with this PR but there is an old inconsistency between dfm and tokens. `tokens_subset.tokens_xptr()` is similar to `dfm_subset()`. I think I should make `tokens_subset.tokens()` to behave the same.

``` r
library(quanteda)
#> Package version: 4.4.1
#> Unicode version: 15.1
#> ICU version: 74.1
#> Parallel computing: 28 of 28 threads used.
quanteda_options(verbose = FALSE)
toks <- tokens(c("a A", "b"))
types(tokens_subset(toks, c(TRUE, FALSE)))
#> [1] "a" "A"

xtoks <- as.tokens_xptr(toks)
types(tokens_subset(xtoks, c(TRUE, FALSE)))
#> [1] "a" "A" "b"

dfmt <- dfm(toks, tolower = FALSE)
featnames(dfm_subset(dfmt, c(TRUE, FALSE)))
#> [1] "a" "A" "b"
```

<sup>Created on 2026-07-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

koheiw:
Now, all consistent!

``` r
library(quanteda)
#> Package version: 4.4.1
#> Unicode version: 15.1
#> ICU version: 74.1
#> Parallel computing: 28 of 28 threads used.
quanteda_options(verbose = FALSE)
toks <- tokens(c("a A", "b"))
types(tokens_subset(toks, c(TRUE, FALSE)))
#> [1] "a" "A" "b"

xtoks <- as.tokens_xptr(toks)
types(tokens_subset(xtoks, c(TRUE, FALSE)))
#> [1] "a" "A" "b"

dfmt <- dfm(toks, tolower = FALSE)
featnames(dfm_subset(dfmt, c(TRUE, FALSE)))
#> [1] "a" "A" "b"
```

<sup>Created on 2026-07-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

koheiw:
@kbenoit I don't know how to fix test-coverage but this branch is ready for your review.

kbenoit:
> @kbenoit I don't know how to fix test-coverage but this branch is ready for your review.

I found and fixed the `test-coverage` failure. The problem was in `TokensObj::recompile()`: `duplicated` was uninitialized, and empty type slots did not trigger the compaction pass. The coverage build uses different compiler flags, which exposed this undefined behaviour and left phantom empty features in many test objects.

(Thanks to Codex 5.6 Sol for figuring this out!)

kbenoit (approved):
Looks good. Great to improve the consistency (and efficiency) of the under-the-hood rewrite introduced in 4.0 with xptr objects.

I incremented the release version to 4.5 since we have introduced a number of behaviour changes and at least one new public-facing function.

kbenoit:
@koheiw Go ahead and merge if you are ok with https://github.com/quanteda/quanteda/pull/2506/commits/5bd5513d22509c8ce46a28e9b6218b95152202ff

koheiw:
@kbenoit thank you very much! Let's submit to the CRAN.

PR #2508, opened 2026-07-22
Update package website

koheiw:
This is a long overdue update of the package website. The main change is adding an example for convolutional neural network. I also made many small changes including updating the theme to Bootstrap 5. This branch requires #2506.


kbenoit:
Looks good. I noted the issue with older Bootstrap too. I like how you kept the same look in the new BS.

kbenoit (approved):

PR #2511, opened 2026-07-28
Fix non-TBB build error in cpp_tokens_segment

kbenoit:
## Problem

As noted in https://github.com/odelmarcelle/sentopics/pull/6#issuecomment-5096754974 by @odelmarcelle (thanks!), installing quanteda 4.5.0 fails to compile on setups **without TBB** with:

```
tokens_segment.cpp:125:33: error: use of undeclared identifier 'texts'
  125 |     for (std::size_t h = 0; h < texts.size(); h++) {
```

The `#else` (non-TBB) branch of `cpp_tokens_segment` used an undeclared identifier `texts` in its loop condition. The variable is `xptr->texts`, aliased earlier as `H = xptr->texts.size()`, and the TBB branch already iterates with `H`.

The bug is invisible on builds **with** TBB (`QUANTEDA_USE_TBB true`) because the `#else` block is preprocessed out and never compiled — which is why it slipped through.

## Fix

Use the existing loop bound `H` in the non-TBB branch, matching the TBB branch.

## Verification

Syntax-checked the file with `TBB` undefined (the failing configuration), forcing the previously-broken branch:

```
clang++ -std=gnu++20 -fsyntax-only -I<R include> -I<Rcpp include> -I../inst/include tokens_segment.cpp
# compiles cleanly
```

## Testing note

This is a compile-time bug in a branch not exercised by TBB builds; an R-level test cannot catch it (if it triggers, the package fails to install and no tests run). Functional coverage of `tokens_segment()` on multi-document input already exists in `tests/testthat/test-tokens_segment.R`. The regression-proof guard would be a CI job that builds with `TBB` undefined.


koheiw:
That is my bug... Thanks for fixing. It is a great idea to add CI tests without TBB.

koheiw (approved):

PR #2517, opened 2026-08-02
Fix deprecation of ntoken and ntype

koheiw:
The deprecation messages were not printed...


kbenoit (approved):
Good catch. We were new using **lifecycle** in 4.0 and clearly, did not use it correctly in some cases!

